PR #27, opened 2013-05-18
Allow setting verbosity by name (not just number)

udoprog:
this allows for string arguments to verbose, like; --verbose debug


adrienthebo:
@udoprog thanks for this contribution; I hadn't thought of a solution like this! Nice implementation!

PR #28, opened 2013-05-18
remove ply break point

udoprog:
looks like a mistaken commit


adrienthebo:
And another facepalm. Seriously, thanks for catching these.

PR #29, opened 2013-05-18
refactored breaking specs for forge module

udoprog:
- removed shadowed accessor for version on the forge module.


adrienthebo:
Insert facepalm here. Thank you!

PR #34, opened 2013-05-23
Update Readme with doc for setting PUPPETFILE and PUPPETFILE_DIR

csschwe:
Added a paragraph and example for setting PUPPETFILE and PUPPETFILE_DIR environment variables for those of us that take awhile to find this option in the code.


adrienthebo:
To be completely honest I had forgotten that I had even implemented this. I was aiming for feature parity with librarian-puppet and saw that those variables were being respect, ported them over, and promptly forgot about that. Thanks very much for this contribution!

Issue #24, opened 2013-05-07
Make modules dir overridable


igalic:
Right now, r10k's very opinionated configuration management is not overridable. If we want to create more complex structures of modules such proposed here: http://somethingsinistral.net/blog/configuration-management-as-legos/

In our environment we'd like to do something like that, however a configuration of:

```
:cachedir: '/var/cache/r10k'
:sources:
  :modules:
    remote: 'git@my-org/puppet-modules'
    basedir: '/etc/puppet/modules'
  :site:
    remote: 'git@my-org/puppet-site'
    basedir: '/etc/puppet/site'
  :projects: 'git@my-org/puppet-projects'
    basedir: '/etc/puppet/projects'

:purgedirs:
  - '/etc/puppet/modules'
  - '/etc/puppet/site'
  - '/etc/puppet/projects'
```

will create a very nested directory structure. It would be nice if it r10k simply cloned the envs into: `/etc/puppet/modules/$environments` instead of creating `/etc/puppet/modules/modules/$environments` -- say, with a switch for "shallow: true", or alternatively, if we could tell it to put the sites into "/etc/puppet/$environment/sites". (although I wouldn't know what kind of configuration scheme to use here)

We are aware that this would break full compatibility with librarian-puppet. This is something that r10k users would have to opt into.

csschwe:
I think this would be very useful  +1

larstobi:
+1

adrienthebo:
@igalic @csschwe @larstobi would #35 be sufficient for this use case?

igalic:
here's a test: https://gist.github.com/igalic/5767142 = wonder what I'm doing wrong here. Other than the fact that most of my modules are probably lacking their appropriate branches? But then why are the forge modules not installed either?

Ancillas:
Are you passing the -p flag to r10k when you run it?

abohne:
+1 on this.  I am using `/etc/puppet/modules/$environments` and all my branches (development, production, etc) reside under that directory.

So I want to end up with something like this:
`/etc/puppet/modules/development/stdlib`

With `stdlib` being defined as in `Puppetfile`.

#35 seems to get half-way there, but it doesn't quite solve the issue.

adrienthebo:
@abohne for clarification, you want all Puppetfile deployed modules at the root of the environment?

adrienthebo:
[...]

PR #35, opened 2013-05-24
Add support for setting moduledir in the Puppetfile

csschwe:
Add support to allow the module to be set in the Puppetfile

Example:

```
moduledir '/etc/puppet/modules/production/puppetlabs/'

mod 'lvm',
  :git => 'https://github.com/puppetlabs/puppetlabs-lvm.git'
```

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -57,6 +57,11 @@ class Puppetfile
     @forge = forge
   end
 
+  # @param [String] moduledir
+  def set_moduledir(moduledir)
+    @moduledir = moduledir
+  end
+
   # @param [String] name
   # @param [*Object] args
   def add_module(name, args)
@@ -97,6 +102,10 @@ class Puppetfile
 
     def forge(location)
       @librarian.set_forge(location)
+    end
+
+    def moduledir(location)
+      @librarian.set_moduledir(location)
     end
 
     def method_missing(method, *args)


adrienthebo:
Heh... for some reason I thought it would be a lot more work to implement this.

Thanks very much for this contribution, this seriously excellent. Right now I'm trying to get 1.0.0 stable so I can release it, and right now I don't have a really good way to continue to merge things in while I'm handling an RC cycle. As soon as 1.0.0 is out I'll merge this in. Thanks for taking the time to implement this!

For reference, this implements GH-24.

csschwe:
The only other thing I was thinking about.  Would it make more sense to do:

```
basemoduledir: '/etc/puppet/modules'
moduledir: 'production/puppetlabs'
```

so that `basemoduledir` is optional and allows portability

```
moduledir: 'production/puppetlabs'
```

I think this approach would work better with dynamic environments.  The next patch I am working on would allow each module to have a moduledir specified.  The only complicated part I have found with the multiple moduledir implementation is fully implementing purging.

adrienthebo:
We could do something like this:

``` ruby
basedir '/etc/puppet/modules'
moduledir '%{basedir}/production/puppetlabs'
```

This makes things less magical. Thoughts?

And when you want to specify multiple moduledirs, things start to get interesting.

One part of this is how to support that in the DSL. If we're trying to conform to how librarian works, then this could mean Puppetfiles could/should support the following:

``` ruby
moduledir 'production/puppetlabs' do
  mod 'puppetlabs/apache'
  mod 'puppetlabs/mysql'
end

mod 'camptocamp/kmod', :moduledir => 'production/camptocamp'
```

This is really convenient for users, but it makes the DSL more complex. Even if we do this by itself:

``` ruby
moduledir 'production/puppetlabs' do
  mod 'puppetlabs/apache'
  mod 'puppetlabs/mysql'
end
```

This means adding some sort of state to the DSL.

Doing global state with something like a single `moduledir` is pretty simple but trying to have some sort of state like this is going to mean a more complex DSL.

csschwe:
I have an implementation of the following that works.  The issue I have run into is getting purging to work correctly.  If you remove the last entry in a moduledir it is kind of a hack to get rid of it.

```
mod 'kmod', :moduledir => 'camptocamp'
mod 'lvm', :moduledir => 'puppetlabs'
```

I do have an example I can put up on my git repo.  I need to clean it up a bit first.

electrical:
I have several separate directories like 'librarie' and 'shared' in which i put certain modules.
What @adrienthebo is suggesting would be very useful indeed. ( being able to specify 1 dir and multiple modules )

Having a basedir ( absolute path; default to /etc/puppet/modules ? ) and a module dir ( relative path ) would make sense ( what @csschwe suggested )

adrienthebo:
I'm going to merge this for now, but I can't commit to this as an API change until 1.1.0 is released - meaning it's possible this might be changed before we release 1.1.0.

@csschwe thanks for the contribution and sorry about the delays!

PR #36, opened 2013-05-24
Request for Comment: Support multiple moduledirs in a Puppetfile

csschwe:
This is for discussion purposes.

Here is one possible way of doing multiple `moduledirs` in a Puppetfile.  I coded this to work with `r10k deploy environment -p`

Example Puppetfile:

```
moduledir 'my_modules'

mod 'lvm',
  :git => 'https://github.com/puppetlabs/puppetlabs-lvm.git',
  :moduledir => 'puppetlabs'

mod 'r10k',
  :git => 'https://github.com/adrienthebo/r10k.git'
```

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -57,6 +57,11 @@ class Puppetfile
     @forge = forge
   end
 
+  # @param [String] moduledir
+  def set_moduledir(moduledir)
+    @moduledir = moduledir
+  end
+
   # @param [String] name
   # @param [*Object] args
   def add_module(name, args)
@@ -97,6 +102,10 @@ class Puppetfile
 
     def forge(location)
       @librarian.set_forge(location)
+    end
+
+    def moduledir(location)
+      @librarian.set_moduledir(location)
     end
 
     def method_missing(method, *args)
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -20,6 +20,10 @@ class Puppetfile
   #   @return [String] The base directory that contains the Puppetfile
   attr_reader :basedir
 
+  # @!attribute [r] modbasedir
+  #   @return [String] The module base directory if set in Puppetfile 
+  attr_reader :modbasedir
+
   # @!attribute [r] moduledir
   #   @return [String] The directory to install the modules #{basedir}/modules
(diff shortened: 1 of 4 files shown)


lib/r10k/util/purgeable.rb
```diff
@@ -19,12 +19,17 @@ module Purgeable
 
   # @return [Array<String>] The present directory entries in `self.managed_directory`
   def current_contents
-    dir = self.managed_directory
```
adrienthebo, line 22: I think it might be easier to expand the `Purgeable` mixin to understand `managed_directories` as a first class concept. We could always assume that each file that includes this mixin could include multiple directories, and possibly have selective directory purging or something.

In short, I think my initial implementation of this class was a bit limited and this probably needs to be generally expanded. Dunno how exactly though.

lib/r10k/puppetfile.rb
```diff
@@ -57,23 +61,50 @@ def set_forge(forge)
...
+      end
+    end
+    @modules << R10K::Module.new(name, mod_dir, args)
   end
 
   include R10K::Util::Purgeable
 
   def managed_directory
-    @moduledir
+    @dirs = @modules.map { |mod| mod.basedir }.uniq
```
adrienthebo, line 97: Things like this `uniq` statement emphasize the later point; the Purgeable stuff needs to actually understand multiple purgeable directories.

lib/r10k/puppetfile.rb
```diff
@@ -57,23 +61,50 @@ def set_forge(forge)
     @forge = forge
   end
 
+  # @param [String] modbasedir
+  def set_modbasedir(modbasedir)
```
adrienthebo, line 65: I'm feeling a bit obtuse tonight; what's the difference between `modbasedir` and `moduledir`?

lib/r10k/puppetfile.rb
```diff
@@ -20,6 +20,10 @@ class Puppetfile
   #   @return [String] The base directory that contains the Puppetfile
   attr_reader :basedir
 
+  # @!attribute [r] modbasedir
```
adrienthebo, line 23: Oh, wow, I shouldn't read PRs in reverse order. `modbasedir` == the root directory for moduledirs. Git it.

PR #44, opened 2013-06-13
When moduledir is relative, prepend the basedir

### LINKED ISSUE #43: When using a relative moduledir r10k deploys the modules relative to the current directory
```
# The location to use for storing cached Git repos
:cachedir: '/var/tmp/r10k-cache'

# A list of git repositories to create
:sources:
  :modules:
    remote: 'git@panic:puppet-modules'
    basedir: '/tmp/puppet/modules'

# This directory will be purged of any directory that doesn't map to a
# git branch
:purgedirs:
  - '/tmp/puppet/modules'
```

With Puppetfiles ala

```
moduledir 'production'

forge "http://forge.puppetlabs.com"

mod 'puppetlabs/mysql', "0.6.1"
mod 'puppetlabs/nodejs', "0.2.1"
mod 'puppetlabs/stdlib', "3.2.0"
mod 'puppetlabs/vcsrepo', "0.1.1"
mod 'ripienaar/concat', "0.2.0"
mod "ispavailability/elasticsearch", "0.0.7"

mod "composer",
:git => "git@panic:puppet-composer"

mod "git",
:git => "git@panic:puppet-git"

mod "gitpubsub",
:git => "git@panic:puppet-gitpubsub"
```

results in r10k deploying those modules into `$PWD/production/`

See this gist for a complete example: https://gist.github.com/igalic/5767142

igalic:
The basic premise works, however right now we're appending the moduledir
again at the end. That shouldn't happen (imo), but maybe I'm
misunderstanding some core-concept of r10k here.

In general, this fixes #43  and when used with `moduledir ''` it also accommodates for different use-cases, such as managing more than just modules with r10k.

### DIFF
--- a/lib/r10k/module.rb
+++ b/lib/r10k/module.rb
@@ -2,7 +2,6 @@ require 'r10k'
 require 'r10k/util/path'
 
 module R10K::Module
-  include R10K::Util::Path
 
   # Register an inheriting  class for later generation
   def self.included(klass)
@@ -30,7 +29,7 @@ module R10K::Module
     @basedir = basedir
 
     if args.is_a? Hash and args.has_key?(:moduledir)
-      raise "Setting an absolute :moduledir path is not supported in this place!" if is_absolute?(args[:moduledir])
+      raise "Setting an absolute :moduledir path is not supported in this place!" if R10K::Util::Path.is_absolute?(args[:moduledir])
       @moduledir = File.join(@basedir, args[:moduledir])
     end
 
--- a/lib/r10k/module.rb
+++ b/lib/r10k/module.rb
@@ -17,24 +17,33 @@ module R10K::Module
   # with `name, args`, and generates an object of that class.
   #
   # @param [String] name The unique name of the module
-  # @param [String] basedir The root to install the module in
+  # @param [String] basedir The root of the module installations, in case we have to modify moduledir
+  # @param [String] moduledir The root to install the module in
   # @param [Object] args An arbitary value or set of values that specifies the implementation
   #
   # @return [Object < R10K::Module] A member of the implementing subclass
-  def self.new(name, basedir, args)
+  def self.new(name, basedir, moduledir, args)
+
+    @moduledir = moduledir
+    @basedir = basedir
+
+    if args.is_a? Hash and args.has_key?(:moduledir)
+      @moduledir = File.join(@basedir, args[:moduledir])
+    end
+
     if implementation = @klasses.find { |klass| klass.implement?(name, args) }
-      obj = implementation.new(name, basedir, args)
+      obj = implementation.new(name, @moduledir, args)
       obj
     else
       raise "Module #{name} with args #{args.inspect} doesn't have an implementation. (Are you using the right arguments?)"
     end
   end
 
-  attr_accessor :name, :basedir
+  attr_accessor :name, :basedir, :moduledir
 
   # @return [String] The full filesystem path to the module.
   def full_path
-    File.join(@basedir, @name)
+    File.join(@moduledir, @name)
   end
 end
 
--- a/lib/r10k/module.rb
+++ b/lib/r10k/module.rb
@@ -1,6 +1,8 @@
 require 'r10k'
+require 'r10k/util/path'
 
 module R10K::Module
+  include R10K::Util::Path
 
   # Register an inheriting  class for later generation
   def self.included(klass)
@@ -28,6 +30,7 @@ module R10K::Module
     @basedir = basedir
 
     if args.is_a? Hash and args.has_key?(:moduledir)
+      raise "Setting an absolute :moduledir path is not supported in this place!" if is_absolute?(args[:moduledir])
       @moduledir = File.join(@basedir, args[:moduledir])
     end
 
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -22,9 +22,9 @@ class Forge
 
   attr_accessor :owner, :full_name
 
-  def initialize(name, basedir, args)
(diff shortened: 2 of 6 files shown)


igalic:
btw, I don't think I'm (directly) responsible for the travis failure, but perhaps I should have written some tests for this.

lib/r10k/module.rb
```diff
@@ -17,24 +18,34 @@ def self.included(klass)
...
   #
   # @param [String] name The unique name of the module
-  # @param [String] basedir The root to install the module in
+  # @param [String] basedir The root of the module installations, in case we have to modify moduledir
+  # @param [String] moduledir The root to install the module in
   # @param [Object] args An arbitary value or set of values that specifies the implementation
   #
   # @return [Object < R10K::Module] A member of the implementing subclass
-  def self.new(name, basedir, args)
+  def self.new(name, basedir, moduledir, args)
```
adrienthebo, line 26: Is `args` a required argument? If so then this is a backwards incompatible change. This would mean that we have to jump to r10k version 2. That's totally fine, major versions are cheap, but that's a call that we'll have to make.

adrienthebo:
It looks like the reason for the spec failures is related to https://github.com/adrienthebo/r10k/pull/44#discussion_r4953183 . Aside from that this looks awesome; thanks for the submission!

lib/r10k/module.rb
```diff
@@ -17,24 +18,34 @@ def self.included(klass)
...
+
+    @moduledir = moduledir
```
adrienthebo, line 28: These shouldn't be instance variables, since they're only meaningful in this method and shouldn't be available anywhere else.

lib/r10k/module.rb
```diff
@@ -17,24 +18,34 @@ def self.included(klass)
...
+    @basedir = basedir
+
+    if args.is_a? Hash and args.has_key?(:moduledir)
```
adrienthebo, line 31: Doing manipulations on the path like this breaks encapsulation. The current behavior assumes that the basedir provides the location to install the module, which should be sufficient. Handling the path manipulation comes from the Puppetfile and should be there.

lib/r10k/module.rb
```diff
@@ -17,24 +18,34 @@ def self.included(klass)
...
-      obj = implementation.new(name, basedir, args)
+      obj = implementation.new(name, @moduledir, args)
       obj
     else
       raise "Module #{name} with args #{args.inspect} doesn't have an implementation. (Are you using the right arguments?)"
     end
   end
 
-  attr_accessor :name, :basedir
+  attr_accessor :name, :basedir, :moduledir
```
adrienthebo, line 44: I might be misreading this, but is `basedir` being more or less replaced with the `moduledir` instance variable? I don't see where one was really used over the other.

lib/r10k/module/forge.rb
```diff
@@ -22,9 +22,9 @@ def self.implement?(name, args)
 
   attr_accessor :owner, :full_name
 
-  def initialize(name, basedir, args)
+  def initialize(name, moduledir, args)
     @full_name = name
-    @basedir   = basedir
+    @moduledir   = moduledir
```
adrienthebo, line 27: Replacing `basedir` with `moduledir` breaks SemVer. If people are relying on that method and it's suddenly replaced then they'll get a nasty surprise. In general existing methods can't change in functionality or be removed until major versions are released.

adrienthebo:
I've done some work to fixup my concerns in this PR at https://github.com/adrienthebo/r10k/tree/pull-44 ; could you test it and see if it works for your environment?

igalic:
I just realized that comments are _here_ not in your tree.
So I'm moving [this comment](https://github.com/adrienthebo/r10k/commit/003077459b68fbdde326a2ae9e7bb429eb33430a#commitcomment-3558398) over here:

Something here breaks my current config: https://gist.github.com/igalic/5918686

adrienthebo:
@igalic sorry about the delay in getting this merged, I've had to spend the last few months working on r10k. I should have some time this week to get it merged in; thanks for the contribution and thanks for your patience!

igalic:
no worries, I'm just listening to @antaflos' stories from puppetconf, from which I'm concluding that you've been rather busy ;)

adrienthebo:
I've looked at merging this pull request in, and other options for solving this, and there is no way to do it without breaking the current API. It requires changing too many different parts of the current code base and it's impossible to make those changes in a backwards compatible manner. It's frustrating but that's the cost of respecting semver.

I've been kicking around a different solution for this. Adding this feature would be backwards incompatible, but only if we changed the Puppetfile API. In addition the Puppetfile as a format is not my favorite thing since it gets ugly and messy to use Ruby as a data format. I've been thinking of implementing something like this:

``` yaml
---
puppetlabs-modules:
  apt:
    source: git
    uri: git://github.com/puppetlabs/puppetlabs-apt
    ref: '1.1.0'
  puppetlabs_yum:
    source: forge
    module: stahnma/puppetlabs_yum
site-modules:
  nginx:
    source: git
    uri: git://gitserver/nginx.git
```

antaflos:
I for one would have no problem switching from the Puppetfile format to an r10k-specific format, _especially_ if that allowed for r10k to get closer to supporting your [Configuration Management as Legos](http://somethingsinistral.net/blog/configuration-management-as-legos/) concept.

Do I understand it right, that in your example r10k would manage the apt and puppetlabs_yum modules in `/etc/puppet/environments/$environment/puppetlabs-modules/{apt,puppetlabs_yum}` and the nginx module in `/etc/puppet/environments/$environment/site-modules/nginx`?

igalic:
:+1:

adrienthebo:
Closing in favor of GH-141, sorry about the duplication!

Issue #52, opened 2013-08-01
Purging all environments when specifying more than one 


lutaylor:
We are currently having the following problem.

If we specify more than one git repo in our yaml file it will purge out the directories right after deploying them.

e.g.
...
    [R10K::Task::Module::Sync - INFO] Deploying fedora into /etc/puppet/environments/metro/modules
    [R10K::Task::Module::Sync - INFO] Deploying stackbase into /etc/puppet/environments/metro/modules
    [R10K::Task::Deployment::PurgeEnvironments - INFO] Purging stale environments from /etc/puppet/environments
    [R10K::Task::Deployment::PurgeEnvironments - INFO] Purging stale environments from /etc/puppet/environments

Here is our yaml file

```
# The location to use for storing cached Git repos
:cachedir: '/var/cache/r10k'

# A list of git repositories to create
:sources:
 :metro:
   remote: 'git@github.com:discoverygardenpuppet/env_metro.git'
   basedir: '/etc/puppet/environments'
 :env_ir:
   remote: 'git@github.com:discoverygardenpuppet/env_ir.git'
   basedir: '/etc/puppet/environments'

:purgedirs:
  - '/etc/puppet/environments'
```

If we comment out either one and rerun it deploys fine and doesn't delete them after. I have tried deleting the /var/cache/r10k as well and it has no impact.

Ancillas:
I created a small workaround for this last week in my environment, but it's not good enough to be considered a patch.

https://gist.github.com/Ancillas/6135084

Line 72 is what I commented out to stop r10k from purging.

adam-vessey:
Been looking into this a bit... Looks like it's because both use the same "basedir", and so detect each others directories as being "stale"... So when the first purges it's stale stuff, it deletes those of the second, and then the second deletes those of the first. This seem to come out of a bad assumption made in https://github.com/adrienthebo/r10k/blob/master/lib/r10k/util/purgeable.rb#L21-L28, where assumes each "managed directory" is only used by a single source. 

... I'm thinking that it may be necessary to:
1. list the contents of each managed directory (in a hash of arrays?), 
2. iterate over each source and remove the "desired" contents from our listing
3. delete all entries remaining in the listing from the filesystem

[...]

Issue #48, opened 2013-07-02
Multiple apps with same branch names


Finchster:
It appears that if you have multiple sources/apps defined and a branch has the same name in both sources then it tries to merge the two branches together under one environment which gets messy. After a few tests it does not create two separate environments.

The preferred method for handling this would be to append the repo's name to the branch name and name the environment $repo_$branch. This way dynamic environments can be created by separate application owners without worrying about not re-using a branch name.

adam-vessey:
This is pretty much implemented in https://github.com/adrienthebo/r10k/pull/53.

Ancillas:
Ah, nm.  I get it now.

glarizza:
I'm hitting this same issue in 1.0.0 too --> https://gist.github.com/glarizza/5a81e45f2f489d87d464 and the prefixing should alleviate it...

adrienthebo:
This should be fixed by GH-53, so I'm targeting this at `1.1.0.`

binford2k:
IMHO, the prefix solution is kind of hackyugly. It means that my modulepath of:

`/etc/puppetlabs/puppet/environments/$environment/modules`

Doesn't work. Instead I need to use:

`/etc/puppetlabs/puppet/environments/puppet_$environment/modules`

and something similar for hiera. It works, but I don't think it should be necessary.

adam-vessey:
@binford2k Use the [Puppetfile](https://github.com/adrienthebo/r10k#puppetfile-support) to grab the modules on which each environment is dependent? Your modpath can include multiple paths...

glarizza:
The bigger deal is that this is going to be a backwards-incompatible change in the middle of a series...the environment folders that are created aren't going to match your existing $modulepath setting in puppet.conf.  While I think the prefixing is necessary and fixes some big R10k bugs, actually prefixing the FILENAME on disk should be kept for a major release.

Is there a way internally to track the branch as "puppet_#{branchname}" but only create the folder name on-disk as "#{branchname}"?  Within a single repo/source, you're not going to have conflicting branch names, so I don't think we'd have any clashes here.  Thoughts?

adam-vessey:
If I'm understanding correctly, what would address your concerns is to only default to using prefixes when there are multiple repos under the same basedir, and only in those basedirs where there are multiple repos being cloned?

adrienthebo:
[...]

PR #53, opened 2013-08-02
Allow optional prefixing.

adam-vessey:
Allows "prefix" to be specified on each source (as a boolean true/false, defaulting to false), whether or not the name will be used to prefix the directory name used.

### DIFF
--- a/lib/r10k/deployment.rb
+++ b/lib/r10k/deployment.rb
@@ -59,8 +59,10 @@ class Deployment
   private
 
   def load_sources
-    @_sources = @config.setting(:sources).map do |(name, hash)|
-      R10K::Deployment::Source.vivify(name, hash)
+    sources = @config.setting(:sources)
+    prefix = sources.length > 1
+    @_sources = sources.map do |(name, hash)|
+      R10K::Deployment::Source.vivify(name, hash, prefix)
     end
   end
 
--- a/lib/r10k/deployment/environment.rb
+++ b/lib/r10k/deployment/environment.rb
@@ -27,11 +27,13 @@ class Environment
   # @param [String] remote
   # @param [String] basedir
   # @param [String] dirname The directory to clone the root into, defaults to ref
-  def initialize(ref, remote, basedir, dirname = nil)
+  # @param [String] source_name An additional string which may be used with ref to build dirname
+  def initialize(ref, remote, basedir, dirname = nil, source_name = "")
     @ref     = ref
     @remote  = remote
     @basedir = basedir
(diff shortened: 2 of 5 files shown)


adrienthebo:
@adam-vessey this is absolutely excellent, thanks for taking the time to add this! It's late here and I'm not in the most clear-headed of states, but to confirm - this will add an optional `prefix` key for sources that will be prefixed to each environment?

adam-vessey:
That's it. It's not actually the value of the `prefix` key that we use (it's currently expected to be a boolean), but... the errr...  I'm not familiar with the nomenclature in yaml... The name of the source node? So each source is "under" a name, right? So if `git@github.com:example/repo` had a branch `master` and `test`:

```
# The location to use for storing cached Git repos
:cachedir: '/var/cache/r10k'

# A list of git repositories to create
:sources:
 :our_example:
   remote: 'git@github.com:example/repo
   prefix: true

:purgedirs:
  - '/etc/puppet/environments'
```

would result in environments named `our_example_master` and `our_example_test`.

Then, when we look for "stale" branches, we know to only search for things with our prefix, so multiple repos don't wipe each other out.

adam-vessey:
It is admittedly kinda weird... If there are multiple sources and any one in not prefixed, everything except those environments from the unprefixed source will be deleted (and if there are two unprefixed, everything will be deleted), as in #52...  We should maybe consider making it automatically prefix if there are multiple sources?

Finchster:
I think the automatic prefix with multiple sources makes more sense as I cant see a use case for having one with and one without?

Thanks for looking at this, as I was saying in #48, this is a killer feature for us to use in production as it will allow us to split up our puppet modules per in-house application and let different teams maintain them individually.

adam-vessey:
Now automatically prefixes if there is more than one source, and prefixing can still be forced on with the `prefix` parameter if there is only one source.

adrienthebo:
Merged into master in 1441b0c; this should be released in 1.1.0. Right now I'm preparing for Puppetconf so I don't have the time to work on the new release until after the 24th, but after then I should have a lot more time to work on r10k. @adam-vessey thanks very much for the submission and @Finchster thanks for offering feedback!

diginc:
RE: Multiple sources, one with prefix => true and others with prefix => false AND with shared basedirs.

This seems to bug purging if they share the same basedir directory, seems the non-prefixed source remains and the prefixed one gets deleted:

Details moved to #64

Edit: Details moved to bug.

adrienthebo:
@diginc could you file a new bug report for this behavior?

PR #59, opened 2013-09-05
Pulp based forge compatibility for r10k

abhaychrungoo:
puppet module install {...} -v 0.0.1 AND
puppet module install {...} -v v0.0.1

behave identically for forge.puppetlabs.com

For a pulp based forge, the first works, and the second throws up.

r10k generates the second

This minor patch changes the command generated for pulp-puppetfile-forge compatibility by losing the 'v'
We've been using this for a couple of months now. A merge will move us off a fork.


lib/r10k/module/forge.rb
```diff
@@ -35,14 +35,21 @@ def initialize(name, basedir, args)
 
   def sync(options = {})
     return if insync?
-
+    # A Pulp based puppetforge http://www.pulpproject.org/ wont support a 
+    # puppet module install abc/xyz --version=v1.5.9 command
+    # but puppetlabs forge will support puppet module install abc/xyz --version=1.5.9 command
+    # removing v from the semver.to_s for constructing the command ensures compatibility across both
+    if @version
+     buffer = @version.inspect
```
adrienthebo: If this is assuming version is of type `Semver`, it would be preferred to use `to_s`. `inspect` is generally used to print a human readable version of an object but we want the string representation. In this case `to_s` is aliased to `inspect`, but I'm a pedant and I think it's slightly more correct. :)

lib/r10k/module/forge.rb
```diff
@@ -35,14 +35,21 @@ def initialize(name, basedir, args)
...
+     version_string = buffer.start_with?('v') ? buffer.sub(/v/,'') : buffer
```
adrienthebo: we can do `buffer.gsub!(/^v/, '') to strip off any leading 'v' if present, which would be a bit more simple.

adrienthebo:
@fourteenKata thanks for the contribution! I've added a couple of notes in the diff, would you be able to address them before I merge this in?

abhaychrungoo:
Forgive the git/ruby journeyman. I've cleaned it up. 
SemVer will produce atleast a v0.0.0 
we dont really need /^v/, but it conveys the intent better

abhaychrungoo:
In the unit tests, I have left both in there. 

```
  its(:version) { should eq '8.0.0' }
  its(:version) { should eq 'v8.0.0' }
```

---

$ rspec forge_spec.rb 
Run options: exclude {:integration=>true}
...................

Finished in 0.05989 seconds
19 examples, 0 failures

adrienthebo:
I updated the tests for the forge_spec.rb in master, which introduced a merge conflict. Because of this I've rebased this branch against master and merged it in 6b1d4a7. This should be released in 1.1.0; thanks again for the contribution!

PR #62, opened 2013-09-11
update spec tests

jlambert121:
I tried to fix the travis errors and mocha warnings.

It appears the memoization stuff was removed for the Git::Cache module.
Updated for registry methods.

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'rspec', '~> 2.14.0'
   s.add_development_dependency 'mocha', '~> 0.10.5'
-  s.add_development_dependency 'minitest',
+  s.add_development_dependency 'minitest'
 
   s.files        = Dir.glob "{bin,lib}/**/*"
   s.require_path = 'lib'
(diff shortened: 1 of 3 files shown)


jlambert121:
Looking at the travis build error I now notice mocha is pinned to ~> 0.10.5, any reason not to update that?  (0.14.0 is what I tested against)

r10k.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'rspec', '~> 2.14.0'
   s.add_development_dependency 'mocha', '~> 0.10.5'
+  s.add_development_dependency 'minitest'
```
adrienthebo, line 29: Why is minitest needed for this?

adrienthebo:
I'm actually hoping to remove the mocha dependency entirely and use the rspec mocking framework instead, but between Puppetconf and the Puppetconf plague I haven't gotten a chance to finalize that work. Is the main problem the registry behavior change?

jlambert121:
Minitest was because of the updated version of mocha - if you're looking at getting rid of mocha that would remove the need as well.

Yeah, everything was around the registry stuff.

adrienthebo:
So... you caught me being a lazy dev. I was having issues with mocha as a library, overhauled a chunk of the API, broke the specs, and let the build stay in a broken state. I do hate closing pull requests and I do appreciate the time you took to do this, but it did give me the kick in the pants I needed to things cleaned up - so thanks! :)

Since I did remove the mocha dependency and refactor the registry tests, I'm closing this pull request.

jlambert121:
No worries - I was just fixing the tests before I started playing with some things.

PR #66, opened 2013-09-14
Also checkout the environment's branch after cloning the environment.

eliasp:
When doing a

```
r10k deploy environment
```

The result will be a directory tree like:

```
etc
+-- puppet
    +-- environments
        +-- development
        +-- integration
        +-- production
        +-- test-feature-foo
        +-- test-feature-bar
```

Whereas each environment represents a branch of the source repository.

The problem is with the respective environment's directory: instead of containing the respective branch, it'll always contain a checkout of the `production` branch (I wasn't able yet to find out what causes here the default to `production` instead of something else like the first branch name in alpha-numerical order like `development`).

This PR ensures, the corresponding environment's branch is checked out from the cloned repository.

It would be also possible to clone a specific branch right away (`git clone --branch --single-branch …`), but this would break functionality for < git 1.7.10.

### DIFF
--- a/lib/r10k/git/working_dir.rb
+++ b/lib/r10k/git/working_dir.rb
@@ -78,6 +78,7 @@ class WorkingDir < R10K::Git::Repository
     # that doing a normal `git pull` on a directory will work.
     git "clone --reference #{@cache.path} #{@remote} #{@full_path}"
     git "remote add cache #{@cache.path}", :path => @full_path
+    git "checkout #{@ref}", :path => @full_path
   end
 
   def fetch


eliasp:
Would it probably make sense to ensure the correct branch is checked out for already cloned environments? Otherwise users would have to delete/manually checkout them to profit from this fix for already existing environments.

eliasp:
Oh, and this is probably the point where the "default" branch is taken from (`git branch -a`):

```
  remotes/origin/HEAD -> origin/production
```

adrienthebo:
This seems like a pretty ugly bug, do you know if this was the result of a  regression or does this affect all versions of r10k?

eliasp:
Can't tell for sure, started using different environments beyond "playing around with r10k" just now.
If you could come up with a test for this, a quick `git bisect` might tell us more about it.

adrienthebo:
As it turns out I broke the build, and didn't catch it. Yay!

I did some investigating and it turns out that the initial clone is in the wrong branch, but subsequent runs will correct this. Thanks for catching and fixing this!

Issue #87, opened 2014-02-02
Replacement of librarian-puppet with local sources


ironpinguin:
Hi,

we use also local resources in our project for modules where are managed with the rest of the application code in a git repository.
To migrate our project to r10k we need the possibility to define modules with the :path argument.

I start to implement this in a fork. It will be nice if you use this new module from my pull request.

Ciao
Michele

adrienthebo:
This seems like a reasonable thing to include. Could you provide some use cases/motivations for how you use this feature and examples for how others could use it?

bvis:
Hi,

We are using this feature as well, this is very useful when you are trying to develop a puppet module and you do not want to publish it still to forge.

It is very useful to force the usage of a particular revision in the repo instead.

walidshaari:
yes with the lack of standardising how you can have an internal forge right now (e.g. pulp vs. django-forge,..etc) that would be a good quick solution to the development use case.

PR #88, opened 2014-02-02
Port of librarian-puppet local sources

ironpinguin:
See Issue #87 :-)


adrienthebo:
We merged a separate implementation of local sources that was geared at protecting modules check in to the r10k control repository; unfortunately that pull request has superseded this pull request. In addition using the `:path` symbol presents a potential conflict with a possible `:path` option for Git modules, so I don't want to incorporate code that will conflict with future designs. I'm closing this out; sorry for the delay in feedback and apologies for not being able to merge this.

PR #101, opened 2014-03-24
Feature: Implement locking an environment to a ref

ryandoyle:
Heya

I've added a feature that allows a user to configure locking a git ref (our use case: a git tag) to an environment. I've added some extra test coverage and cleaned up some unused constructor args while I was digging in the area.

rspec tests all pass, although I did have some trouble running `rake spec:system` (although the failures are in a different area).

Cheers!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -261,6 +261,21 @@ without being able to accidentally impact other groups.
         remote: 'git@github.com:my-org/app2-environments'
         basedir: '/etc/puppet/environments'
 
+### Locking an environment to a ref
+
+You may want to lock an environment to a specific git ref (such as a 'tested' tag). This can be specified in the
+yaml configuration file.
+
+        :cachedir: '/var/cache/r10k'
+        :sources:
+          # branch in /etc/puppet/environments
+          :plops:
+            remote: 'git@github.com:my-org/org-shared-modules'
+            basedir: '/etc/puppet/environments'
+            environment:
+              master:
+                ref: 'tested'
+
 More information
 ----------------
 
--- a/lib/r10k/deployment/environment.rb
+++ b/lib/r10k/deployment/environment.rb
@@ -28,12 +28,12 @@ class Environment
   # @param [String] basedir
   # @param [String] dirname The directory to clone the root into, defaults to ref
   # @param [String] source_name An additional string which may be used with ref to build dirname
-  def initialize(ref, remote, basedir, dirname = nil, source_name = "")
+  def initialize(ref, remote, basedir, source_name = "")
     @ref     = ref
     @remote  = remote
     @basedir = basedir
     alternate_name =  source_name.empty? ? ref : source_name + "_" + ref
-    @dirname = sanitize_dirname(dirname || alternate_name)
+    @dirname = sanitize_dirname(alternate_name)
 
     @working_dir = R10K::Git::WorkingDir.new(@ref, @remote, @basedir, @dirname)
 
--- a/lib/r10k/deployment/environment.rb
+++ b/lib/r10k/deployment/environment.rb
@@ -7,9 +7,9 @@ class Environment
 
   include R10K::Logging
 
-  # @!attribute [r] ref
-  #   The git ref to instantiate into the basedir
-  attr_reader :ref
+  # @!attribute [r] name
+  #   The git name to instantiate into the basedir
+  attr_reader :name
 
   # @!attribute [r] remote
   #   The location of the remote git repository
@@ -23,19 +23,19 @@ class Environment
   #   @return [String] The directory name to use for the environment
   attr_reader :dirname
 
-  # @param [String] ref
+  # @param [String] name
   # @param [String] remote
   # @param [String] basedir
-  # @param [String] dirname The directory to clone the root into, defaults to ref
-  # @param [String] source_name An additional string which may be used with ref to build dirname
-  def initialize(ref, remote, basedir, source_name = "")
-    @ref     = ref
+  # @param [String] dirname The directory to clone the root into, defaults to name
+  # @param [String] source_name An additional string which may be used with name to build dirname
+  def initialize(name, remote, basedir, source_name = "")
+    @name     = name
     @remote  = remote
     @basedir = basedir
-    alternate_name =  source_name.empty? ? ref : source_name + "_" + ref
+    alternate_name =  source_name.empty? ? name : source_name + "_" + name
     @dirname = sanitize_dirname(alternate_name)
 
-    @working_dir = R10K::Git::WorkingDir.new(@ref, @remote, @basedir, @dirname)
+    @working_dir = R10K::Git::WorkingDir.new(@name, @remote, @basedir, @dirname)
 
     @full_path = File.join(@basedir, @dirname)
   end
--- a/lib/r10k/deployment/environment.rb
+++ b/lib/r10k/deployment/environment.rb
@@ -23,19 +23,24 @@ class Environment
   #   @return [String] The directory name to use for the environment
   attr_reader :dirname
 
+  # @!attribute [r] ref
+  #   @return [String] Git ref that this environment is based on
+  attr_reader :ref
+
   # @param [String] name
   # @param [String] remote
   # @param [String] basedir
-  # @param [String] dirname The directory to clone the root into, defaults to name
   # @param [String] source_name An additional string which may be used with name to build dirname
-  def initialize(name, remote, basedir, source_name = "")
+  # @param [String] ref Optional git ref to use
+  def initialize(name, remote, basedir, source_name = nil, ref = nil)
     @name     = name
     @remote  = remote
     @basedir = basedir
-    alternate_name =  source_name.empty? ? name : source_name + "_" + name
+    alternate_name =  source_name.nil? ? name : source_name + "_" + name
     @dirname = sanitize_dirname(alternate_name)
+    @ref = ref.nil? ? name : ref
 
-    @working_dir = R10K::Git::WorkingDir.new(@name, @remote, @basedir, @dirname)
+    @working_dir = R10K::Git::WorkingDir.new(@ref, @remote, @basedir, @dirname)
 
     @full_path = File.join(@basedir, @dirname)
   end
--- a/lib/r10k/deployment/source.rb
+++ b/lib/r10k/deployment/source.rb
@@ -97,7 +97,7 @@ class Source
     if @cache.cached?
       @environments = @cache.branches.map do |branch|
         if @prefix
-          R10K::Deployment::Environment.new(branch, @remote, @basedir, nil, @name.to_s())
+          R10K::Deployment::Environment.new(branch, @remote, @basedir, @name.to_s())
         else
           R10K::Deployment::Environment.new(branch, @remote, @basedir)
         end
--- a/lib/r10k/deployment/source.rb
+++ b/lib/r10k/deployment/source.rb
@@ -53,12 +53,12 @@ class Source
     new(name, remote, basedir, prefix_outcome, environment_refs)
   end
 
-  def initialize(name, remote, basedir, prefix = nil, environment_refs = {})
+  def initialize(name, remote, basedir, prefix = nil, environment_refs = nil)
     @name    = name
     @remote  = remote
     @basedir = basedir
     @prefix = prefix.nil? ? false : prefix
-    @environment_refs = environment_refs
+    @environment_refs = environment_refs.nil? ? {} : environment_refs
 
     @cache   = R10K::Git::Cache.generate(@remote)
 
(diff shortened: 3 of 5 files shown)


adrienthebo:
Whoa, this is very neat. So the core of the idea is there will be an `environment` setting, and the values are pairs of `{'environment name' => {'ref' => 'ref to check out'}}`, more or less?

ryandoyle:
Hey Adrien

Yep, that is the general idea. The environment setting is completely optional and will just default to the head of the branch if none is passed.

brettswift:
Interesting..   a way of doing immutable environments!

I'm interested in how many different applications your puppet master supports, and what your release process is around this code?   We've been thinking along these lines as well.

The thought I'd have on release management here, would be an ENC or Hiera configuration of the puppet module that manages the Puppet Master.    The pairings would be in there, and you could just push up a new version of each environment.  IE  Test branch at 0.0.2, and prod at 0.0.1. 

Is that what you've done?  I'd like to hear more.. we're working with Puppet Enterprise and are thinking along the same lines.

brettswift:
Second question - would you expect this to work in combination of the multiple repositories example?

ryandoyle:
Hi Brett

> I'm interested in how many different applications your puppet master supports, and what your release process is around this code? We've been thinking along these lines as well.

The puppet master supports multiple applications but (currently) only using the single git repository. We are hoping to change this.

> The thought I'd have on release management here, would be an ENC or Hiera configuration of the puppet module that manages the Puppet Master....

We're wanting to keep release management as simple as possible. Our use case for this is to have a git tag 'tested' on our master branch. This way we can branch off master, and then merge this directly back into master without fear of breaking production. Instead of having a production and master branch, we just have the one - but it only gets into production when tagged as passing all automated tests (via Jenkins or some other CI tool).

> Second question - would you expect this to work in combination of the multiple repositories example?

No reason not to work with multiple repositories, the `environment` configuration is tied to the named-source

adrienthebo:
My initial impression of this is very positive; it looks like it solves the problem in a concise manner. I have a few minor thoughts about some of the implementation details but they're just that - minor. :)

Have you been able to test this functionality out? Does it have any sort of edge cases or concerns that users should be aware of? What sort of documentation should be added for this new feature?

ryandoyle:
Hi Adrien,

>  I have a few minor thoughts about some of the implementation details but they're just that - minor. :)

There is some re-factoring that I'd like to do as well. I was going to implement a cleaner way to build an `R10K::Deployment::Environment` (perhaps a factory that builds different types of environments) but I thought that could be left until there are more environment options. (I see this as the start of applying different options to different environments). If there are changes you'd like now, these can be done no problem.

> Have you been able to test this functionality out?

Not any more than testing locally and unit-level tests. Originally I was going to wait until this was merged upstream to start using it. You do make a good point though, if you would prefer it was more thoroughly tested before merged, I am happy to do that.

> Does it have any sort of edge cases or concerns that users should be aware of?

Not that I know of. A user can add environment locking and remove it at will.

> What sort of documentation should be added for this new feature?

I've added documentation in 54694f2. Is this enough? I tried to follow the same style - let me know if this needs less/more/rewording etc...

Thanks,
Ryan

adrienthebo:
> There is some re-factoring that I'd like to do as well. I was going to implement a cleaner way to build an R10K::Deployment::Environment (perhaps a factory that builds different types of environments) but I thought that could be left until there are more environment options. (I see this as the start of applying different options to different environments). If there are changes you'd like now, these can be done no problem.

What kind of different environments are you thinking of? This does need to be done, since I'm looking at implementing SVN support for environments (per GH-104) but haven't fully sorted out the details of how the refactoring will look. What sort of cases are you interested in supporting?

> Not any more than testing locally and unit-level tests. Originally I was going to wait until this was merged upstream to start using it. You do make a good point though, if you would prefer it was more thoroughly tested before merged, I am happy to do that.

I've recently discovered that it's really important for the original contributor to try out their use cases and run the examples they expect to need before things get merged. It shakes a lot of bugs out earlier in the process and it can be fixed more quickly. For instance, my r10k working directory is littered with test cases like this:

``` yaml
---
sources:
  test:
    remote: 'git@github.com:adrienthebo/r10k-fixture-repo'
    basedir: /home/adrien/development/r10k/testing/deploy-testing/puppetlabs-modules
```

This way I can just go into the right directory, run some deployments, and see what explodes. I would really appreciate it if you could test the scenarios you expect to use in your environment, and see if they work. :)

> > Does it have any sort of edge cases or concerns that users should be aware of?
> 
> Not that I know of. A user can add environment locking and remove it at will.

Cool. And this is exclusionary by default; ie only environments listed will be pulled in?

> > What sort of documentation should be added for this new feature?
> 
> I've added documentation in 54694f2. Is this enough? I tried to follow the same style - let me know if this needs less/more/rewording etc...

Perfect!

If you want to do the environment refactoring in this pull request I can leave this open. Alternately if this is something that can be deferred then I think it's best we do this. I prefer to have several smaller pull requests and bump version numbers a bit more often so that changes get out to users sooner, and there's definitely interest in this feature.

Thanks for all your work on this!

ryandoyle:
> > I was going to implement a cleaner way to build an R10K::Deployment::Environment (perhaps a factory that builds different types of environments) 
> 
> What kind of different environments are you thinking of? 

There are are two types of environments without this patch as I see it.
- A normal environment
- A prefixed environment

Prefixed environments are created by passing in different options to `R10K::Deployment::Environment`. 

EG in `R10K::Deployment::Source`

``` ruby
if @prefix
  R10K::Deployment::Environment.new(branch, @remote, @basedir, nil, @name.to_s())
else
  R10K::Deployment::Environment.new(branch, @remote, @basedir)
end
```

This patch changes that again, there are now 4 combinations of environments once you introduce an optional ref that the environment can be created from (as before it was always the branch name).

> I've recently discovered that it's really important for the original contributor to try out their use cases and run the examples they expect to need before things get merged.

Understandable. Ironically, I haven't had the time to do this yet. This work was done to help the team that is managing the Puppet deployment infrastructure. Hopefully I'll get a chance to deploy this over the next week or so.

> Cool. And this is exclusionary by default; ie only environments listed will be pulled in?

All environments will still be pulled in. They will use the head of the branch name for the ref unless otherwise specified (the same as before). 

> If you want to do the environment refactoring in this pull request I can leave this open. Alternately if this is something that can be deferred then I think it's best we do this.

Yeah I'd rather defer this. The other work I am doing in #107 touches these files as well (I have branched off master for that work, not this branch).

adrienthebo:
@ryandoyle to follow up on this issue, what's the status of this feature? Is there any additional work that needs to be done on this?

ryandoyle:
Hi Adrien, I've been on holidays from work for the last week so I haven't yet had a chance to put this in our production environment. The testing I have done has been manually QA (as well as unit tests of course). If you'd like I can come up with a more extensive test plan before it is merged in.

README.markdown
```diff
@@ -261,6 +261,21 @@ without being able to accidentally impact other groups.
...
+You may want to lock an environment to a specific git ref (such as a 'tested' tag). This can be specified in the
+yaml configuration file.
+
+        :cachedir: '/var/cache/r10k'
+        :sources:
+          # branch in /etc/puppet/environments
+          :plops:
+            remote: 'git@github.com:my-org/org-shared-modules'
+            basedir: '/etc/puppet/environments'
+            environment:
```
zachfi, line 275: Does this deploy on the specified environment, or does this deploy all environments, but treat this one special?

zachfi:
This works like a charm.  Setting the ref to a tag or sha both work as expected.  This only overrides the single branch while still deploying the rest of the remote's branches.

ryandoyle:
> Does this deploy on the specified environment, or does this deploy all environments, but treat this one special?

Hi, as you discovered it deploys all environments at the tip of that branch unless otherwise given a `ref` to use. Glad to see it worked out for you :)

adrienthebo:
@ryandoyle I wound up needing to heavily refactor how environments behave as part of GH-104. Because of this there are going to be some reasonably hefty rebases that need to be done to sync this branch and the other work done. However since your work came first I'll handle the necessary rebasing and fixups, but thought I would let you know in advance. :)

With respect to getting this released I need to get GH-104 finished up which will take about 4 - 8 more hours of work, get this synced up on top which will take about another hour, and then it'll probably need about 2 hours of time to update documentation. With all of this in mind I think this should be all wrapped up in about a week, depending on when I can make time for this work.

ryandoyle:
@adrienthebo thanks for the update, sounds good!

adrienthebo:
I've been mulling over these changes in light of the refactoring that I did for GH-104, and right now I'm not so sure about this approach. Right now the git environment source provides the current list of branches and their associated refs, which is pretty easy to reason about. If you have an environment, it will directly map to a branch and the ref used for that branch will be the latest one available. However with the pinning added in this pull, the 1:1 relationship between branches and environments gets a lot fuzzier. We still have to pin an environment as the same name as a branch, but the actual value of the environment is completely decoupled from the associated branch.

So in this problem space, we have a number of things we may want to achieve. We may want to track the latest version of all branches, which is the current behavior. We may want to track a subset of branches (which is discussed more in GH-129) but the values of those environments should track the associated branch. Finally we may want to have a specific list of environments that are pinned to an exact ref, which is different behavior from the first two cases.

Since there is a pretty big distinction between tracking branches and pinning environments to refs, I've been thinking about a different approach. What if specific environments could be declared entirely separately from sources? Perhaps something like this?

``` yaml
---
sources:
  remote: ...
  basedir: ...
environments:
  pinned-ref:
    type: git
    remote: 'git://git-server.site/git-repo.git'
    dirname: 'production'
    ref: 'faae79a3863cad964618b401357a2d3f098d39cd'
```

This provides the same functionality but no longer conflates environments being dynamically generated from git branches and environments being specifically created from refs. What do you think of this?

ryandoyle:
Hi Adrien,

It is true that is breaks that 1:1 mapping. I guess it helps to understand the original use case for us was the ability to have a way to run Puppet against a tested and blessed ref. We didn't want the extra overhead of having a `production` branch, so we thought that a tag of master would point to the latest "blessed" code was a good way to keep the simplicity.

Its up to you if you want to have a separate `environment` block although it starts to feel a bit clunky.

Perhaps having this configuration in the same `source` block makes more sense? It also reduces duplication that would come from having to configure multiple `remote`s and `basedir`s.

```
:sources:
  :plops:
    remote: '/home/ryan/workspace/r01ktest'
    basedir: '/var/tmp/puppet/environments'
    prefix: true
    static-environment:
      someenvironment: 
        ref: 'sometag'
      anotherenvironment:
        ref: 'faae79a38'
```

If there exists a branch called `someenvironment` or `anotherenvironment` then either the branch or the `static-environment` option would win (and log a suitable warning).

Also, don't feel obliged to implement this functionality. It does move away from the original idea that `r10k` was based off.

vchepkov:
I am interested in this feature as well or in something very similar.
I have a "base" git repository which defines core infrastructure and application teams have their repositories where they develop application specific manifests. We follow the same branching scheme in all repositories and I want to deploy 'development' branch of application modules in 'development' environment, 'testing' in 'testing' and so on.
ideally it would look something similar to this in Puppetfile, so it will be dynamic

```
mod 'app',
  :git    => 'https://stash/apps/app.git',
  :branch => ${environment}
```

But if I can just pin several environments with any other syntax it would be fine too.
I understand that I can have different versions of Puppetfile in each branch, but this would make fast-forward merges impossible. 

I tried to use an external script as a workaround:

```
mod 'app',
  :git    => 'https://stash/apps/app.git',
  :branch => `/usr/local/bin/puppet-branch`.split[0].chomp
```

unfortunately, r10k doesn't provide any indicators for the shell script to determine which environment is currently processed, so it didn't go anywhere. Would it be possible for r10k to set an environment variable before it processes Puppetfile, R10K_ENV, for example, until a more 'elegant' solution is implemented?

adrienthebo:
This functionality is necessary, but unfortunately this implementation has some fairly hefty implications. As mentioned in https://github.com/adrienthebo/r10k/pull/101#issuecomment-43167568 this breaks the 1:1 relationship between branches and environments.

On the topic of tags, I can see environments being generated from tags being implemented as a separate source from branches.  This way, you could do the following:

``` yaml
sources:
  branches:
    type: git
    remote: '...'
    basedir: '...'
  tags:
    type: git-tags
    remote: '...'
    basedir: '...'
    # And here would be some specification of what tags to create
```

For the case of individually pinning environments to commit I see something like this working:

``` yaml
sources:
  branches:
    type: git
    remote: '...'
    basedir: '...'

environments:
  some_environment_name:
    type: git
    basedir: '/etc/puppetlabs/puppet/environments'
    ref: '1.2.7' # could be a branch, or tag, or commit SHA
```

This behavior allows environments to be deployed from all tags automatically, or single environments to be created and locked to a single commit. We get all the functionality desired but without breaking some of the existing concepts in r10k.

I'm still working on speccing out the final behavior for git tags. If anyone has opinions please comment on GH-135 so that I can determine the final goals that people have.

Because of these revised design goals, I'm going to close this issue. I apologize for the delay in responding to this and that I couldn't get this merged in, and I really appreciate all the work that you put into this.

Issue #97, opened 2014-03-14
r10k puppetfile install clones git repos in a detached head state


natewalck:
When using 'r10k puppetfile install', it clones all the repos down in a detached head state. I would expect the modules to be checked out to the default branch and be able to  check out additional branches (if they exist).

adrienthebo:
Unfortunately this issue is dramatically more complex than it may initially seem. Checking out a commit or tag is generally consistent as HEAD just points to a commit, but checking out a branch means initially creating a new branch and then later doing a pull on that branch. I started on an implementation of this but it got complex and messy very quickly.

In addition, supporting this use case directly conflicts with #47. Git is extremely bad at checking out only parts of a repository; sparse checkouts exist but do not serve to shift the desired subdirectory to the root of the repository. The only way to fake this out is to either do a `git subtree merge` which is deeply flawed for many reasons, or a `git filter-branch` to rewrite all of history, which is equally flawed for different reasons. In either case, these scenarios require drastically altering history which makes the repository worthless for further development. The best approach to this is to stop treating the checked out module as a Git repository and extract/copy the subdirectory into place.

In any case, all of the possible solutions horribly mangle the Git repository in question which makes them unsuitable for use further development. Trying to support actual branch checkouts per this issue and #47 together would require writing monstrous code that would look like it was copied from unspeakably terrible hieroglyphic murals sourced from R'lyeh.

While it would be convenient to use r10k to check out repositories for development purposes, unfortunately it's just not practical to do so; the cost of trying to make this work is enormously greater than asking users to `git clone` repositories. Git repository handling is going to move further away from having checkouts that are usable, which means that there won't be a current checkout at all, let along a checked out branch.

c33s:
for development it would still be very very helpful. in fact i would even need the conflicting #47
if i mark a repo as "developent: true" it would be ok, to checkout the whole repo and then switch branch, additionally adding a path would be great, so i can control that whole repo modules reside in an own directory like modules_dev.

c33s:
voting for reopen.

```
r10k puppetfile install --full-clones
```
just as a helper for iterating over all repos defined in the puppetfile and do a normal `git clone` or additional have a `:development => true` flag for each module
```
mod 'abstractit-puppet',
     :git => 'https://github.com/abstractitptyltd/abstractit-puppet.git',
     :install_path => 'modules/fork',
     :commit => 'c105b45fa7c3786163e8c491fe22de980c0534f8'
     :full_clone => true
```
[...]

Issue #92, opened 2014-02-16
[Enhancement] Support mod 'owner/module' for git-based modules


ferventcoder:
This behavior works with librarian-puppet:

``` ruby
mod 'rismoney/chocolatey',
  :git => 'git://github.com/chocolatey/puppet-chocolatey.git'
```

But doesn't work with r10k:

```
skylight$ r10k -v DEBUG2 puppetfile install
[R10K::Task::Puppetfile::Sync - INFO] Loading modules from Puppetfile into queue
[R10K::Task::Module::Sync - INFO] Deploying rismoney/chocolatey into /Users/rob/code/mine/server/modules
[R10K::Git::Cache - DEBUG1] Execute: "git fetch --prune, args: {:git_dir=>\"/Users/rob/.r10k/git/git---github.com-chocolatey-puppet-chocolatey.git\"}"
[R10K::Git::WorkingDir - DEBUG1] Execute: "git clone --reference /Users/rob/.r10k/git/git---github.com-chocolatey-puppet-chocolatey.git git://github.com/chocolatey/puppet-chocolatey.git /Users/rob/code/mine/server/modules/rismoney/chocolatey"
[R10K::Git::WorkingDir - DEBUG2] [git clone --reference /Users/rob/.r10k/git/git---github.com-chocolatey-puppet-chocolatey.git git://github.com/chocolatey/puppet-chocolatey.git /Users/rob/code/mine/server/modules/rismoney/chocolatey] STDERR: Cloning into '/Users/rob/code/mine/server/modules/rismoney/chocolatey'...
[R10K::Git::WorkingDir - DEBUG1] Execute: "git remote add cache /Users/rob/.r10k/git/git---github.com-chocolatey-puppet-chocolatey.git, args: {:path=>\"/Users/rob/code/mine/server/modules/rismoney/chocolatey\"}"
[R10K::Git::WorkingDir - DEBUG1] Execute: "git checkout master, args: {:path=>\"/Users/rob/code/mine/server/modules/rismoney/chocolatey\"}"
[R10K::Git::WorkingDir - DEBUG2] [git checkout master, args: {:path=>"/Users/rob/code/mine/server/modules/rismoney/chocolatey"}] STDOUT: Your branch is up-to-date with 'origin/master'.
[R10K::Git::WorkingDir - DEBUG2] [git checkout master, args: {:path=>"/Users/rob/code/mine/server/modules/rismoney/chocolatey"}] STDERR: Already on 'master'
[...]

Issue #89, opened 2014-02-04
Can r10k deal with dependency in Modulefile?


rvora:
It looks like r10k ignores dependencies of modules listed in Puppetfile. Is this the expected behavior?

If yes, then r10k is not a drop-in-replacement for librarian-puppet and in case is it possible to use librarian-puppet when 'r10k deploy environment -p' is called?

rvora:
I see --ignore-dependencies being used in lib/r10k/module/forge.rb. So that explains why module's dependencies are ignored by 'puppet module install'. I guess I would like to know why --ignore-dependecies was chosen as the default and only behavior and how should one figure out all the dependencies without going through trial-and-error iterations.

adrienthebo:
- Is r10k a drop-in replacement for librarian-puppet?

Unfortunately no, not yet. Since the README was misleading about this I've updated it accordingly: https://github.com/adrienthebo/r10k/commit/29f9471c04e7f91b9d0e8f2243a71a25664738cf
- Why are forge module dependencies ignored?

A forge module may bring in a dependency that conflicts with either another forge module or a git module, which could create some very strange interactions. Since modules installed from Git can't install module dependencies it's more consistent to not handle dependencies rather than do it for some modules and risk conflicts.
- Can librarian-puppet be used when `r10k deploy environment -p` is  called?

Theoretically it's possible. I've looked at some of the librarian API and it looked like it wouldn't be very straightforward to do, but if someone is willing to look into handling this then I would be thrilled.
- How can dependencies be  resolved for a Puppetfile?

I believe that you can use librarian-puppet's lockfile to synthesize this information, although it may require a bit of manual work.
- When will dependency resolution be implemented?

It depends on a number of things. First off there's a library for dependency resolution that should be slotted in and can provide dependency resolution in one pass, but I'm waiting for it to be released so it can be depended. It also depends on how much free time I have - I'm not paid to work on r10k so I have to work on it on my evenings and weekends. While I understand that it's painful to not have this feature the reality is that I simply don't have the resources to implement this by myself in the immediate future.

rvora:
Thanks Adrien for the detailed response. It makes sense.

Before anything else, thanks very much for your generous work on r10k. r10k makes it easy to implement dynamic environments - the way all new puppet installs should be done in my view.

I would think that librarian-puppet will also fail on the kind of conflicting transient dependency problems you are talking about though I have not tried it out to verify.

[...]

Issue #78, opened 2014-01-10
R10K doesn't ensure module names are valid


pzim:
if Puppetfile has mod references with the org in the name, it appears to run successfully, but then prunes the modules dir when finished. It would be helpful to have it check mod names for validity prior to attempting to clone their repositories.

Example Puppetfile entries to reproduce:

mod 'puppetlabs/postgresql',
  :git => "https://github.com/puppetlabs/puppetlabs-postgresql",
  :ref => '2.5.0'
mod 'puppetlabs/puppetdb',
  :git => "https://github.com/puppetlabs/puppetlabs-puppetdb",
  :ref => '2.0.0'

blbradley:
I think this has to do with how r10k takes the name from the 'mod' statement in the Puppetfile. It works if you remove the author name from 'mod' when using a git source.

Ex:

```
mod 'postgres',
  :git => 'https://github.com/puppetlabs/puppetlabs-postgresql',
  :ref => '2.5.0'
```

Also, since r10k doesn't resolve module dependencies, you will need to define and resolve them manually in the Puppetfile.

adrienthebo:
The original issue with pruning Git modules that included the owner in the module name should be resolved as part of GH-92. That being said we should validate the rest of the modulename and ensure it matches `[\w_]`.

adrienthebo:
This may have been addressed in 865a0212e as part of GH-92.

adrienthebo:
Verified in 4ac4f76 as working:

```
└> cat Puppetfile
mod 'puppetlabs/postgresql',
:git => "https://github.com/puppetlabs/puppetlabs-postgresql",
:ref => '2.5.0'
mod 'puppetlabs/puppetdb',
:git => "https://github.com/puppetlabs/puppetlabs-puppetdb",
:ref => '2.0.0'
└> bexec r10k puppetfile install
└> ls modules 
postgresql  puppetdb
```

erinn:
Is this module name validation correct? Is there any guide on naming conventions? The regex doesn't allow for underscores or hyphens. In the https://docs.puppetlabs.com/puppet/latest/reference/modules_fundamentals.html guide it references create my_module, this wouldn't work with R10k as it is. I have been digging around trying to figure out what is what, but I figured I would ask.

elyscape:
[...]

PR #107, opened 2014-04-22
REVIEW: Librarian-puppet integration

ryandoyle:
Hey

I was considering working on replacing r10k's own Puppetfile management with `librarian-puppet`. I have been poking around librarian-puppet to see how hard it would be to integrate. From what I can see so far - it doesn't look /too/ bad.

If this were done, it would also close off a number of issues.
- #89 
- #97 
- #92 
- #89 
- #78 
- #47 
- #50 

And maybe more (they are all I could see). Adrien/others, what do you think? It feels this is the right direction to take.

Cheers!

Update: I've been able to replace `r10k puppetfile install` with a native librarian-puppet call which is working quite well.


ryandoyle:
Quick update: It's coming together quite well. The internal provider and Librarian now implement the same interface (configration of which one to use is in the code at the moment). I'm now up to driving the decision of which provider based on some option in the `r10k.yaml` file.

For those playing along at home, this is the branch I am pushing to: https://github.com/ryandoyle/r10k/compare/librarian-puppet-integration

adrienthebo:
@ryandoyle if you like you can file a pull request against r10k and add REVIEW in the title if you would like running feedback on it, that's what we do in Puppet land.

ryandoyle:
@adrienthebo cheers - I've converted this into a pull-request.

r10k.gemspec
```diff
@@ -32,6 +32,8 @@ Gem::Specification.new do |s|
   s.add_dependency 'faraday',                       '~> 0.8.8'
   s.add_dependency 'faraday_middleware',            '~> 0.9.0'
   s.add_dependency 'faraday_middleware-multi_json', '~> 0.0.5'
+  s.add_dependency 'librarian-puppet',              '~> 0.9.15'
+  s.add_dependency 'puppet'
```
nanliu: There are some history worth reviewing regarding why this should only be a development dependency: https://github.com/rodjek/librarian-puppet/pull/62

adrienthebo: I actually don't want to add hard dependencies on librarian-puppet or puppet. If Puppet is installed via a package this dependency will drag in another copy via gems, which can introduce nasty conflicts with which provider is used. 

In addition I want librarian-puppet to be an optional dependency; the functionality should be there but it isn't required for all operations. Puppet itself has gating around optional dependencies (for example [msgpack](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/feature/msgpack.rb)), and I think we should do the same thing. If librarian-puppet is selected as the Puppetfile implementation we should ensure the gem is installed at runtime and fail if it's not installed.

lib/r10k/puppetfile_provider/factory.rb
```diff
@@ -0,0 +1,16 @@
+require 'r10k/puppetfile_provider/internal'
+require 'r10k/puppetfile_provider/librarian_puppet'
+
+module R10K
+module PuppetfileProvider
+class Factory
+
+  def self.driver(basedir, moduledir = nil, puppetfile_path = nil)
+    Internal.new(basedir, moduledir, puppetfile_path)
+    # TODO: Implement a way to swap between providers
```
adrienthebo: This will require a better way to initialize application state; I have some ideas on this but it might need a bit more work.

lib/r10k/puppetfile_provider/librarian_puppet.rb
```diff
@@ -0,0 +1,77 @@
...
+    private
+
+    def environment
+      @environment ||= Librarian::Puppet::Environment.new(:pwd => @basedir)
+    end
+
+    def while_destructive
+      environment.config_db.local['destructive'] = 'true'
+      yield if block_given?
+      ensure
```
adrienthebo: Style comment: method level ensures should be outdented to the same level as the method definition.

lib/r10k/task/deployment.rb
```diff
@@ -136,7 +136,6 @@ def call
 
         if @puppetfile
           puppetfile = env.puppetfile
-          puppetfile.load
```
adrienthebo, line 139: I'm guessing that the explicit `#load` is no longer necessary; is loading done lazily when invoking `puppetfile.modules`?

lib/r10k/task/puppetfile.rb
```diff
@@ -7,21 +7,13 @@ module R10K
...
       logger.info "Loading modules from Puppetfile into queue"
-
-      @puppetfile.load
-      @puppetfile.modules.each do |mod|
-        task = R10K::Task::Module::Sync.new(mod)
-        task_runner.insert_task_after(self, task)
-      end
-
-      purge_task = Purge.new(@puppetfile)
-      task_runner.append_task purge_task
```
adrienthebo, line 24: Is it safe to assume that with these deletions, the Module level tasks are no longer needed?

spec/unit/deployment/environment_spec.rb
```diff
@@ -21,4 +21,31 @@
       subject.dirname.should == 'sourcename_master'
     end
   end
+
+  describe '#sync' do
```
adrienthebo, line 25: Test coverage! \o/

spec/unit/puppetfile_provider/librarian_puppet_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        expect(librarian).to receive(:sync)
+        librarian.purge
+        expect(environment.config_db.local['destructive']).to eql 'true'
+      end
+    end
+    describe 'when Librarian is not destructive' do
+
+      before :each do
+        environment.config_db.local['destructive'] = 'false'
+        expect(librarian).to receive(:while_destructive).and_call_original
```
adrienthebo, line 59: Will these tests actually touch the system?

adrienthebo:
I'm generally pretty happy with this; there are a few things that I would like to discuss further but all in all, great work!

ryandoyle:
> actually don't want to add hard dependencies on librarian-puppet or puppet

No problem. I can remove the dependency (and rather add as a development dependency). I will implement some exception handling when Librarian is used and it is not installed.

> This will require a better way to initialize application state; I have some ideas on this but it might need a bit more work.

Yeah, I'm still somewhat stuck when it comes to configuring which driver to use. Ideally it would be configured in the `r10k.yaml` file. The only problem with this is `r10k puppetfile` commands do not require an `r10k.yaml` file.

I am hesitant to use a command line option as the provider isn't something that can be swapped easily. That said,  it might be enough for the initial implementation.

> Style comment: method level ensures should be outdented to the same level as the method definition.

Fixed

> Is it safe to assume that with these deletions, the Module level tasks are no longer needed?

Good find. Yeah the module-level  tasks are no longer used. They are implemented in the Puppetfile providers and do not use the task system. I've pushed a deletion of this file.

> Will these tests actually touch the system?

Nup, the `config_db` is implemented as a double.

ryandoyle:
> > actually don't want to add hard dependencies on librarian-puppet or puppet
> 
> No problem. I can remove the dependency (and rather add as a development dependency). I will implement some exception handling when Librarian is used and it is not installed.

This is now done. I've moved the gem specifications to the Gemfile in the `development` group. Additionally, the gems are only loaded if `Librarian-puppet` is chosen as a provider. 

Currently the Travis CI tests are failing for 2.1 and 1.8.7 but these look to be an intermittent problem with the Travis CI infrastructure.

ryandoyle:
@adrienthebo what is your thinking on choosing a provider? 

There are two modes as I see it.
- `r10k deploy` commands - This requires an `r10k.yaml` file so it makes sense to configure the provider in there
- `r10k puppetfile` commands - This does not use nor require static configuration in its current state.

I was /thinking/ along the lines of:
- `r10k deploy` commands can use the provider configured in `r10k.yaml`, the internal provider being a default if none is provided.
- `r10k puppetfile` commands require an option argument. EG: `r10k puppetfile install --puppetfile-provider=librarianpuppet`, again with the internal provider being the default.

Thoughts?

ryandoyle:
I've implemented selecting the Puppetfile provider for `r10k deploy` commands by being able to specify it in the `r10k.yaml` file using the setting `puppetfileprovider`. EG

```
:puppetfileprovider: 'librarian'
```

_or_

```
:puppetfileprovider: 'internal'
```

If nothing is configured, it will use the internal Puppetfile driver.

adrienthebo:
Unfortunately I've been struck down with some pretty major RSI induced pain, so I'm not going to be able to do much review and feedback on this in the immediate future. I'll update this when my hands have healed more and I can be more engaged in this pull request.

ryandoyle:
Sorry to hear that @adrienthebo. We'll continue when you're better : )

Cheers

fatmcgav:
Is this PR usable? 

cc @adrienthebo  @ryandoyle

ryandoyle:
Hi @fatmcgav, this PR would need to be re-worked again as it doesn't merge cleanly any more. @adrienthebo is this something that would still be considered for merging?

adrienthebo:
Unfortunately, I've been avoiding this pull request because I haven't wanted to close it, but I haven't been in a position to merge it. I would really like to incorporate this sort of inclusion but right now we don't have the ability to include these changes and maintain them. I should have been more up front about this rather than letting pull requests sit so long; I'm sorry about that. In the future I would like to revisit this integration, but since I can't merge this right now and it would require a bigger rework to include, I'm going to close it out. Again - sorry that I couldn't merge this and sorry about how long it take for me to actually close this.

PR #128, opened 2014-05-12
Fix backtick and URL to render properly on github.

DavidS:


adrienthebo:
Thanks for the contribution!

PR #130, opened 2014-05-13
Add $basedir to $moduledir if not absolute path (solves #44)

naevtamarkus:
Having new directories created to $CMD does not seem to have any practical use... so this should not break backwards compatibility.
It solves #44 very simply.

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -59,7 +59,11 @@ class Puppetfile
 
   # @param [String] moduledir
   def set_moduledir(moduledir)
-    @moduledir = moduledir
+    if moduledir[0] != 47   # ASCII for '/' (adds $basedir if not absolute path)
+      @moduledir = File.join(@basedir, moduledir)
+    else
+      @moduledir = moduledir
+    end
   end
 
   # @param [String] name


lib/r10k/puppetfile.rb
```diff
@@ -59,7 +59,11 @@ def set_forge(forge)
 
   # @param [String] moduledir
   def set_moduledir(moduledir)
-    @moduledir = moduledir
+    if moduledir[0] != 47   # ASCII for '/' (adds $basedir if not absolute path)
```
adrienthebo, line 62: This needs to include validation that the given path is only a directory name and isn't a relative path. For instance, this would accept `moduledir "foo/bar/modules"` but the underlying code would fail.

naevtamarkus, line 62: Not really, maybe I did not explain myself correctly:
- If the moduledir starts with a slash, it is an absolute path, and 
  therefore we don't change anything, to avoid breaking backwards 
  compatibility.
- If the moduledir does not start with a slash, it is a relative path 
  and therefore $basedir needs to be appended, whether the relative path 
  is simple, or composed by various subdirectories.

Hope this clarifies it... or what kind of validation were you talking about?

adrienthebo:
I've merged in GH-141 which implements this in a different manner, sorry about the duplication!

PR #132, opened 2014-05-14
Feature/extra parameters

naevtamarkus:
Hi,

This adds two optional parameters (puppetfile and subdir) to each source inside /etc/r10k.yaml, that used to be pre-defined constants (Puppetfile and modules, respectively). This adds a lot of flexibility by allowing other non-modules repositories to be deployed with r10k as well (like hieradata, roles, etc) 

For example:

```
:sources:
  # This will clone the git repository and instantiate an environment per
  # branch in /etc/puppet/environments
  :moria_modules:
    remote: 'xxx.xx:/git/puppet/config/moria-site.git'
    basedir: '/root/environments'
    puppetfile: 'Puppetfile.modules'
    subdir: 'modules'

  :moria_roles:
    remote: 'xxx.xx:/git/puppet/config/moria-site.git'
    basedir: '/root/environments'
    puppetfile: 'Puppetfile.roles'
    subdir: 'roles'

  :moria_hieradata:
    remote: 'xxx.xx:/git/puppet/config/moria-site.git'
    basedir: '/root/environments'
    puppetfile: 'Puppetfile.hieradata'
    subdir: 'hieradata'
```

This would take each source and deploy, based on different Puppetfiles (even in the same repo), in a separate subdirectory (whose parent is $basedir/$env), like this:

```
[root@puppet01 production]# pwd; ls
/root/environments/production
hieradata/  modules/  roles/  Puppetfile.hieradata  Puppetfile.modules  Puppetfile.roles  
```

(where each Puppetfile is a simple set of "mod" lines)

This feature has the potential to allow other non-puppet uses for r10k, like software deployment. It is also backwards compatible. It proposed it for 1.3.x, but it works for 1.2.x as well.

I am sorry, but I am not familiar with the rspec test ruby development... so I could not contribute to that part.

Hope you like it!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -265,6 +265,36 @@ without being able to accidentally impact other groups.
         remote: 'git@github.com:my-org/app2-environments'
         basedir: '/etc/puppet/environments'
 
+
+### Single Environment Repository with multiple Puppetfiles
+
+It is also possible to extend the use of environments beyond the modules themselves, by creating multiple sources that point to the same remote, each pulling from a different `Puppetfile` and deploying to a different subdirectory, by using the `puppetfile:` and the `subdir:` parameters. This is specially interesting for those cases when you need to sync groups of repositories like site manifests, hieradata, roles or profiles.
+
+For example:
+
+    :sources:
+      # This will clone the git repository and instantiate an environment per
+      # branch in /etc/puppet/environments
+      :modules:
+        remote: 'git@github.com:my-org/site-environments'
+        basedir: '/etc/puppet/environments'
+      :manifests:
+        remote: 'git@github.com:my-org/site-environments'
+        basedir: '/etc/puppet/environments'
+        puppetfile: 'Puppetfile.manifests'
+        subdir: 'manifests'
+      :hieradata:
+        remote: 'git@github.com:my-org/site-environments'
+        basedir: '/etc/puppet/environments'
+        puppetfile: 'Puppetfile.hiera'
+        subdir: 'hieradata'
+
+The Environment Repository (site-environments) would need to have at least three files: Puppetfile (default), Puppetfile.manifests and Puppetfile.hieradata. All those three $puppetfile files would have the same format, but each would be deployed in a different subdirectory below the environment name, with the shape $basedir/<env_branch>/$subdir:
+- git repositories defined in Puppetfile (branch X) would be deployed in /etc/puppet/environments/X/modules (default)
+- git repositories defined in Puppetfile.manifests (branch X) would be deployed in /etc/puppet/environments/X/manifests
+- git repositories defined in Puppetfile.hiera (branch X) would be deployed in /etc/puppet/environments/X/hieradata
+
+
 More information
 ----------------
 
--- a/README.markdown
+++ b/README.markdown
@@ -289,7 +289,7 @@ For example:
         puppetfile: 'Puppetfile.hiera'
         subdir: 'hieradata'
 
-The Environment Repository (site-environments) would need to have at least three files: Puppetfile (default), Puppetfile.manifests and Puppetfile.hieradata. All those three $puppetfile files would have the same format, but each would be deployed in a different subdirectory below the environment name, with the shape $basedir/<env_branch>/$subdir:
+The Environment Repository (site-environments) would need to have at least three files: Puppetfile (default), Puppetfile.manifests and Puppetfile.hiera. All those three $puppetfile files would have the same format, but each would be deployed in a different subdirectory below the environment name, with the shape $basedir/<env_branch>/$subdir:
 - git repositories defined in Puppetfile (branch X) would be deployed in /etc/puppet/environments/X/modules (default)
 - git repositories defined in Puppetfile.manifests (branch X) would be deployed in /etc/puppet/environments/X/manifests
 - git repositories defined in Puppetfile.hiera (branch X) would be deployed in /etc/puppet/environments/X/hieradata
--- a/lib/r10k/deployment/environment.rb
+++ b/lib/r10k/deployment/environment.rb
@@ -28,13 +28,15 @@ class Environment
   # @param [String] basedir
   # @param [String] dirname The directory to clone the root into, defaults to ref
   # @param [String] source_name An additional string which may be used with ref to build dirname
-  def initialize(ref, remote, basedir, dirname = nil, source_name = "")
+  def initialize(ref, remote, basedir, dirname = nil, source_name = "", puppetfile_file = nil, subdir = nil)
     @ref     = ref
     @remote  = remote
     @basedir = basedir
     alternate_name =  source_name.empty? ? ref : source_name + "_" + ref
     @dirname = sanitize_dirname(dirname || alternate_name)
-
+    @puppetfile_file = puppetfile_file
+    @subdir = subdir
+    
     @working_dir = R10K::Git::WorkingDir.new(@ref, @remote, @basedir, @dirname)
 
     @full_path = File.join(@basedir, @dirname)
@@ -57,7 +59,7 @@ class Environment
   end
 
   def puppetfile
-    @puppetfile ||= R10K::Puppetfile.new(@full_path)
+    @puppetfile ||= R10K::Puppetfile.new(@full_path, @subdir, @puppetfile_file)
   end
 
   def modules
--- a/lib/r10k/deployment/source.rb
+++ b/lib/r10k/deployment/source.rb
@@ -45,17 +45,21 @@ class Source
   def self.vivify(name, attrs, prefix = false)
     remote  = (attrs.delete(:remote) || attrs.delete('remote'))
     basedir = (attrs.delete(:basedir) || attrs.delete('basedir'))
+    subdir = (attrs.delete(:subdir) || attrs.delete('subdir'))
+    puppetfile_file = (attrs.delete(:puppetfile) || attrs.delete('puppetfile'))
     prefix_config = (attrs.delete(:prefix) || attrs.delete('prefix'))
     prefix_outcome = prefix_config.nil? ? prefix : prefix_config
 
     raise ArgumentError, "Unrecognized attributes for #{self.name}: #{attrs.inspect}" unless attrs.empty?
-    new(name, remote, basedir, prefix_outcome)
+    new(name, remote, basedir, prefix_outcome, puppetfile_file, subdir)
   end
 
-  def initialize(name, remote, basedir, prefix = nil)
+  def initialize(name, remote, basedir, prefix = nil, puppetfile_file = nil, subdir=nil)
     @name    = name
     @remote  = remote
     @basedir = basedir
+    @subdir = subdir || 'modules'
+    @puppetfile_file = puppetfile_file || 'Puppetfile'
     @prefix = prefix.nil? ? false : prefix
 
     @cache   = R10K::Git::Cache.generate(@remote)
@@ -97,9 +101,9 @@ class Source
     if @cache.cached?
       @environments = @cache.branches.map do |branch|
         if @prefix
-          R10K::Deployment::Environment.new(branch, @remote, @basedir, nil, @name.to_s())
+          R10K::Deployment::Environment.new(branch, @remote, @basedir, nil, @name.to_s(), @puppetfile_file, @subdir)
         else
-          R10K::Deployment::Environment.new(branch, @remote, @basedir)
+          R10K::Deployment::Environment.new(branch, @remote, @basedir, nil, "", @puppetfile_file, @subdir)
         end
       end
     else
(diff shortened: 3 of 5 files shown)


adrienthebo:
This introduces two new features, the ability to specify an alternate name for Puppetfiles and the ability to specify a different directory for the modules that the Puppetfile installs. I'm going to address those separately.

First off, the Puppetfile as-is is built do pull in external module dependencies. It is not meant to manage hiera data or any other data as that overloads the purpose of the Puppetfile. While I am open to adding the feature for an environment to specify additional hiera data I want to see the intended behavior and outcomes well specified before the code is implemented. In addition it is possible to specify an alternate hiera directory as is outlined in https://github.com/adrienthebo/r10k/blob/master/doc/dynamic-environments/configuration.mkd#separate-hiera-data , so I would like to know why that functionality cannot be used instead.

Secondly, I'm not clear on what the value of having multiple Puppetfiles is. Is the value the ability to deploy different modules into different directories? If so, the separation is entirely logical as the the modulepath directive will effectively flatten this directory structure when reading modules. If logical separation is desired that can be done in pure Ruby without adding more behavior to r10k.

In addition, the example configuration file is relying on undefined behavior; if multiple environments with the same name are deployed into a single directory there are no guarantees about what will happen. In the future if environments will collide like this, r10k will treat that as a hard error and will not update any environments.

With respect to the subdir directive, how does that differ from the 'moduledir' Puppetfile directive?

Finally, this adds new features to r10k that adds flexibility, but that comes with a cost. New features require testing, validation, and documentation, and the more configurable a tool becomes the harder it becomes to understand. Can the value of these new features justify the cost of adding and maintaining them?

naevtamarkus:
Hmm you are talking about many things, so let me try to address those 
one by one:

> This introduces two new features, the ability to specify an alternate 
> name for Puppetfiles and the ability to specify a different directory 
> for the modules that the Puppetfile installs. I'm going to address 
> those separately.

Fair enough. I started with one and the other followed very easily.

> First off, the Puppetfile as-is is built do pull in external module 
> dependencies. It is not meant to manage hiera data or any other data 
> as that overloads the purpose of the Puppetfile. While I am open to 
> adding the feature for an environment to specify additional hiera data 
> I want to see the intended behavior and outcomes well specified before 
> the code is implemented
> 
> Agree on its initial purpose... but this is a classic artificial 
> restriction: the author did not think about it at first. What the 
> Puppetfile ACTUALLY does is "Picking up a remote git repo and placing it 
> in a directory", and it happens that he/she meant it ONLY for puppet 
> modules. What a pity!!

But I want to believe we want to make good use of the tools we have: if 
this Puppetfile has much more potential, why not using it? I am not 
changing the syntax of the Puppetfile itself: I am extending r10k so 
that it can much more with it, without breaking backwards compatibility.

> In addition it is possible to specify an alternate hiera directory as 
> is outlined in 
> https://github.com/adrienthebo/r10k/blob/master/doc/dynamic-environments/configuration.mkd#separate-hiera-data 
> , so I would like to know why that functionality cannot be used instead.

That function is not related to the features I am refering to: it may be 
useful for creating extra environments, but not for having repositories 
of many types within the same environment.

> Secondly, I'm not clear on what the value of having multiple 
> Puppetfiles is. Is the value the ability to deploy different modules 
> into different directories? If so, the separation is entirely logical 
> as the the modulepath directive will effectively flatten this 
> directory structure when reading modules. If logical separation is 
> desired that can be done in pure Ruby without adding more behavior to 
> r10k.

Modulepath is only for modules... and I agree, all modules will 
"flatten" at the end. But inside a Puppet environment there are more 
things than modules: you have site manifests, and you have hiera data. I 
need a tool to manage the environment, and not only modules. If there is 
a great tool for modules and I still have to develop my git deployment 
bash scripts for hieradata and node classification... why do I need the 
great module tool anyway?

There is a workaround to this: having the node classification and 
hieradata inside the same git repo, together with the Puppetfile. But in 
big environments (like ours will be) you cannot have all this inside one 
repo. Just to give you an example: part of the hieradata is coming from 
one group (site-wide config) and part from another group 
(cluster-specific config), with different permissions, release schedules 
and all.

> In addition, the example configuration file is relying on undefined 
> behavior; if multiple environments with the same name are deployed 
> into a single directory there are no guarantees about what will 
> happen. In the future if environments will collide like this, r10k 
> will treat that as a hard error and will not update any environments.

I used a similar example like the one you have in the README.md (your 
last one, called Multiple Environment Repositories Example), with the 
exception that I can have multiple Puppetfiles per remote and you can 
only have one, so you need to use two different remotes... but in both 
we have the same basedir. Why do you say this will break in the future? 
It works now...

> With respect to the subdir directive, how does that differ from the 
> 'moduledir' Puppetfile directive?

Unfortunately, the moduledir feature is slightly broken: you can only 
have a single moduledir within the Puppetfile. I tried having multiple 
ones (trying to divide the modules from the hieradata, for example) and 
the purge wipes all moduledirs but the last one. And moduledir is 
actually NOT part of the original Puppetfile format... it is an ugly 
hack, if you allow the expression.

The subdir feature I propose, the "alternative directory", does exactly 
that, but the right way: in the r10k.yaml file you should specify where 
is the source of the Puppetfile (remote:) and where to deploy its 
contents (basedir: and subdir:). And Puppetfile is just a simple list of 
git repos, like it always was.

> Finally, this adds new features to r10k that adds flexibility, but 
> that comes with a cost. New features require testing, validation, and 
> documentation, and the more configurable a tool becomes the harder it 
> becomes to understand. Can the value of these new features justify the 
> cost of adding and maintaining them?

I understand. I can offer to write the documentation, but unfortunately 
I don't have the knowledge to write tests. I am good at reading code (in 
any language) and smart enough to copy-paste and/or modify those things 
that are logical to me. Would you mind telling me how to run the tests 
myself? Then I could see what breaks, and fix the tests accordingly by a 
try-and-error approach. Then, testing the new features should not be 
that difficult, I hope.

For me, these two features mean a great deal... I am one of those that 
think that either I do it right or I don't do it at all. I think I got 
far enough to have two very nice features that would make my life, and 
hopefully others, a better one... there must not be so much more until 
it's documented and tested.

BR/Pablo

naevtamarkus:
Updated documentation.

I still treated both parameters as a single feature, because they come from the same concept/idea, and for laziness.

adrienthebo:
> > First off, the Puppetfile as-is is built do pull in external module
> > dependencies. It is not meant to manage hiera data or any other data
> > as that overloads the purpose of the Puppetfile. While I am open to
> > adding the feature for an environment to specify additional hiera data
> > I want to see the intended behavior and outcomes well specified before
> > the code is implemented.
> 
> Agree on its initial purpose... but this is a classic artificial
> restriction: the author did not think about it at first. What the
> Puppetfile ACTUALLY does is "Picking up a remote git repo and placing it
> in a directory", and it happens that he/she meant it ONLY for puppet
> modules. What a pity!!
> 
> But I want to believe we want to make good use of the tools we have: if
> this Puppetfile has much more potential, why not using it? I am not
> changing the syntax of the Puppetfile itself: I am extending r10k so
> that it can much more with it, without breaking backwards compatibility.

It's entirely possible to put more than modules into a Puppetfile. That does not mean that this is a good solution, that it can be taught effectively, or that there's a better solution. I would rather have well implemented features that do one thing well rather than try to provide for every possible use case because it's possible. Combining hiera data into an environment is a legitimate need but it should be treated as hiera data and not a module.

> > In addition it is possible to specify an alternate hiera directory as
> > is outlined in
> > https://github.com/adrienthebo/r10k/blob/master/doc/dynamic-environments/configuration.mkd#separate-hiera-data
> > , so I would like to know why that functionality cannot be used instead.
> > That function is not related to the features I am refering to: it may be
> > useful for creating extra environments, but not for having repositories
> > of many types within the same environment.

The example doesn't make more environments; it allows you to create a single environment with different source repositories. A given environment would use modules from `/etc/puppet/environments/$environment` and it would use hiera data from `/etc/puppet/hiera/$environment`.

> > Secondly, I'm not clear on what the value of having multiple
> > Puppetfiles is. Is the value the ability to deploy different modules
> > into different directories? If so, the separation is entirely logical
> > as the the modulepath directive will effectively flatten this
> > directory structure when reading modules. If logical separation is
> > desired that can be done in pure Ruby without adding more behavior to
> > r10k.
> > Modulepath is only for modules... and I agree, all modules will
> > "flatten" at the end. But inside a Puppet environment there are more
> > things than modules: you have site manifests, and you have hiera data. I
> > need a tool to manage the environment, and not only modules. If there is
> > a great tool for modules and I still have to develop my git deployment
> > bash scripts for hieradata and node classification... why do I need the
> > great module tool anyway?
> 
> There is a workaround to this: having the node classification and
> hieradata inside the same git repo, together with the Puppetfile. But in
> big environments (like ours will be) you cannot have all this inside one
> repo. Just to give you an example: part of the hieradata is coming from
> one group (site-wide config) and part from another group
> (cluster-specific config), with different permissions, release schedules
> and all.

As demonstrated earlier with hiera, it's possible to compose a single environment from multiple repositories.

> > In addition, the example configuration file is relying on undefined
> > behavior; if multiple environments with the same name are deployed
> > into a single directory there are no guarantees about what will
> > happen. In the future if environments will collide like this, r10k
> > will treat that as a hard error and will not update any environments.
> > I used a similar example like the one you have in the README.md (your
> > last one, called Multiple Environment Repositories Example), with the
> > exception that I can have multiple Puppetfiles per remote and you can
> > only have one, so you need to use two different remotes... but in both
> > we have the same basedir. Why do you say this will break in the future?
> > It works now...

The example in with multiple repositories used prefixing so that each environment from each source was prefixed the source name, so you would have `moria_modules_production` for the `production` branch in the `moria_modules` source. In your example, you have three possible production environments - one coming from `moria_modules`, one coming from `moria_roles`, and one coming from `moria_hieradata`. They are different environments but happen to have the same path, so they will try to manage the same repository. This happens to work because they have the same remote, but this is due to fortunate behavior on top of a bug. If two environments manage the same location but have different remotes you'll see thrashing with the environment version. It's nonsensical to allow environments to overlap; if you want to compose environments out of multiple sources you can still that but with differently named directories.

> > With respect to the subdir directive, how does that differ from the
> > 'moduledir' Puppetfile directive?
> > Unfortunately, the moduledir feature is slightly broken: you can only
> > have a single moduledir within the Puppetfile. I tried having multiple
> > ones (trying to divide the modules from the hieradata, for example) and
> > the purge wipes all moduledirs but the last one. And moduledir is
> > actually NOT part of the original Puppetfile format... it is an ugly
> > hack, if you allow the expression.

If this is a limitation of the `moduledir` setting that should be corrected. See below for my discussion on that. However I don't want to entangle the Puppetfile configuration with environment configuration any more than necessary.

> The subdir feature I propose, the "alternative directory", does exactly
> that, but the right way: in the r10k.yaml file you should specify where
> is the source of the Puppetfile (remote:) and where to deploy its
> contents (basedir: and subdir:). And Puppetfile is just a simple list of
> git repos, like it always was.
> 
> > Finally, this adds new features to r10k that adds flexibility, but
> > that comes with a cost. New features require testing, validation, and
> > documentation, and the more configurable a tool becomes the harder it
> > becomes to understand. Can the value of these new features justify the
> > cost of adding and maintaining them?
> > I understand. I can offer to write the documentation, but unfortunately
> > I don't have the knowledge to write tests. I am good at reading code (in
> > any language) and smart enough to copy-paste and/or modify those things
> > that are logical to me. Would you mind telling me how to run the tests
> > myself? Then I could see what breaks, and fix the tests accordingly by a
> > try-and-error approach. Then, testing the new features should not be
> > that difficult, I hope.
> 
> For me, these two features mean a great deal... I am one of those that
> think that either I do it right or I don't do it at all. I think I got
> far enough to have two very nice features that would make my life, and
> hopefully others, a better one... there must not be so much more until
> it's documented and tested.

If all you want is the ability to have multiple module directories, this was partially implemented by GH-44. In the long run I plan on implementing this sort of DSL for Puppetfiles:

``` ruby
moduledir 'default'

# mod declarations here will use the default moduledir

moduledir 'roles' do
  # mod declarations here will use the roles moduledir
end

moduledir 'profiles' do
  # mod declarations here will use the profiles moduledir
end
```

This change requires replumbing some of the internals of Puppetfiles so it has to wait for r10k 2.0. That being said it's possible and I plan on adding it. However, this is still only a logical separation - at the end of the day the Puppet master doesn't really care about the division of modules like this.

If you want to split up Puppetfiles you can already do this in ruby:

``` ruby
# main Puppetfile

load File.expand_path('Puppetfile.roles', File.dirname(__FILE__))
load File.expand_path('Puppetfile.profiles', File.dirname(__FILE__))
```

R10k 1.3 will add the ability to implement custom sources and environments, so if you have ideas about how r10k should deploy environments you'll be able to add those as extensions following the 1.3 release. In later versions of r10k Puppetfiles and modules will also be pluggable so all of the actual implementations of how bits on disk will be able to behave however you want. Unfortunately this pull request does not go in a direction I see r10k going. Since I don't intend on moving forward with this pull request, I'm unfortunately going to have to close it.

I've chosen to keep the r10k feature set less complex and well focused based on the reasoning laid out in the blog post at http://brianegranger.com/?p=249 ; that should hopefully clarify my stance on feature additions for r10k.

@acidprime I know that you've handling how r10k will run in large scale environments; I would appreciate an eye on this pull request in case this solves problems that I'm not understanding.

PR #141, opened 2014-05-27
Support moduledir with relative path

cbliard:
When using the moduledir directive in a Puppetfile, the path was assumed to
be an absolute path, even if the path did not start with a slash. In that case, the
directory was relative to the directory from where the user ran the r10k command.

Now when the moduledir path is relative, the path is joined with the Puppetfile basedir.
When the moduledir path is absolute, the behavior is the same as before.

I saw there were other pull requests : https://github.com/adrienthebo/r10k/pull/44 and https://github.com/adrienthebo/r10k/pull/130 . At least this one has some tests.

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -1,3 +1,4 @@
+require 'pathname'
 require 'r10k/module'
 require 'r10k/util/purgeable'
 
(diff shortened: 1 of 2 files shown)


adrienthebo:
Thanks for the contribution!

Issue #103, opened 2014-03-24
Documentation: Show usernames


rnelson0:
Please show the username of who should run various commands, or at least a $ vs # prompt. This is probably more important as git hooks and such are documented (Issue #102). I think all the documentation as of right now is run as root on the puppet master but I'm not 100% sure, thus the request.

adrienthebo:
@rnelson0 does https://github.com/adrienthebo/r10k/commit/3441eb9049a0ebada6a68a698cf4eabddbcd63c3 look like it meets your needs?

rnelson0:
I think that covers it on the deep documentation blurb. I do think the main README needs the same thing and, now that I'm more familiar with r10k I will try and find some time to generate a PR for that.

Aside: https://github.com/adrienthebo/r10k/blob/master/doc/dynamic-environments.mkd links to https://github.com/adrienthebo/r10k/blob/master/doc/dynamic-environments/usage, but that is a 404, like all the other links there.

adrienthebo:
Merged in e1b5afd, this should be released in 1.4.0. Although since it's documentation the release version isn't super critical...  ¯_(シ)_/¯

PR #151, opened 2014-06-10
Update README.mkd

rnelson0:
Describe what user must run r10k (GH-103) and link to usage document describing how to enable non-root users.


adrienthebo:
Thanks for the contribution!

PR #156, opened 2014-06-20
move configuration information higher in the readme

rmnwolf:
I think placing the configuration information higher in the readme makes a little more sense. What do you think?

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -49,6 +49,53 @@ Puppet Enterprise uses its own Ruby, so you need to use the correct version of g
     /opt/puppet/bin/gem install r10k
     r10k help
 
+Configuration
+---------------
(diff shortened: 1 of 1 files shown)


adrienthebo:
I wound up refactoring the README to contain requirements, installation instructions and links for usage to keep things more concise. Because of that this pull request no longer merges so I'm going to have to close it. 

Links for dynamic environments including configuration are now kept in https://github.com/adrienthebo/r10k/blob/master/doc/dynamic-environments.mkd - does this order look good?

adrienthebo:
Since this merge cannot merge cleanly I'm going to close it; if you would like to rework this pull request let me know and I can reopen it. Thanks!

PR #160, opened 2014-07-01
Quick Start Guide

ghost:
This change adds a new 'Quick Start' guide. The goal of this guide is to provide a succinct (but limited) walk-through on getting a 'full stack' of Puppet, Hiera, and r10k up and running in about 30 minutes.


adrienthebo:
Thanks!

PR #161, opened 2014-07-02
Deployments fail where a branch has \W in the git branch name

fiddyspence:
Where an r10k git repo uses branches where the name is a version, e.g. 1.2.3 a deployment will fail in `task/deployment.rb` because the `env.dirname` has been sanitised but the branchname which is later used to match is not.  This patch is the simple case of following the gsub performed elsewhere but in a private method.

### DIFF
--- a/lib/r10k/task/deployment.rb
+++ b/lib/r10k/task/deployment.rb
@@ -23,8 +23,14 @@ module Deployment
         # while this should be optimized that optimization can wait a while.
         names.each do |env_name|
 
+          # Elsewhere, we sanitise the env.dirname to sub any \W to _
+          # For this to match here (a single environment deploy where the 
+          # branch name has \W we need to sanitise the branch name to match
+
+          safe_branch_name = env_name.gsub(/\W/,'_')
+
           matching = all_environments.select do |env|
-            env.dirname == env_name
+            env.dirname == safe_branch_name
           end
 
           if matching.empty?


adrienthebo:
Compatibility is hard. :( My motivation for changing this was that the branch name normalization was not guaranteed, but people could be expected to know the branch that they created. Is this just for compatibility reasons or is it easier to use the normalized branch name? If a single environment is updated, should we try to update both the normalized and non-normalized name?

fiddyspence:
It's because it doesn't work if the branch name is 1.5.6 though - one part of the thing is normalized - the dirname => 1_5_6 but the branch name is still 1.5.6.

1.5.6 != 1_5_6 so no deploy:

```
pcpuppetjenkins:/build # /usr/bin/r10k deploy environment 1.5.6 -c /var/lib/jenkins/jobs/validate_specific_component/workspace/foocomponent/r10k.yaml
Faraday: you may want to install system_timer for reliable timeouts
[r10k - WARN] Calling `r10k --config <action>` as a global option is deprecated; use r10k <action> --config
[R10K::Source::Git - WARN] Environment "1.5.6" contained non-word characters, correcting name to 1_5_6
[R10K::Task::Deployment::DeployEnvironments - WARN] Environment 1.5.6 not found in any source
```

To be fair, it's a bit cheeky asking it to deploy an environment with an illegal name in the first place (but I'm not really deploying the thing to an environment right now).  Maybe I need to librarian instead.

adrienthebo:
My concern is that the old approach to deploying a specific environment meant that users needed to know exactly how r10k normalizes branch names. Is this a reasonable thing to assume?

fiddyspence:
The thing is whether or not a single environment deploy works, a user still needs to understand that when deploying _all_ environments where the branch name contains \W there's a normalisation happening.  The warning output I personally think is sufficient to do this:

```
spenceware:r10k fids$ git branch
 1.5.6
  my/shiny/new/feature
  my_new_env
  production`

[R10K::Source::Git - WARN] Environment "1.5.6" contained non-word characters, correcting name to 1_5_6
[R10K::Source::Git - WARN] Environment "my/shiny/new/feature" contained non-word characters, correcting name to my_shiny_new_feature

pcpuppetjenkins:/var/lib/jenkins/jobs/validate_specific_component/workspace/foocomponent/deploy # ls -l
total 16
drwxr-xr-x 5 root root 4096 Jul  3 09:06 1_5_6
drwxr-xr-x 5 root root 4096 Jul  3 09:07 my_new_env
drwxr-xr-x 5 root root 4096 Jul  3 09:07 my_shiny_new_feature
drwxr-xr-x 5 root root 4096 Jul  3 09:08 production
```

This patch just makes it possible to deploy a single environment with \W in the branch name - it doesn't change the need for the user to understand any of the use cases of the tool anywhere else.

adrienthebo:
Alright, sounds good. Could you rebase this into a single commit and reference the PR number in the commit? It'll make it easier when I write up release notes.

adrienthebo:
Merged into master in b4ef4f6, this should be released in 1.3.1. Thanks for the contribution!

PR #174, opened 2014-07-28
Use non-blocking read on subprocess stderr

### LINKED ISSUE #169: read on stderr hanging for git ssh operations with ControlPersist
When using ssh with ControlMaster/ControlPersist, while the ssh client double forks to open the master process, the new child retains the old stderr fd.  This causes r10k to hang on git commands that use ssh here:
https://github.com/adrienthebo/r10k/blob/master/lib/r10k/util/subprocess/posix/runner.rb#L42

The stderr read hangs until the ssh master session is killed or the timeout is reached if ControlPersist is set to a number.  This obviously defeats any benefit to using this optimization.  I have re-opened a bug with openssh about this behavior:
https://bugzilla.mindrot.org/show_bug.cgi?id=1988

For the time being, I have worked around the issue by using the GIT_SSH environment variable pointing to a ssh wrapper script that does a 2>/dev/null.  This is extremely clunky, but bypasses the issue.

I'm not sure if there is anything that can be done on the r10k subprocess management side that isn't a huge hack, but I thought it was worth logging this issue anyway.

jeremy-asher:
Here's a proposed fix using non-blocking reads to resolve #169.  I'm not sure if this is the cleanest solution as it does add some complexity to the normal subprocess code.  Let me know what you think.

### DIFF
--- a/lib/r10k/util/subprocess/posix/runner.rb
+++ b/lib/r10k/util/subprocess/posix/runner.rb
@@ -39,7 +39,11 @@ class R10K::Util::Subprocess::POSIX::Runner < R10K::Util::Subprocess::Runner
     end
 
     stdout = @stdout_r.read
-    stderr = @stderr_r.read
+    # Use non-blocking read for stderr_r to work around an issue with OpenSSH
+    # ControlPersist: https://bugzilla.mindrot.org/show_bug.cgi?id=1988
+    # Blocking should not occur in any other case since the process that was
+    # attached to the pipe has already terminated.
+    stderr = read_nonblock(@stderr_r)
 
     @result = R10K::Util::Subprocess::Result.new(@argv, stdout, stderr, @status.exitstatus)
   end
@@ -109,5 +113,20 @@ class R10K::Util::Subprocess::POSIX::Runner < R10K::Util::Subprocess::Runner
 
     @io.stdout = @stdout_w
     @io.stderr = @stderr_w
+  end
+
+  # Perform non-blocking reads on a pipe that could still be open
+  # Give up on reaching EOF or blocking and return what was read
+  def read_nonblock(rd_io)
+    data = ''
+    begin
+      # Loop until EOF or blocking
+      loop do
+          # do an 8k non-blocking read and append the result
+          data << rd_io.read_nonblock(8192)
+      end
+    rescue EOFError, Errno::EAGAIN, Errno::EWOULDBLOCK
+    end
+    data
   end
 end


lib/r10k/util/subprocess/posix/runner.rb
```diff
@@ -110,4 +114,19 @@ def attach_pipes
...
+  # Give up if blocking would occur and return an empty string
+  def read_nonblock(rd_io)
+    data = ''
+    begin
+      # Try an 8k non-blocking read
+      data << rd_io.read_nonblock(8192)
+      # Read the rest
+      data << rd_io.read
+    rescue EOFError
+    rescue Errno::EAGAIN, Errno::EWOULDBLOCK
```
jeremy-asher: This should probably be IO::WaitReadable

jeremy-asher:
I dug around for a while, but this seems to be the only reasonable solution with r10k's subprocess framework.  I'm not happy with needing the exception handling, but this style may be required in Ruby.  I'll do a little more cleaning up first if you think this makes sense.

lib/r10k/util/subprocess/posix/runner.rb
```diff
@@ -110,4 +114,19 @@ def attach_pipes
...
     @io.stderr = @stderr_w
   end
+
+  # Perform a non-blocking read on a pipe that could still be open
+  # Give up if blocking would occur and return an empty string
+  def read_nonblock(rd_io)
+    data = ''
+    begin
+      # Try an 8k non-blocking read
+      data << rd_io.read_nonblock(8192)
```
adrienthebo: If I'm reading this correctly, this assumes that if we can read the first 8k of the stream, then we can read the rest of the stream without blocking. Is this a safe assumption? If the pipe has 9k of data and we try this, will we still hang?

adrienthebo:
Is this issue triggered the first time we launch a process that uses SSH ControlPersist, or does this happen for every invocation of Git/SSH that we run?

lib/r10k/util/subprocess/posix/runner.rb
jeremy-asher: Good point.  This wouldn't necessarily work in all cases.  In my testing there was never anything in stderr when the ssh master was running during a git fetch, but that wouldn't always be the case.  I was tempted to make this read much bigger, but I couldn't find a good safe constant to set it to.  I settled on 8k since that was one of the default read sizes in the Ruby io code, but it was somewhat arbitrary.

jeremy-asher:
This happens for every git fetch using ssh during the invocation of r10k that started the master session, so if the ControlPersist is set to 10 seconds, there will be a 10 second wait after each fetch.  If a previously started ssh master session is running (possibly due to a long timeout), the next invocation of r10k will use it without blocking at all since the child processes will not be sharing stderr with the previously started master session.  This makes the problem a little harder to spot in some cases.

lib/r10k/util/subprocess/posix/runner.rb
jeremy-asher: The reason I used the read at all was because the read function in Ruby handles reading until EOF and a few other things automatically.  I suppose the best solution is to loop, doing read_nonblock with some size until EOF is reached.

adrienthebo: I'm nervous about only doing a nonblocking read for the first 8 kilobytes; if you can rework this to do nonblocking reads for the entire contents of the pipe I'll accept it.

jeremy-asher: I'll definitely fix that.  Do you think 8k reads make sense?

adrienthebo: We really don't have any guarantees about the contents of the pipe, so as long as we can read in a nonblocking manner till EOF 8k is probably as good as any other value We can always optimize later if need be.

jeremy-asher: Okay.  I'll work on that update when I get some time.  I will also look into creating some new tests for stderr output.

jeremy-asher:
I've updated this PR with a non-blocking read loop.  Let me know what you think.

jeremy-asher:
Actually, looks like the exceptions I changed aren't available in Ruby 1.8.7.  I'll revert that change...

jeremy-asher:
Okay, fixed the exceptions for 1.8.7.

jeremy-asher:
Made a little test program for the hell of it:

``` c
#include <stdio.h>
#include <unistd.h>

int main(int argc, const char* argv[]) {
    // 11 char test string
    char numstr[] = "0123456789-";
    int i, j;
    // print 106 lines of 78 chars (> 8192) to stderr
    for (i=0; i<116; i++) {
        for (j=0; j<7; j++) {
            fprintf(stderr, numstr);
        }
        fprintf(stderr, "\n");
    }
    // print test string in the second block of 8192
    fprintf(stderr, "test\n");
    // fork and daemonize, but leave stderr open
    int pid = fork();
    if (pid == 0) {
        daemon(0, 1);
        close(0);
        close(1);
        sleep(10);
    }
    return 0;
}
```

``` ruby
#!/usr/bin/env ruby
require 'r10k'
require 'r10k/util/subprocess'
proc = R10K::Util::Subprocess.new(['./daemon-test'])
result = proc.execute
exit(result.stderr.end_with? "test\n")
```

This will test to make sure stderr is being read.  It will hang for the 10 second sleep without this patch.

jeremy-asher:
Looks like there's an FD leak in Ruby 1.8.7 exec calls that's affecting r10k and this issue.  I'll open another PR with the fix for that.  See here: https://bugs.ruby-lang.org/issues/5041.

lib/r10k/util/subprocess/posix/runner.rb
```diff
@@ -110,4 +114,20 @@ def attach_pipes
...
     @io.stderr = @stderr_w
   end
+
+  # Perform non-blocking reads on a pipe that could still be open
+  # Give up on reaching EOF or blocking and return what was read
+  def read_nonblock(rd_io)
+    data = ''
+    begin
+      # Loop until EOF or blocking
+      while
```
adrienthebo: Ruby has the `loop` method that we can use to do something until specifically breaking, I would prefer to use that rather than a condition-less while.

lib/r10k/util/subprocess/posix/runner.rb
```diff
@@ -110,4 +114,20 @@ def attach_pipes
...
+          # do an 8k non-blocking read and append the result
+          data << rd_io.read_nonblock(8192)
+      end
+    rescue EOFError
```
adrienthebo: All of these errors can be on a single line.

jeremy-asher:
Okay, threw in those changes in.

adrienthebo:
Thanks for taking the time to track down and solve this!

PR #176, opened 2014-07-29
[README] Fixed an error in command description

cfoehrdes:
You described the "purge" command but listed "check"


adrienthebo:
Thanks!

PR #179, opened 2014-08-10
[SPIKE] Speedup entire environment deployment

s0enke:
This changes introduces a new option `puppetfile_if_changed` for the
`deploy environment` task.

It will check if the environment changed since last time and
automatically update modules if there is a change.

The goal is to have a fast deployment without clever post-receive hook
hacks. This option allows to run r10k e. g. as a cronjob each minute and
will always ensure up-to-date env without any manual intervention.

Thus we get a resilient and self-healing deployment for puppet
masters. The env will auto-heal even after network or machine failures.
Developers can expect to always have the latest env deployed without
manual intervention. All advantages of pull-style deploy apply.

This has only be implemented for git as a spike. Will add tests and doc
if the general idea looks good.

### DIFF
--- a/lib/r10k/cli/deploy.rb
+++ b/lib/r10k/cli/deploy.rb
@@ -51,12 +51,14 @@ scheduled. On subsequent deployments, Puppetfile deployment will default to off.
           DESCRIPTION
 
           flag :p, :puppetfile, 'Deploy modules from a puppetfile'
+          flag :i, :puppetfile_if_changed, 'Deploy modules from a puppetfile if the environment changed since last deploy'
 
           run do |opts, args, cmd|
             deploy = R10K::Deployment.load_config(opts[:config])
 
             task = R10K::Task::Deployment::DeployEnvironments.new(deploy)
             task.update_puppetfile = opts[:puppetfile]
+            task.update_puppetfile_if_changed = opts[:puppetfile_if_changed]
             task.environment_names = args
 
             purge = R10K::Task::Deployment::PurgeEnvironments.new(deploy)
--- a/lib/r10k/environment/git.rb
+++ b/lib/r10k/environment/git.rb
@@ -45,6 +45,10 @@ class R10K::Environment::Git < R10K::Environment::Base
     @puppetfile  = R10K::Puppetfile.new(@full_path)
   end
 
+  def outdated?
+    @working_dir.outdated?
+  end
+
   # Clone or update the given Git environment.
   #
   # If the environment is being created for the first time, it will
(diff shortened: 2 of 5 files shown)


s0enke:
@adrienthebo any thoughts on this?

deviantony:
+1 great idea.

s0enke:
@adrienthebo any thoughts on this?

adrienthebo:
@s0enke I'm still not clear on how the `-i` option would differ from using `-p` selectively - if you only want to update modules at specific times, why not use `-p` more selectively?

Also, there's a very big problem in this - updates to modules may not stem from changes to the Puppetfile or Git repository. There are two big examples of this right now. First off, Forge modules can be installed with a version of `:latest`. If someone pushes a new version of a module and a user is tracking that module with `:latest`, then it should be installed right away regardless of changes or lack thereof to the Puppetfile or repository in general. The other case is very similar - if someone is tracking a Git module via the branch and a new commit is pushed to that branch it needs to be updated without having to change the  Puppetfile.

Given these problems, is there a way to use this approach and address the above issues?

s0enke:
@adrienthebo thx for your feedback :-)

> @s0enke I'm still not clear on how the -i option would differ from using -p selectively - if you only want to update modules at specific times, why not use -p more selectively?

Actually, what I was trying to achieve is a "update everything in a timely manner" behavior. This enables the following:
- I can update my puppet master every minute, e. g. with a cronjob (`-p` was too slow for that)
- I don't have to build clever fragile mechanisms, e. g. hooks like "if branch X got pushed, execute command Y on all puppet masters", because
- I want a self-healing, but up-to-date puppetmaster infrastructure: The entire system should be resilient against network failures, nodes being down etc (http://agiletesting.blogspot.de/2010/03/automated-deployment-systems-push-vs.html)

As I see your concerns, I will have to dig deeper into it or maybe choose a different strategy.

adrienthebo:
One of the issues that you raised is that you don't want want to build clever fragile mechanisms, but this change moves the cleverness and fragility into r10k. Because modules can be updated independently of environments, this is guaranteed to produce incorrect behavior. Every current module implementation allows for modules to be updated without changing the Puppetfile, which will make `-i` silently fail.

What might work much more cleanly is to build in a _decision making capability_ into r10k after an environment is deployed. I can foresee something like having r10k run a script after an environment is deployed that is passed the old revision and new revision. That script could either return information on stdout or indicate via return code what should happen next - return with no further changes, deploy modules, or perhaps fail. This would be a much more powerful and flexible feature, and it will allow users to tailor the behavior of r10k as they see fit, rather than build a single workflow pattern into the application itself.

adrienthebo:
Because this pull request hasn't seen any activity in a few months I'm closing it for now; if you would like to continue to work on it please comment on this pull request and I can reopen it.

PR #180, opened 2014-08-10
FD leak fixes

jeremy-asher:
Several FDs were not being closed, some due to an issue on Ruby 1.8.7 (https://bugs.ruby-lang.org/issues/5041).  This is visible with lsof while r10k is running for a reasonably long Puppetfile.

(relates to #174)

### DIFF
--- a/lib/r10k/util/subprocess/posix/runner.rb
+++ b/lib/r10k/util/subprocess/posix/runner.rb
@@ -107,6 +107,11 @@ class R10K::Util::Subprocess::POSIX::Runner < R10K::Util::Subprocess::Runner
     @stdout_r, @stdout_w = ::IO.pipe
     @stderr_r, @stderr_w = ::IO.pipe
 
+    @stdout_r.fcntl(Fcntl::F_SETFD, Fcntl::FD_CLOEXEC)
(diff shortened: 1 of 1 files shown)


adrienthebo:
Does this supersede #174? And out of curiosity how many modules are we talking about?

jeremy-asher:
I discovered this while working on #174, but it doesn't fix the issue with ControlPersist.

You can catch these leaks if you run an lsof while running a Puppetfile that takes more than a few seconds.  I had noticed the large number of pipes in the past when running over 100 modules, but I didn't think it was a major issue at the time.  I would like to start calling r10k directly in some longer lived code and the leaks would probably become significant in that case.

The Ruby 1.8.7 exec leak cause the test program I linked in #174 to inherit all r10k pipe handles and then trigger a block on reading from @stdout_r, but it doesn't seem to be causing any significant issues in the real world, with or without ControlPersist.  I'm still not sure how openssh manages to do it, but it seems to close all the inherited pipes for the Control Master process and just hangs on to stderr from the ssh client, side-stepping that issue.

jeremy-asher:
Actually, in normal use, git is launching ssh, so that's probably the reason the pipes don't make it all the way down.

adrienthebo:
Thanks for chasing this down and submitting the fix!

Issue #178, opened 2014-08-08
git repo fails to clone then r10k purges everything breaking the masters


kholloway:
Occasionally my git repo fails to clone for some reason, when that happens r10k then purges the environments directories and wipes out everything, our puppet masters then fail to properly work.

Seems like the destination directory should be left alone if r10k fails to run to completion for all repos BEFORE wiping out what was a working directory/setup.

Is there any current CLI option I'm missing to fix this?

Thanks!
### R10k commands tried:

`r10k deploy environment -pv`
`r10k deploy environment`
### R10k CLI output

```
[R10K::Source::Git - INFO] Determining current branches for "git@gitlab.name.changed:teamname/puppet3.git"
[R10K::TaskRunner - ERROR] Task #<R10K::Task::Deployment::DeployEnvironments:0x00000001dffdf8> failed while running: Couldn't update git cache for git@
gitlab.name.changed:teamname/puppet3.git: "fatal: Could not read from remote repository."
[R10K::Task::Deployment::PurgeEnvironments - INFO] Purging stale environments from /etc/puppet/environments
```
### /etc/r10k.yaml

```
# The location to use for storing cached Git repos
:cachedir: '/var/cache/r10k'

# A list of git repositories to create
:sources:
  # This will clone the git repository and instantiate an environment per
  # branch in /etc/puppet/environments
  :puppet3:
    remote: 'git@gitlab.name.changed:teamname/puppet3.git'
    basedir: '/etc/puppet/environments'
```
### Puppetfile from repo puppet3.git

```
forge 'forge.puppetlabs.com'

## Specify Forge modules here
mod 'puppetlabs/vcsrepo', '1.1.0'
mod 'puppetlabs/stdlib', '4.3.2'
mod 'puppetlabs/concat', '1.0.4'

# Specify any local Gitlab repos here
# Old Puppet repo modules - by default master will be checked out
mod "old_puppet_repo",
   :git => 'git@gitlab.name.changed:teamname/puppet.git'
```

adrienthebo:
Well, that's something I didn't expect to see. Could you run `r10k deploy environment -v debug2 --trace` and try to reproduce this with backtraces and the higher logging?

kholloway:
Below is the output from the debug, our gitlab server fails the connection for some reason followed by R10k purging everything. As soon as the gitlab server works fine again then everything comes back again as expected.

```
[...]

PR #182, opened 2014-08-14
Fix for when a repository fails to clone or pull and purge deletes everything

kholloway:
Fix for #178

### DIFF
--- a/lib/r10k/task_runner.rb
+++ b/lib/r10k/task_runner.rb
@@ -35,6 +35,8 @@ class TaskRunner
 
           @errors[current] = e
           @succeeded = false
+          # Clear all remaining tasks after a failure, keeps the purge from deleting valid content
+          @tasks.clear
         end
         @tasks.shift
       end


lib/r10k/task_runner.rb
```diff
@@ -35,6 +35,8 @@ def run
 
           @errors[current] = e
           @succeeded = false
+          # Clear all remaining tasks after a failure, keeps the purge from deleting valid content
+          @tasks.clear
```
adrienthebo, line 39: With this change, if a Puppet module fails to install then this will abort the entire run. We need a way to delineate between errors that we can proceed from (an environment couldn't be updated, a module failed to update) and errors that must abort the entire run.

adrienthebo:
If a git repository cannot be fetched then we should still be able to fetch the branches for that branches, I would like to investigate why that's happening and if we can fix that before taking the approach of immediately aborting the update. What do you think of trying to update the known environments when the git repo fails to update instead?

kholloway:
How about having task_runner check if it's doing a purge task of any kind and just skipping it if any previous error has occured instead of bailing out on the rest of the tasks..?

kholloway:
How about this? It skips all the purge tasks if we encounter any previous failure but doesn't abort the run itself or skip anything other than the purge tasks.

``` diff
--- /usr/lib64/ruby/gems/2.0.0/gems/r10k-1.3.2/lib/r10k/task_runner.rb  2014-08-14 15:05:01.030455823 -0500
+++ /tmp/task_runner.rb 2014-08-14 15:04:48.305459701 -0500
@@ -23,7 +23,11 @@
         current = @tasks.first
         current.task_runner = self
         begin
-          current.call
+          if current.class.name =~ /purge/i && !@succeeded
+            puts "We had a previous failure and now a purge request, skip it instead"
+          else
+            current.call
+          end
         rescue Interrupt => e
           logger.error "Aborted!"
           $stderr.puts e.backtrace.join("\n").red if @trace
```

adrienthebo:
That patch will be triggered if any sort of failure occurred, such as if a module was installed. It's a very imprecise solution and doesn't directly address why this failure is occurring in the first place.

adrienthebo:
The root cause for this issue was due to improperly using an instance variable instead of a lazy method (https://github.com/adrienthebo/r10k/commit/f8f8a8d118344ae29c01fe9050b61a80bb9cdb70) and has been fixed; I'm closing this issue in favor of the committed fix. Thanks for your work on this regardless!

PR #219, opened 2014-10-30
[puppetfile] Wrap SyntaxError and LoadError on load

adrienthebo:
SyntaxError and LoadError don't inherit from StandardError, so if
the process of loading a Puppetfile raises either one of those, it will
completely unwind the stack and crash the r10k process. Since this is a
special case where we're loading Ruby code that's not actual application
code we need to rescue and wrap these errors.


adrienthebo:
These tests are incorrectly passing; the changes made were ported from master which has rspec 3.1, so the block form of raise_error is being silently ignored.

PR #238, opened 2014-11-14
[action/deploy/environment] fix missing environment validation

lazyfrosch:
Broken by the rewrite to visitor in
  f894a9dfb2cbefc88b3d5827f4af29a059c8f301

Question is: should we do a test for that? :)

The current test only checks if the .validate call fails on something, not if it is called.


adrienthebo:
Thanks for the catch! It would be nice to test this but this would be better to test via an acceptance test or integration test; it would take an absurd amount of mocking to unit test to catch this regression. I'm going to merge this as is, feel free to submit a second pull request for tests if you're feeling ambitious.

PR #243, opened 2014-11-24
Use the SPDX format in the license

tampakrap:


adrienthebo:
What is SPDX?

tampakrap:
http://spdx.org/licenses/

adrienthebo:
Could you update the commit message to reference this link and indicate that this is being done for the sake of OpenSUSE packages?

tampakrap:
done

adrienthebo:
Thanks for the contribution! Do you need a 1.3.6 release so that you can ship this out on your end? And if so do you need any other changes made for packaging?

tampakrap:
The followings are still pending:
1) a change in gemspec git ls-files, to exclude a few dot files to make rpmlint happy. I'll submit a pull request for this shortly
2) cri ~> 2.6 (https://github.com/adrienthebo/r10k/pull/226)
3) multi_json ~> 1.10
4) faraday ~> 0.9

I haven't checked at all yet in the code if the updates in the deps are possible without any changes in r10k

PR #245, opened 2014-11-28
Issue/master/gh 95 puppetfile action

adrienthebo:
This replaces the last uses of the Task namespace and deprecates the remaining classes.


adrienthebo:
This needs to be validated; there looks like there's some bad behavior with the puppetfile check action.

PR #250, opened 2014-12-08
Update to faraday 0.9.0

tampakrap:
The user_agent attribute is dropped apparently, and no replacement was needed
since it seems to send the desired user-agent string


adrienthebo:
It looks like we weren't sending the specified user data anyways; that can be fixed later. Thanks for the fix!

PR #253, opened 2014-12-09
Allow module names to contain dashes

claytono:
It's not recommended that module names contain dashes, yet some of them
do.  This changes the module name parsing to allow alpha-numeric
characters plus underscores and dashes.

Additionally, this logs the name of the invalid entry when an invalid
module name is found.

### DIFF
--- a/lib/r10k/module/base.rb
+++ b/lib/r10k/module/base.rb
@@ -91,12 +91,12 @@ class R10K::Module::Base
   private
 
   def parse_title(title)
-    if (match = title.match(/\A(\w+)\Z/))
-      [nil, match[1]]
-    elsif (match = title.match(/\A(\w+)[-\/](\w+)\Z/))
+    if (match = title.match(/\A(\w+)[-\/]([[:alnum:]_-]+)\Z/))
       [match[1], match[2]]
+    elsif (match = title.match(/\A([[:alnum:]_-]+)\Z/))
+      [nil, match[1]]
     else
-      raise ArgumentError, "Module names must match either 'modulename' or 'owner/modulename'"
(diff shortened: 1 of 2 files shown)


claytono:
The test failure looks to be a transient error.  I don't see an option to retry the test.

lib/r10k/module/base.rb
```diff
@@ -91,12 +91,12 @@ def properties
   private
 
   def parse_title(title)
-    if (match = title.match(/\A(\w+)\Z/))
-      [nil, match[1]]
-    elsif (match = title.match(/\A(\w+)[-\/](\w+)\Z/))
+    if (match = title.match(/\A(\w+)[-\/]([[:alnum:]_-]+)\Z/))
```
adrienthebo, line 94: I would rather use `\w` as it includes `[:alnum:]` and `_`.

spec/unit/module/base_spec.rb
```diff
@@ -24,7 +24,7 @@
     it "raises an error when the title is not correctly formatted" do
```
adrienthebo, line 24: This should include a test for hyphenated module names.

adrienthebo:
The Forge module provider does not allow for module names to have hyphens (https://github.com/adrienthebo/r10k/blob/master/lib/r10k/module/forge.rb#L170) but I couldn't find any hyphenated modules on the Forge. Does this sound alright to you?

adrienthebo:
There are still outstanding concerns on this pull request, do you think you'll be able to get to this in the next couple of weeks?

claytono:
I'm not convinced it's possible to have robust handling of owner-modulename type formats plus module names that contains hyphens.  Along with that, and the fact that future parser doesn't allow these, I'm closing this PR and I opened a new one that solely contains the enhanced error reporting when bad names are found.

PR #255, opened 2014-12-11
Add 'strip_branchname_prefix' option for git

jonnangle:
Gitolite supports [personal branches](http://gitolite.com/gitolite/user.html#pers), which are a convenient way to allow a large group of developers push access to any number of branches within their own dedicated namespace. Personal branches require a fixed prefix on the branch name (e.g. "personal/". It works really nicely as a way to deliver dynamic Puppet environments, but it does add some noise to the naming - given enough development streams, you wind up with lots and lots of Puppet environments called `personal_<something>`

This commit provides a 'strip_branchname_prefix' option, to allow you to strip a fixed string from the start of a branch name before r10k goes and does its stuff.

(I'm not sure how widely used personal branches are for Puppet, and therefore how many other r10k users might want something like this - but we needed this to migrate to r10k from our custom post-hook.)

### DIFF
--- a/doc/dynamic-environments/git-environments.mkd
+++ b/doc/dynamic-environments/git-environments.mkd
@@ -63,3 +63,25 @@ Valid values:
   * 'correct': Non-word characters will silently be replaced with underscores.
   * 'error': Branches with non-word characters will be ignored and an error will
     be emitted.
+
+### strip_branchname_prefix:
+
+This setting tells r10k to strip a fixed string from the start of a Git branch
+name before using it. If the branch name does not start with the specified
+prefix, nothing happens and processing continues as normal. Please note that
+the string gets stripped before any of the other settings that might influence
+the name of the Puppet environment (e.g. 'invalid_branches' or 'prefix') take
+effect.
+
+For example, a Git branch called 'personal/username/branchname' will cause r10k
+to create a Puppet environment named 'personal_username_branchname' by default
+(assuming that 'invalid_branches' is set to 'correct' or 'correct_and_warn',
+and 'prefix' is not enabled). However, if you were to set:
+
+    strip_branchname_prefix: personal/
+
+on the source in r10k.yaml, this prefix would be stripped and you would end up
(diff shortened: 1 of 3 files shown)


adrienthebo:
I see the need for functionality like this, but I'm not convinced that this is the best approach. This deals with behavior that's very specific to Gitolite which means that the majority of people won't be able to make use of it, and it's solving a single case where the Git branch name doesn't match the intended environment name. Is there a way to provide a more general way to convert a Git branch name to a Puppet environment name?

jonnangle:
I agree :) I was originally going to make it more general-purpose, but I convinced myself not to. Partly because I wasn't sure whether there was much of a demand for mapping Git branch names to Puppet environment names outside of this one specific requirement, and partly because I thought it might overlap with existing functionality such as `invalid_branches` in confusing ways. Though I'm not sure that what I've ended up with is any less overlappy and confusing, to be honest.

Do you think it would be better to have something like a regular expression that you could use to match against the Git branch name, and pull capture group matches out to use in the creation of a Puppet environment name? Something like:

```
branch_mapper: /^personal/(\w+)/(\w+)$/ "$1_$2"
```

adrienthebo:
Truth be told, I don't fully know how people need to do branch naming, but I do know that it's being done. My concern is that there are tons of special cases that could be supported so I'm averse to implementing each special case, but barring custom plugins that implementing this remapping I haven't hit on a good general mapping solution.

So since this pull request is implementing a new feature, could you review the section in the contributing document (https://github.com/puppetlabs/r10k/blob/master/CONTRIBUTING.mkd#feature-requests) and review the questions regarding feature requests?

adrienthebo:
This pull request hasn't seen an update in about a year, and would need to be rebased in order to get merged. Moreover this is implementing code for an extremely specific use case, and I'm reluctant to take on the maintenance costs for features that won't be widely used. It's possible to plug in additional environment implementations so this could be implemented as a plugin; if you would like to continue work on this as a plugin I'm happy to talk about that in depth. But since this pull request hasn't seen activity, I'm closing it out.

PR #256, opened 2014-12-11
Parametrize Puppetfile load methods

raphink:
This allows to load more Puppetfiles at will from within a Puppetfile, e.g.:

``` ruby
dsl.load(File.dirname(__FILE__)+'/Puppetfile.common')
```

which can be very useful if you have several sites that share a common set of modules.

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -40,19 +40,19 @@ class Puppetfile
     @forge   = 'forge.puppetlabs.com'
   end
 
-  def load
-    if File.readable? @puppetfile_path
-      self.load!
+  def load(path = @puppetfile_path)
+    if File.readable? path
+      self.load!(path)
     else
-      logger.debug "Puppetfile #{@puppetfile_path.inspect} missing or unreadable"
+      logger.debug "Puppetfile #{path.inspect} missing or unreadable"
     end
   end
 
-  def load!
+  def load!(path = @puppetfile_path)
     dsl = R10K::Puppetfile::DSL.new(self)
-    dsl.instance_eval(puppetfile_contents, @puppetfile_path)
+    dsl.instance_eval(puppetfile_contents(path), path)
   rescue SyntaxError, LoadError => e
-    raise R10K::Error.wrap(e, "Failed to evaluate #{@puppetfile_path}")
+    raise R10K::Error.wrap(e, "Failed to evaluate #{path}")
   end
 
   # @param [String] forge
@@ -98,8 +98,8 @@ class Puppetfile
 
   private
 
-  def puppetfile_contents
-    File.read(@puppetfile_path)
+  def puppetfile_contents(path = @puppetfile_path)
+    File.read(path)
   end
 
   class DSL


adrienthebo:
Why not use Kernel#require or Kernel#load?

raphink:
Because I won't get the `forge` and `mod` methods available if I do that.

adrienthebo:
Egh, you're right, to make this possible you need a an instance_eval.

raphink:
Yes, which is why I want to use `R10K::Puppetfile::DSL#load`, since it already does this and manages errors cleanly.

adrienthebo:
In practice this will behave like you have one Puppetfile that has nested Puppetfile instances but the implementation doesn't consider this. I would rather we use a separate method like 'puppetfile xyz' instead of load, which would spawn a new puppetfile instance and delegate loading to the nested instance, which would effectively be the Composite pattern. Thoughts?

raphink:
A new 'puppetfile' method would do just as well, and will certainly feel cleaner.

raphink:
Now to tell the truth, what we've been doing is a little more complex than this.

We have 4 different sites, using modules with the same pinning, but a different set. So, we have a `Puppetfile.common` file which lists all the modules with the pinning, and then we include it from the Puppetfile and filter out the modules we don't need for a specific site, like this:

``` ruby
class R10K::Puppetfile::DSL
  def mod(name, args=[])
    m = name[/[^\/]+\/(.*)/, 1]
    @librarian.add_module(name, args) if @modules.include? m
  end 
end

require 'pathname'
dir = Pathname.new(__FILE__).parent
@modules = YAML.load_file(File.join(dir, 'modules.yml'))

common = File.join(dir, 'Puppetfile.common')
begin
  dsl.instance_eval File.read(common)
rescue => e
  raise R10K::Error.wrap(e, "Failed to evaluate #{common}")
end
```

where `modules.yml` contains a list of modules we want to use on that particular site.

I'd love for a clean way to do this without monkey patching `R10K::Puppetfile::DSL`. Any thoughts on how this could come with a `R10K::Puppetfile::DSL::puppetfile` method?

adrienthebo:
Huh. That's an interesting use case. Why are you doing filtering in this manner where you have a separate list instead of doing that in the Puppetfile?

raphink:
I can't override the `mod` method from within the Puppetfile without reloading everything, so it ends up being easier to monkey patch and then load a second file.

adrienthebo:
Well rather, why does modules.yml need to exist? Why do you have this large list of modules that you filter down?

raphink:
The various sites we manage each have a different set of modules, but we want to keep the pinning identical, so we don't have to manage different versions of modules on different sites.

It seemed easier to have a `Puppetfile.common` file that had all the pinning (and sources, since some modules are from the forge, others from github) for all the modules we use on all sites, synchronize this file on each site, and then filter out which modules are actually required on any given site, rather than duplicate information on each site.

adrienthebo:
Got it. So here is an option - define a method called pin, which checks the `@modules` list, and if it should be present then you call mod, otherwise you ignore it. So you can pin Amy number of modules but only instantiate modules you want,  no monkey patching required.

raphink:
That's an option indeed. Much more simple for sure ;-). /cc @mcanevet

PR #266, opened 2014-12-17
Clean up Subprocess Runner API, remove code duplication from Runner::POSIX

adrienthebo:
This is based on GH-264.


adrienthebo:
This is on the PL fork because of the repo move, reopening under my fork.

Issue #265, opened 2014-12-17
Unix commands with > 64KiB of output may deadlock


adrienthebo:
Linux > 2.6.11 has a maximum pipe buffer size of 64KiB, and when the buffer is filled further writes will block until data is read from the pipe. This means that r10k may launch processes that will block on writing to stdout, and r10k will block on waiting for the process to return. In order to make this work the stdout/stderr pipes need to be continually drained to prevent this deadlock.

adrienthebo:
See https://tickets.puppetlabs.com/browse/RK-11

adrienthebo:
Merged in 2308016, this should be released in 1.5.0. Further updates on this issue should be made at https://tickets.puppetlabs.com/browse/RK-11.

To avoid duplicating discussion between this thread and the JIRA issue I'm locking this issue.

PR #267, opened 2014-12-17
(GH-265) Drain POSIX command runner IO pipes

adrienthebo:
This is based on GH-276


lib/r10k/util/subprocess/runner/pump.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  private
+
+  def pump
+    backoff = @delay
+    while @run
+      begin
+        @string << @io.read_nonblock(4096)
+        backoff /= 2 if backoff > @delay
```
adrienthebo, line 43: This should have an upper bound on the backoff time so this doesn't start sleeping for extremely long times.

lib/r10k/util/subprocess/runner/posix.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  def mkpipes
+    @stdout_r, @stdout_w = pipe
+    @stderr_r, @stderr_w = pipe
+  end
+
+  # Perform non-blocking reads on a pipe that could still be open
+  # Give up on reaching EOF or blocking and return what was read
+  def read_nonblock(rd_io)
```
elyscape, line 99: This doesn't seem to be used anywhere since the refactor into Pump.

adrienthebo, line 99: Good catch!

adrienthebo:
This pull request was based off of a branch that was on my repository before I moved it and was then off the puppetlabs repo; I'm reopening this against my own fork.

Issue #269, opened 2014-12-23
Checking out an unresolvable git ref outputs a horrid error message


adrienthebo:
Given a scenario were r10k is trying to deploy a Git module from an empty repository with no branches, it will provide a pretty dismal error message:

```
[R10K::Action::Puppetfile::Install - ERROR] Could not locate hash at /home/adrien/development/r10k/testing/puppetfile/empty-repository/modules/foo
```

The _hash_ part of the error message is misleading as it's not a hash that can't be located, it's a Git reference that can't be resolved.

r10k should indicate what actually failed in this case and provide a useful error message.

adrienthebo:
See https://tickets.puppetlabs.com/browse/RK-10

adrienthebo:
Merged in 0bc6a95; this will be released in 1.4.1.

PR #270, opened 2014-12-23
[git/errors] (GH-269) Improve error messaging on checkout

adrienthebo:
If a working directory is instructed to check out a Git reference that's
unresolvable it would previously spit out a pretty useless error
message; this commit corrects the emitted error message and adds
additional context for the error.

### DIFF
--- a/lib/r10k/git/errors.rb
+++ b/lib/r10k/git/errors.rb
@@ -10,31 +10,17 @@ module R10K
       attr_reader :ref
       attr_reader :git_dir
 
-      def initialize(*args)
+      def initialize(mesg, options = {})
         super
-
-        @hash    = @options[:ref]
+        @ref     = @options[:ref]
         @git_dir = @options[:git_dir]
       end
 
-      HASHLIKE = %r[[A-Fa-f0-9]]
-
-      # Print a friendly error message if an object hash is given as the message
       def message
-        if @mesg
-          msg = @mesg
-        else
-          msg = "Could not locate hash"
-
-          if @hash
-            msg << " '#{@hash}'"
-          end
-        end
-
+        msg = super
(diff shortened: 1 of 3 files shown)


lib/r10k/git/errors.rb
```diff
@@ -10,24 +10,24 @@ class UnresolvableRefError < GitError
       attr_reader :ref
       attr_reader :git_dir
 
-      def initialize(*args)
+      def initialize(mesg, options = {})
+        if mesg.is_a? Hash
```
andersonmills: What's the reason for receiving options through both an options hash and `mesg` as options hash?

adrienthebo: It's a quick of having variable arity methods in ruby, where callers can provide a message, additional options, or both. If both are passed then it's easy, but if just an options hash is passed and not a message, then all arguments will be shifted down - mesg will be the hash, and options will be nil.

That being said, I think this is pretty messy after looking at it again and should be revised.

andersonmills: I will note that this type of variable-arity pandering (!) doesn't seem to exist in any of the other r10k errors.

adrienthebo: Yeah, a lot of it was due to me trying to allow people to omit an explicit error message with https://github.com/puppetlabs/r10k/pull/270/files#diff-6645d47c52f4a2cd3f67ed2795e18f8bR23 but it's caused more heartache than it's worth.

andersonmills:
Tested positive and negative case. Happy with it. :+1:

andersonmills:
Currently this is the JIRA ticket RK-10, but I wouldn't be surprised if that ticket ID changes.

PR #271, opened 2014-12-23
More verbose module name error messages

claytono:
When a module name doesn't match one of the acceptable regexes, then
print out the name of the module that was the problem.


adrienthebo:
Note - this will need to be backported onto 1.4.x.

adrienthebo:
Merged into 1.4.x in 42df611 and merged into master in 12f157a.

adrienthebo:
Thanks for the contribution!

PR #276, opened 2015-01-06
Maint/master/cleanup posix runner api

adrienthebo:


adrienthebo:
See https://github.com/puppetlabs/r10k/pull/266

spec/shared-examples/subprocess-runner.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect(result.stdout).to eq 'no-execute.sh'
+    end
+  end
+
+  describe "running '/bin/false'" do
+    subject { described_class.new(%w[/bin/false]) }
+
+    it "sets the exit code to 1" do
+      result = subject.run
+      expect(result.exit_code).to eq 1
```
andersonmills, line 41: Notably, /bin/false doesn't exist on my OS X system. (It's /usr/bin/false.) Do we care?

adrienthebo, line 41: I suspect that this will make the tests fail on your system, which will be a problem. I think that we could use use the PATH variable instead of a fully qualified binary.

andersonmills, line 41: Tests failing was how I found this, of course. :smile:

adrienthebo, line 41: Specs fixed to use unqualified command paths.

andersonmills:
:+1: I have tested the deploying of modules, git, and forge usages of Subprocess, and believe in it.

andersonmills:
I also believe in the tests.

PR #277, opened 2015-01-07
(RK-11) empty pipe buffers during command execution

adrienthebo:
This pull request is based on GH-276. 

See https://tickets.puppetlabs.com/browse/RK-10 and https://github.com/puppetlabs/r10k/issues/265 .


andersonmills:
:+1: I think this works and the tests test that it does.

PR #278, opened 2015-01-09
(RK-5) Add Basic User Scenario Integration Test

cowofevil:
Create an integration test for a basic r10k workflow. Segregate integration
tests from existing acceptance and spec tests. Created a folder layout for
integration tests. Created a pre-suite and library for future r10k integration
tests.

### DIFF
--- a/integration/tests/user_scenario/basic.rb
+++ b/integration/tests/user_scenario/basic.rb
@@ -14,8 +14,6 @@ site_pp_path = File.join(git_environments_path, 'manifests', 'site.pp')
 site_pp = create_site_pp(master_certname, '  include helloworld')
 
 #Verification
-test_folder = '/tmp/manifest'
-test_file = '/tmp/manifest/random_fifth'
 notify_message_regex = /I am in the production environment/
 
 #Teardown
--- a/integration/tests/user_scenario/basic.rb
+++ b/integration/tests/user_scenario/basic.rb
@@ -1,0 +1,50 @@
+require 'git_utils'
+require 'master_manipulator'
+test_name "RK-5 - C59118 - Basic User Scenario"
+
+#Init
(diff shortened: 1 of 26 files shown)


elyscape:
Why are the configurations suffixed with 64mda?

cowofevil:
@elyscape The "mda" suffix is short for "Master, Dashboard and Agent". The designation informs the tester that one of the hosts used for testing has all the above mentioned roles. We include the configs solely for local testing of integration tests and we actually run with a much larger matrix of platforms in our CI pipeline.

zreichert:
Is there a Jira ticket to track this work?   If so could you please incorporate the ticket number as documented here https://confluence.puppetlabs.com/display/QA/Guide+for+Git+in+QA

integration/tests/user_scenario/basic.rb
```diff
@@ -0,0 +1,50 @@
...
+step 'Push Changes'
+git_add_commit_push(master, 'production', 'Update site.pp and add module.', git_environments_path)
+
+#Tests
+step 'Deploy "production" Environment via r10k'
+on(master, 'r10k deploy environment -v')
+
+agents.each do |agent|
+  step "Run Puppet Agent"
+  on(agent, puppet('agent', '--test'), :acceptable_exit_codes => 2) do |result|
```
zreichert, line 44: You should expand your acceptable_exit_codes to include 0.

cowofevil, line 44: @zreichert The Puppet run should always produce a notify message which will always register as a change so in this case if the Puppet run didn't return "2" we know that a failure occurred. Adding "0" wouldn't necessarily hurt anything, but would technically make the validation looser.

cowofevil:
@zreichert I've updated the PR, can you review again?

PR #280, opened 2015-01-09
(RK-13) Use ChildProcess for executing commands on JRuby.

adrienthebo:
This is based on https://github.com/puppetlabs/r10k/pull/277 .


lib/r10k/util/subprocess/runner/childprocess.rb
```diff
@@ -0,0 +1,45 @@
...
+    process.io.stdout = stdout_w
+    process.io.stderr = stderr_w
+
+    begin
+      stdout_pump.start
+      stderr_pump.start
+      process.start
+      process.wait
+      stdout_pump.halt!
+      stderr_pump.halt!
```
joshcooper, line 34: I assume these methods close the underlying file descriptor? If so, you probably want put these in nested begin-ensure blocks to ensure we don't leak file descriptors.

adrienthebo, line 34: Good catch! I was relying on the garbage collector to close the file descriptors when the IO objects go out of scope but since that's proven to be problematic I'll throw in the explicit closes.

adrienthebo, line 34: Fixed in https://github.com/adrienthebo/r10k/commit/d816ea9cbf53806477cfc718f20ea1bd54960f07.

PR #285, opened 2015-01-16
[runner/posix] (RK-18) Read stdout/stderr to EOF

adrienthebo:
In order to capture all command output we need to read until an explicit EOF on the IO pipes.

This conflicts with GH-169.

### DIFF
--- a/lib/r10k/util/subprocess/runner/posix.rb
+++ b/lib/r10k/util/subprocess/runner/posix.rb
@@ -77,8 +77,8 @@ class R10K::Util::Subprocess::Runner::POSIX < R10K::Util::Subprocess::Runner
       _, @status = Process.waitpid2(pid)
     else
       _, @status = Process.waitpid2(pid)
-      @stdout_pump.halt!
-      @stderr_pump.halt!
+      @stdout_pump.wait
+      @stderr_pump.wait
       stdout = @stdout_pump.string
       stderr = @stderr_pump.string
     end


jeremy-asher:
Is the wait on the child process insufficient to ensure that all data is in the pipe?  Could the truncation referenced in [RK-18](https://tickets.puppetlabs.com/browse/RK-18) instead be caused when the process ends while the Pump thread was waiting on an IO.select?  In that case, wouldn't the halt! call trigger an immediate end of the loop before a final read from the pipe, even if there was data waiting?

I think you'd have to loop, watching the output of IO.select to see if there is more than 4k of data and then terminate when it finally returns nil.  So, theoretically, all output has been received when the wait returns and then select returns nil.

adrienthebo:
It might be possible to just throw another `#read_nonblock(4096)` after `@thread.join` is invoked to ensure that we read as much as possible off of the pipe, but it's totally possible for the thread to dump 8k of data into the pipe and immediately exit. In that case, we're still back in the position where we will still truncate output. The only case where we can guarantee that we'll read all available data is if we read till EOF. In practice this may not happen too frequently, but it's a race condition - it could happen at any point, and if the calling code doesn't account for this truncation, the resulting behavior is undefined.

I would really like to work around this SSH bug, but I think we're split between risking deadlock and truncating output.

jeremy-asher:
Reading until EOF would be easier, but I am pretty sure that's not necessary.  My reading of the way these calls work is that all available data has been read if you can guarantee wait has returned and then select on the read side of the pipe is executed and returned nil.  I may make an attempt on implementing this.  Is there a test that triggers the output truncation bug?

adrienthebo:
To reproduce, run a command that produces some known amount of output, and keep running it until you get truncated output, and you'll probably trigger the race condition more frequently by sleeping for a few seconds and then writing a massive glob of output.

With respect to your approach, it may work - mostly. After the child exits we can keep calling `#read_nonblock` from the pipe until we get EAGAIN or EWOULDBLOCK, and at that point we could consider the pipes drained. However this only works in the case where we have a grandchild process leaving the FD open that will never write anything else to the pipe. The only scenario where this could reasonably happen is with SSH using a control master. In normal cases if the FD is left open then it's possible that the child may write more information to the pipe, in which case we should continue to wait for more output.

The core issue is that SSH is breaking the semantics of EOF by leaving the pipe open and to make that work we have to pack in more workarounds into r10k. We did work around this in the past, but https://tickets.puppetlabs.com/browse/RK-18 indicated that this is pretty problematic.

If we can come up with a reasonably clean solution to work around the SSH bug I'm interested, but I would really prefer that we fix this issue upstream.

adrienthebo:
In addition, this is specific to OpenSSH itself; it doesn't look like libssh2 or jsch support the same approach to Control Masters.

jeremy-asher:
Okay, I will do some testing.  I guess there are semi-legitimate cases where a pipe could be held open by a grandchild process other than with this openssh issue, but that does suggest that the parent had little care for the output.  Reading until EOF is a convenient synchronization mechanism, but I don't think it's essential if we are willing to throw away any grandchild process output.  I'm not sure how else more data could get into the pipe after the child has exited and wait has returned, but I may be missing something.

I am also bothered by having to work around this issue, but even if it is fixed upstream, it seems unlikely that a non-security patch such as this will make it down to the distros most users are using any time soon.

adrienthebo:
How much of a performance boost are you seeing by using the Control Master?

jeremy-asher:
I think I quoted a 3x performance boost in the original issue/PR.  It's quite significant if you are tracking master on a large number of modules.

jeremy-asher:
I struggled to re-produce this truncation, but eventually hacked a sleep into the pump loop which triggered a race and a number of tests failed.  Try a sleep of even 0.001 just about anywhere in that loop and you should catch it.  I have an alternate fix to this PR that preserves the non-blocking IO behavior.  I'll send that shortly for your review.

andersonmills:
Given our stance on PR #287, I'm going ahead and merging this.

Issue #146, opened 2014-06-06
Possible to leverage SSH ControlMaster for git repos?


nvalentine-puppetlabs:
Sitting here watching r10k deploy an environment consisting of nearly 60 module git repos and wondering if it would be possible to utilize SSH's ControlMaster to re-use SSH connections and therefore improve speed of env build?

adrienthebo:
Is it possible? Theoretically. Is it feasible to implement in r10k? I'm not optimistic.

Setting up the ssh configuration to use ControlMaster looks to be a bit intrusive in that it requires touching a number of user configuration files, which is definitely outside of the primary scope of r10k. It looks like it could be implemented by end users that could set up the necessary configuration, and if I'm reading the documentation correctly it looks like all SSH connections started by Git/r10k could benefit. If this is the case then this could be a matter of documenting it.

I'm going to leave this open and target it at the 'unplanned' version, and I'll leave it open for about 3 months for community members to work on. If it doesn't see any activity within that time frame I'll close it out until there's interest expressed in implementing this feature.

jlambert121:
I just did some playing around with this in my environment and found there to be a negligible difference with and without ControlMaster for my environment.  81 modules total, 41 are in git.  On a rerun time went from 40.0s to 39.875, on a clean machine checkout time went from 4m 32.2s to 4m 17.3s.

nrvale0:
@jlambert121, that's rather disappointing. Was this for purge-and-rebuild of a single environment/branch?

jlambert121:
Single environment.  The longer time was a "rm -rf" on the environment first.

nrvale0:
I wonder if any benefit would manifest when doing a build out of all
branches in the control repo? Not that that is a common scenario.

Also, the particular environment I was working in when this idea occurred
to me has a Gitlab server which has surprisingly long connx setup times. I
should test there. Can you gist your setup?

-- via mobile

jeremy-asher:
I've seen a considerable benefit with using ControlMaster/ControlPersist on an environment with over 100 Git repos.  To get any benefit, however, I had to work around an issue with ssh holding stderr open (see #169).  Here's some typical times I'm seeing for re-deploying after an rm -fr on the environment so the cache repos were already up to date:

without persist:
real    2m57.505s
user    0m53.520s
sys     0m10.960s

with persist:
real    1m16.332s
user    0m49.100s
sys     0m10.730s

adrienthebo:
[...]

PR #287, opened 2015-01-27
[subprocess/runner] (RK-18) fix output truncation

jeremy-asher:
This is an alternate solution (to #285) which should fix JIRA issue [RK-18](https://tickets.puppetlabs.com/browse/RK-18) while maintaining compatibility with #169.  I had some difficulty reproducing the issue as written, but I was able to trigger a race condition consistently by adding a sleep to the loop in the Pump class's pump function and running the Runner::Posix spec tests.  This PR includes an additional test which helped with the debugging process.  I am unable to re-produce the problem with this change applied.  The implementation of this PR is explained in the commit message, but the essential issue is that the pump thread could terminate too early while data was still in the pipe due to a lack of synchronization.  Reading until EOF would be simpler, but would cause r10k to block when a grandchild process holds a pipe open (see #169).

### DIFF
--- a/lib/r10k/util/subprocess/runner/posix.rb
+++ b/lib/r10k/util/subprocess/runner/posix.rb
@@ -77,8 +77,8 @@ class R10K::Util::Subprocess::Runner::POSIX < R10K::Util::Subprocess::Runner
       _, @status = Process.waitpid2(pid)
     else
       _, @status = Process.waitpid2(pid)
-      @stdout_pump.halt!
-      @stderr_pump.halt!
+      @stdout_pump.wait
+      @stderr_pump.wait
       stdout = @stdout_pump.string
       stderr = @stderr_pump.string
     end
--- a/lib/r10k/util/subprocess/runner/pump.rb
+++ b/lib/r10k/util/subprocess/runner/pump.rb
@@ -21,22 +21,27 @@ class R10K::Util::Subprocess::Runner::Pump
     @io     = io
     @thread = nil
     @string = ''
-    @run    = true
     @min_delay = 0.001
     @max_delay = 1.0
+    @mutex = Mutex.new
+    @halting = false
+    @waiting = false
   end
 
   def start
     @thread = Thread.new { pump }
   end
 
+  # Exit immediately after the current pump cycle
   def halt!
-    @run = false
+    @halting = true
     @thread.join
   end
 
-  # Block until the pumping thread reaches EOF on the IO object.
+  # Block until the pumping thread reaches EOF or until select times out
+  # checking for updates on the IO object
   def wait
+    @mutex.synchronize { @waiting = true }
     @thread.join
   end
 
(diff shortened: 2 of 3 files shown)


adrienthebo:
I'm still concerned about how this change is only required for OpenSSH, and it's for a feature that I don't think is widely used. In addition all of the other issues related to this concluded that this is a bug in OpenSSH and should be fixed upstream, rather than pushing patches to all the downstream consumers. OpenSSH is the only SSH implementation that I know of that supports a ControlMaster, so if we switch to using libssh2 or something this won't be necessary.

Do you know if there are other r10k that rely on this feature? If it turns out that there is widespread use of the ControlMaster feature then I would be more interested, but I would like to know that this is going to affect, say, 2-5% of all r10k users before I pull it in.

jeremy-asher:
Agreed.  This does add a good amount of complexity to a major code path to deal with a single bug.  If you don't want to support this, I would argue that the non-blocking code should be entirely removed in favor of blocking IO for simplicity.

Is libssh2 in the future of r10k?  Doesn't this depend on the Git client?  I'm open to other suggestions on how to improve performance.  I haven't looked at Git over HTTPS, for example.  I suppose that could be faster than ssh if connections are re-used, but I don't think that is the case.

I can't say how many people use ControlPersist with r10k, but, considering the performance improvement, I would suggest it to everyone who tends to follow master rather than using tagged modules.  In my case, this makes the difference between tolerable and painfully slow deploy times.  If you want to support this feature, I would be happy to resolve the documentation issue (#146) as well.

adrienthebo:
With respect to blocking vs nonblocking I/O, we cannot do a synchronous read after the process has run or we run the risk of deadlocking. That being said if we can read till EOF then it'll be functionally equivalent to the synchronous read, sans the risk of deadlocking.

I'm looking at using libgit2 and rugged instead of shelling out, so the same protocols will be supported but it'll be a separate ssh implementation.

jeremy-asher:
So a blocking read in the separate thread would be unsafe?

Shelling out is a mess, so rugged/libgit2 sounds nice.  I took a quick look and didn't see any facility for connection multiplexing or keepalive which is a shame.  Are you looking at any other methods of improving performance?

adrienthebo:
I'm looking into making the ref type detection automatic so that if you specify a git tag or commit r10k can automatically figure that out. Are you mainly using tags or branches for your git modules?

adrienthebo:
I'm going to add a longer comment on this issue to explain more later, but in short I don't think we can merge this. I really worry about the correctness of this code, and since all other consumers of the OpenSSH code have rejected workarounds to this bug I don't think that we're on the right track. I am interested in looking into performance increases with r10k but I don't believe that this is the method.

adrienthebo:
Also, if you're seeing very slow performance with r10k when doing `r10k deploy environment -p`, there are a lot of other commands you can use to deploy less. https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/usage.mkd#deploying-environments documents the different ways that you can deploy a single environment or module. Long story short `r10k deploy environment -p` is a degenerate case, and while it's convenient to run it will always be dirt slow; you'll get much better performance increases by trying to update less.

jeremy-asher:
Okay, fair enough.  Thanks for providing some context.  For your reference, my use case is mostly branch refs (using :ref) and a few tags.  I do already deploy single environments or single modules only.  It would be interesting to get some data on what other r10k users' Puppetfiles tend to look like.

PR #289, opened 2015-01-28
(CODEMGMT-40) Automate Tests for "Basic Workflow" User Scenario

cowofevil:
Add tests that verify basic user scenario functionality.


integration/tests/user_scenario/basic_workflow/multi_env_custom_forge_git_module_static.rb
```diff
@@ -0,0 +1,110 @@
+require 'git_utils'
+require 'r10k_utils'
+require 'master_manipulator'
+test_name 'CODEMGMT-40 - C59223 - Single Environment with Custom, Forge and Git Modules'
```
zreichert: The English description is the same for both of these.  
Should this one read "Multiple Environments with Custom, Forge and Git Modules Deployed to Static Path"?

Issue #195, opened 2014-10-10
Feature Request?  multi-tenancy, with separate hieradata requires workaround.


brettswift:
I'm experimenting with multi-tenancy, and my use case is a bit unique.   This is a different approach to #101 if multiple applications (tenants) have the git-flow  release model for puppet code.

Goal: separate release cycles of applications along the git-flow branches. 

```
app1 control repo:  develop --> test --> staging --> prod
app1 hieradata repo:  develop --> test --> staging --> prod
app2 control repo:  develop --> test --> staging --> prod
app2 hieradata repo:  develop --> test --> staging --> prod
```

One way of doing this is locking each to a git hash (#101), but that can be cumbersome and requires everyone be on the same puppet file.

What I'd like to get working is multi-tenancy with r10k, so that the above workflow could work.   If hieradata is inside the same repo as the Puppetfile / aka _control repo_ then this is fine,  **but where it collides is separate hieradata repos**.
### Premise
- every "tenant" would have both a _control repo_ and a _hiera data_ repo.  
- **Hieradata** repo is also released in the git-flow style.  (We did this to allow easier hotfixing of data vs code). 
#### Single Tenant /etc/r10k.yaml

``` yaml
:sources:
  data:
    basedir: /etc/puppetlabs/puppet/hieradata
    prefix: false
    remote: ssh://git@stash/codo/hieradata.git
  puppet:
    basedir: /etc/puppetlabs/puppet/environments
    prefix: false
    remote: ssh://git@stash/codo/puppetconfig.git

:purgedirs:
  - /etc/puppetlabs/puppet/environments
  - /etc/puppetlabs/puppet/hieradata
```
#### multitenant /etc/r10k.yaml:

``` yaml
:sources:
  app1_h: 
    basedir: /etc/puppetlabs/puppet/hieradata
    prefix: true
    remote: ssh://git@stash/codo/hieradata.git
  app1_p:
    basedir: /etc/puppetlabs/puppet/environments
    prefix: true
    remote: ssh://git@stash/codo/puppetconfig.git

:purgedirs:
  - /etc/puppetlabs/puppet/environments
  - /etc/puppetlabs/puppet/hieradata
```

config that drives /etc/hiera.yaml

```
puppet::hiera_data_dir: "/etc/puppetlabs/puppet/hieradata/%{::environment}"
```

[...]

PR #295, opened 2015-02-05
[deployment] prefix support for multi-tenancy

brettswift:
Intended to solve #195 

This adds functionality to the **prefix** config variable, changing the type from _boolean_ to a  _boolean/String_ hybrid. 

This feature is backwards compatible - existing deployments of r10k should be unaffected, as true/false are still valid, and the default value is still false.

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -116,7 +116,7 @@ from. It may be any valid URL that the source may check out or clone. The remote
 must be able to be fetched without any interactive input, eg usernames or
 passwords cannot be prompted for in order to fetch the remote.
 
-```
+```yaml
 ---
 sources:
   mysource:
@@ -155,6 +155,10 @@ sources:
     basedir: '/etc/puppet/environments'
     prefix: true # All environments will be prefixed with "mysource_"
 ```
+#### prefix behaviour
+* if `true` environment folder will be prefixed with the name of the source.
+* if `false` (default) environment folder will not be prefixed
+* if `String` environment folder will be prefixed with the `prefix` value.
 
 Examples
 --------
@@ -224,3 +228,34 @@ This will create the following directory structure:
 `-- dev_loadtest     # dev repository, loadtest branch
 ```
 
+#### Multiple tenancy with external hieradata
+
+If hiera data is in a separate repository from your control repository, you
+must override the `prefix` so environment folders line up in both directories:
+
+```yaml
+---
+  sources:
+    main:
+      app1_data:
+        remote: 'git://git-server.site/my-org/app1-hieradata'
+        basedir: '/etc/puppet/hieradata'
+        prefix: "app1"
+      app1_modules:
+        remote: 'git://git-server.site/my-org/app1-puppet-modules'
+        basedir: '/etc/puppet/environments'
+        prefix: "app1"
+```
+
+
+This will create the following directory structure:
+
+```
+/etc/puppet/environments
+|-- app1_production  # app1 modules repository, production branch
+|-- app1_develop     # app1 modules repository, develop branch
+
+/etc/puppet/hieradata
+|-- app1_production  # app1 data repository, production branch
+|-- app1_develop     # app1 data repository, develop branch
+```
--- a/lib/r10k/source/base.rb
+++ b/lib/r10k/source/base.rb
@@ -12,8 +12,8 @@ class R10K::Source::Base
   attr_reader :name
 
   # @!attribute [r] prefix
-  #   @return [true, false] Whether the source name should be prefixed to each
-  #     environment basedir. Defaults to false
+  #   @return [String, nil] The prefix for the environments basedir.
+  #     Defaults to nil.
   attr_reader :prefix
 
   # Initialize the given source.
@@ -23,9 +23,9 @@ class R10K::Source::Base
   # @param options [Hash] An additional set of options for this source. The
   #   semantics of this hash may depend on the source implementation.
   #
-  # @option options [Boolean] :prefix Whether to prefix the source name to the
-  #   environment directory names. All sources should respect this option.
-  #   Defaults to false.
+  # @option options [Boolean, String] :prefix If a String this becomes the prefix.
+  #   If true, will use the source name as the prefix. All sources should respect this option.
+  #   Defaults to false for no environment prefix.
   def initialize(name, basedir, options = {})
     @name    = name
     @basedir = basedir
--- a/lib/r10k/source/git.rb
+++ b/lib/r10k/source/git.rb
@@ -42,8 +42,9 @@ class R10K::Source::Git < R10K::Source::Base
   # @param basedir [String] The base directory where the generated environments will be created.
   # @param options [Hash] An additional set of options for this source.
   #
-  # @option options [Boolean] :prefix Whether to prefix the source name to the
-  #   environment directory names. Defaults to false.
+  # @option options [Boolean, String] :prefix If a String this becomes the prefix.
+  #   If true, will use the source name as the prefix.
+  #   Defaults to false for no environment prefix.
   # @option options [String] :remote The URL to the base directory of the SVN repository
   # @option options [Hash] :remote Additional settings that configure how the
   #   source should behave.
--- a/lib/r10k/environment/name.rb
+++ b/lib/r10k/environment/name.rb
@@ -59,15 +59,27 @@ module R10K
       def dirname
         dir = @name.dup
 
-        if @prefix
-          dir = "#{@source}_#{dir}"
-        end
+        prefix = derive_prefix(@source,@prefix)
 
         if @correct
           dir.gsub!(INVALID_CHARACTERS, '_')
         end
 
-        dir
+        "#{prefix}#{dir}"
+      end
+
+
+      private
+
+      def derive_prefix(source,prefix)
+
+        if prefix == true
+          "#{source}_"
+        elsif prefix.is_a? String
+          "#{prefix}_"
+        else
+          nil
+        end
       end
     end
   end
--- a/spec/unit/environment/name_spec.rb
+++ b/spec/unit/environment/name_spec.rb
@@ -12,6 +12,16 @@ describe R10K::Environment::Name do
       bn = described_class.new('mybranch', :source => 'source', :prefix => true)
       expect(bn.dirname).to eq 'source_mybranch'
     end
+
+    it "prepends the prefix name when prefixing is overridden" do
+      bn = described_class.new('mybranch', {:prefix => "bar", :sourcename => 'foo'})
+      expect(bn.dirname).to eq 'bar_mybranch'
+    end
+
+    it "uses the branch name as the dirname when prefixing is nil" do
+      bn = described_class.new('mybranch', {:prefix => nil, :sourcename => 'foo'})
+      expect(bn.dirname).to eq 'mybranch'
+    end
   end
 
   describe "determining the validate behavior with :invalid" do
--- a/spec/unit/source/svn_spec.rb
+++ b/spec/unit/source/svn_spec.rb
@@ -95,6 +95,72 @@ describe R10K::Source::SVN, 'when prefixing is enabled' do
   end
 end
 
+describe R10K::Source::SVN, 'when prefixing is disabled' do
+  subject do
+    described_class.new(
+    'mysource',
+    '/some/nonexistent/dir',
+    {
+      :remote => 'https://svn-server.site/repo',
+      :prefix => false
+    }
+    )
+  end
+
+  describe "generating environments" do
+    before do
+      allow(subject.svn_remote).to receive(:branches).and_return %w[apache dns robobutler]
+    end
+
+    let(:environments) { subject.generate_environments }
+
+    it "creates an environment for each branch and the trunk" do
+      expect(environments.size).to eq(4)
+    end
+
+    it "does not prefix environments" do
+      expect(environments[0].dirname).to eq 'production'
+      expect(environments[1].dirname).to eq 'apache'
+      expect(environments[2].dirname).to eq 'dns'
+      expect(environments[3].dirname).to eq 'robobutler'
+    end
+  end
+end
+
+
+describe R10K::Source::SVN, 'when prefixing is overridden' do
+  subject do
+    described_class.new(
+    'mysource',
+    '/some/nonexistent/dir',
+    {
+      :remote => 'https://svn-server.site/repo',
+      :prefix => "tenant1"
+    }
+    )
+  end
+
+  describe "generating prefixed environments" do
+    before do
+      allow(subject.svn_remote).to receive(:branches).and_return %w[apache dns robobutler]
+    end
+
+    let(:environments) { subject.generate_environments }
+
+    it "creates an environment for each branch and the trunk" do
+      expect(environments.size).to eq(4)
+    end
+
+    it "prefixes the prefix name to environments" do
+      expect(environments[0].dirname).to eq 'tenant1_production'
+      expect(environments[1].dirname).to eq 'tenant1_apache'
+      expect(environments[2].dirname).to eq 'tenant1_dns'
+      expect(environments[3].dirname).to eq 'tenant1_robobutler'
+    end
+  end
+end
+
+
 describe R10K::Source::SVN, 'registering as a source' do
   it "registers with the :svn key" do
     expect(R10K::Source.retrieve(:svn)).to eq described_class


lib/r10k/source/base.rb
```diff
@@ -23,16 +23,27 @@ class R10K::Source::Base
...
   #
-  # @option options [Boolean] :prefix Whether to prefix the source name to the
+  # @option options [Boolean | String] :prefix Whether to prefix the source name to the
   #   environment directory names. All sources should respect this option.
   #   Defaults to false.
   def initialize(name, basedir, options = {})
     @name    = name
     @basedir = basedir
-    @prefix  = options.delete(:prefix)
+    @prefix  = derive_prefix(name,options.delete(:prefix))
```
adrienthebo: style: should be `...name, options.delete...`

lib/r10k/source/base.rb
```diff
@@ -23,16 +23,27 @@ class R10K::Source::Base
...
     @options = options
   end
 
+  def derive_prefix(name,prefix)
+    if prefix == true || prefix == 'true'
```
adrienthebo: The special handling of `'true':String` is inconsistent with the documentation.

In the case where someone had a truthy value that wasn't true this would change the behavior, but I don't think that case is worth worrying about.

spec/unit/source/git_spec.rb
```diff
@@ -139,6 +139,36 @@
     end
   end
 end
+#
+# describe R10K::Source::Git, 'when prefixing is enabled, and the prefix name is overridden' do
```
adrienthebo: These tests need to be implemented or deleted, committing in commented out code doesn't provide any value.

lib/r10k/source/base.rb
```diff
@@ -23,16 +23,27 @@ class R10K::Source::Base
   # @param options [Hash] An additional set of options for this source. The
   #   semantics of this hash may depend on the source implementation.
   #
-  # @option options [Boolean] :prefix Whether to prefix the source name to the
+  # @option options [Boolean | String] :prefix Whether to prefix the source name to the
```
adrienthebo: style: multiple return values should be comma separated.

adrienthebo: This comment is now out of date; it doesn't specify the behavior of a string prefix.

spec/unit/source/git_spec.rb
```diff
@@ -157,9 +187,9 @@
       expect(bn.dirname).to eq 'mybranch'
     end
 
-    it "prepends the source name when prefixing is on" do
-      bn = described_class.new('mybranch', {:prefix => true, :sourcename => 'foo'})
-      expect(bn.dirname).to eq 'foo_mybranch'
+    it "prepends the prefix name when prefixing is overridden" do
```
adrienthebo: This should be a new test case and should not replace the existing test case.

spec/unit/source/git_spec.rb
brettswift: oops.  my bad.   I thought I reviewed my stuff well enough to catch this.

spec/unit/source/git_spec.rb
brettswift: The flow here changed a bit.   `true` won't actually make it into the `BranchName` class anymore, due to the `derive_prefix` function in source/base.pp

https://github.com/brettswift/r10k/blob/feature/master/multi_tenant_prefix/lib/r10k/source/base.rb#L36-L44

So in BranchName, the only possible values are `false` or `:String`. 

I'm updating the documentation for `initialize` in this class to reflect that.

lib/r10k/source/base.rb
brettswift: Corrected documentation.

I see what you mean about truthy values.   The fix here would be to check 

`if (prefix.is_a?(TrueClass) && prefix)` 

 What's the best practice here?   I don't mind changing it, let me know what is best.

spec/unit/source/git_spec.rb
brettswift: The tests that articulate this behaviour are found here: 

https://github.com/brettswift/r10k/blob/feature/master/multi_tenant_prefix/spec/unit/source/base_spec.rb#L33-L50

You'll see that the only output is false or string. 

It's a little bit unintuitive than my original fix which was done at the `source/git.pp`  but doing it at `source/base.pp` is a little more appropriate I think.

What do you think about this flow?

lib/r10k/source/base.rb
elyscape: You could just use:

``` ruby
if prefix == true
```

Truthy values aren't coerced into booleans, so this will only succeed if `prefix` is actually `true`.

doc/dynamic-environments/configuration.mkd
```diff
@@ -121,6 +121,10 @@ sources:
     basedir: '/etc/puppet/environments'
     prefix: true # All environments will be prefixed with "mysource_"
 ```
+#### prefix behaviour
+* `true` prefix will be set to the `source` name.
+* `[string]` prefix will be set to the
```
elyscape: This sentence is incomplete.

lib/r10k/source/base.rb
```diff
@@ -12,7 +12,7 @@ class R10K::Source::Base
   attr_reader :name
 
   # @!attribute [r] prefix
-  #   @return [true, false] Whether the source name should be prefixed to each
+  #   @return [true, false, String] Whether the source name should be prefixed to each
```
elyscape: Documentation does not specify what a `String` value would mean.  It might make more sense to change this to being either a `String` or `nil`, with the value being the prefix name.

lib/r10k/source/base.rb
```diff
@@ -23,16 +23,27 @@ class R10K::Source::Base
   # @param options [Hash] An additional set of options for this source. The
   #   semantics of this hash may depend on the source implementation.
   #
-  # @option options [Boolean] :prefix Whether to prefix the source name to the
+  # @option options [Boolean, String] :prefix Whether to prefix the source name or prefix name to the
```
elyscape: Unclear. Should probably read

> If a String, the prefix name for the environment directory names. Otherwise, whether to use the source name as the prefix. All sources should respect this option.

lib/r10k/source/git.rb
```diff
@@ -44,7 +44,7 @@ class R10K::Source::Git < R10K::Source::Base
   # @param options [Hash] An additional set of options for this source.
   #
   # @option options [Boolean] :prefix Whether to prefix the source name to the
-  #   environment directory names. Defaults to false.
+  #   environment directory names. Defaults to false. Will be a String of the prefix otherwise.
```
elyscape: Should be brought in line with the change in base.rb and have the type changed to `[Boolean,String]`.  Also, as per [my comment on that change](https://github.com/puppetlabs/r10k/pull/295/files#r24205559), this would be clearer if it read:

> If a String, the prefix name for the environment directory names. Otherwise, whether to use the source name as the prefix. Defaults to false.

lib/r10k/source/base.rb
brettswift: Yea I agree now that you mention it.  A possible return val of false implies true might be possible.  I'm changing this to nil / String.

lib/r10k/source/base.rb
brettswift: Yes, that was some brilliant Engrish on my part.   :)  Thanks for finding it.

brettswift:
@adrienthebo just trying to plan out the adoption of this feature on our roadmap.   Does this look feasible for 1.5 still?  And if you have a ballpark date on 1.5 release, that'd be helpful. 

Let me know if you have more feedback here.

adrienthebo:
@brettswift I have less leeway about what goes in and doesn't go in to any particular release since it also affects QA and our planning; I'll see what I can work out on this issue.

brettswift:
Understood!  Thanks for the update!

adrienthebo:
@brettswift I'm going to try to pull this into our upcoming sprint which starts on the 11th; could you do me a solid and create an issue at https://tickets.puppetlabs.com/browse/RK with the contents of GH-195 so I can track this along with the rest of the tickets?

brettswift:
@adrienthebo   you bet, here you go https://tickets.puppetlabs.com/browse/RK-35

lib/r10k/source/base.rb
```diff
@@ -23,16 +23,27 @@ class R10K::Source::Base
...
+  #   Defaults to false for no environment prefix.
   def initialize(name, basedir, options = {})
     @name    = name
     @basedir = basedir
-    @prefix  = options.delete(:prefix)
+    @prefix  = derive_prefix(name, options.delete(:prefix))
     @options = options
   end
 
+  def derive_prefix(name,prefix)
```
adrienthebo: This should be a private method.

lib/r10k/source/base.rb
```diff
@@ -65,4 +65,17 @@ def accept(visitor)
       end
     end
   end
+
+  private
+
+  def derive_prefix(name,prefix)
```
adrienthebo: So now that you've changed this to how I asked, I want you to change it again! Yay! :)

Per your comment on https://github.com/puppetlabs/r10k/pull/295/files#r24181790, the BranchName class will never be passed true, only false and a string. However the BranchName class is currently encapsulating most of the logic of naming branches, but this code touches branch naming as well. Instead of deriving the prefix here, why don't we push that into the BranchName class and just pass it through here? It'll help decouple things and prevent the Source classes from having to know too much about how prefixing is implemented.

In addition, both the Git and SVN sources have a BranchName method but I think there's no difference between them; it might be reasonable to extract the common code into `R10K::Source::BranchName` and deduplicate those classes.

adrienthebo: For the record, in Source::Git there's one place where prefix is checked, but it's dead code; https://github.com/puppetlabs/r10k/pull/341 will remove that.

brettswift: Ok I'll try to look at this soon.  Might find time tomorrow.

brettswift:
just pushed those changes.  In working with it I think the "invalid_branches"  thing is a little weird and could be decoupled.  The logic of errors and auto correction is coupled by that one flag, and some of the variables are misleading.   IE:  setting it to 'error'  doesn't actually error inside the BranchName class, only to return invalid state.   The error happens elsewhere.

However,  I've left that for now as it's a little off topic for this PR.   Hopefully the refactoring will make it easier to work with in the future.

BranchName is extracted now, although the `opts[:remote]` property is a tiny seepage of the svn class into the BranchName class.   It's not _used_ but only stored for reference later.    I didn't think that was a huge thing but thought I'd mention it.

I welcome more input!

doc/dynamic-environments/configuration.mkd
```diff
@@ -121,6 +121,10 @@ sources:
     basedir: '/etc/puppet/environments'
     prefix: true # All environments will be prefixed with "mysource_"
 ```
+#### prefix behaviour
+* if `true` environment folder will be prefixed with the `source` name.
```
adrienthebo: This reads as if it's literally "source" - might be good to rephrase to something like "the name of the source".

doc/dynamic-environments/configuration.mkd
```diff
@@ -121,6 +121,10 @@ sources:
...
+* if `String` environment folder will be prefixed with the `prefix` value.
+
```
adrienthebo, line 162: No mention of a false value, or default value.

lib/r10k/source/base.rb
adrienthebo: I thought I commented on this before but it doesn't look like I did; since this deals with branch name munging I would like this responsibility to be moved to the BranchName class so all the name manipulation is in a single place.

Given this change, the type of `@prefix` will be `[Boolean, String]`.

lib/r10k/source/git.rb
```diff
@@ -43,8 +43,9 @@ class R10K::Source::Git < R10K::Source::Base
   # @param basedir [String] The base directory where the generated environments will be created.
   # @param options [Hash] An additional set of options for this source.
   #
-  # @option options [Boolean] :prefix Whether to prefix the source name to the
-  #   environment directory names. Defaults to false.
+  # @option options [Boolean, String] :prefix A String value to prefix the environment directory name.
```
adrienthebo: This first line is misleading - best indicate that if it's a String, X occurs, else if it's a Boolean, Y occurs.

lib/r10k/source/branchname.rb
```diff
@@ -0,0 +1,112 @@
...
+require 'r10k/util/core_ext/hash_ext'
+
+# This class controls the behaviour of branch names
+#
+# A branch name becomes the name of the environment on the file system.
+# The default branch name is the name of the source branch.
+# The name can be modified with a prefix, and by default will be automatically
+# modified if the branch name contains characters invalid for a file system.
+#
+class R10K::Source::BranchName
```
adrienthebo: The `R10K::Source` namespace deals with collecting and creating environments, while BranchName is specific to a single environment. I would rather this live in the `R10K::Environment` namespace.

lib/r10k/source/branchname.rb
```diff
@@ -0,0 +1,112 @@
+require 'r10k/git'
```
adrienthebo: None of these required files are used in this class.

lib/r10k/source/branchname.rb
```diff
@@ -0,0 +1,112 @@
...
+class R10K::Source::BranchName
+
+  # @!attribute [r] name
+  #   @return [String] The raw name of the branch directly from the source
+  attr_reader :name
+
+  # @!attribute [r] remote
+  #   @return [String] The url of the svn remote. Only used to storage and
+  #     retrieval when working with this class.
+  attr_reader :remote
```
adrienthebo: This variable looks like it's coming from the `SVN::BranchName` class, but that class never used that value. We can omit it here.

spec/unit/source/base_spec.rb
```diff
@@ -28,4 +28,5 @@
       subject.accept(visitor)
     end
   end
+
```
adrienthebo: Spurious whitespace change

lib/r10k/source/branchname.rb
adrienthebo: Wait, that might be wrong on my part. Stay tuned.

adrienthebo:
There's some fancy footwork that needs to be done for the BranchName object; I'll do that soonish to clean up the API in a few ways. That should simplify your end of things.

lib/r10k/source/branchname.rb
brettswift: Yea, when I first started refactoring I had the exact same thought process.  I nuked it, then it broke the svn class. 

I think I mentioned that in my comment above.   It could be made into a more generic hash, or SVN could handle things better and not need to push anything into BranchName just for storage. 

I let it go on first round of refactoring, not sure if SVN will be end of life at some point? : )

lib/r10k/source/git.rb
brettswift: Ok.  I can't remember what I originally had,  I had changed this from previous feedback.  I'll clean it up. 

The same sentence is copied in these files: 
- environment/branchname.rb (after it's moved)
- source/base.rb
- source/git.rb  (err... uh.. i'll remove the docs from this class.    Honestly this is some of the first ruby code I've ever written so I'm not sure how the rdocs will behave but this might end up in duplicate documentation if it's also in the base class?)

lib/r10k/source/branchname.rb
brettswift: uh, scratch that too :), I see the rest of the args are documented as well - i'll fix all three locations.

lib/r10k/source/git.rb
brettswift: uh, scratch that too :), I see the rest of the args are documented as well - i'll fix all three locations.

adrienthebo:
I've pushed https://github.com/puppetlabs/r10k/pull/359 to deduplicate the name behavior and clean up the SVN remote oddity, could you eyeball it and see if you could base your work on that? Don't worry about changing anything just yet; I just want to make sure that the changes in that PR will simplify what you need to do for this PR.

adrienthebo:
https://github.com/puppetlabs/r10k/pull/341 also needs to get merged to deal with more prefix oddities but should be straightforward to do.

brettswift:
I have merged #359 on a local branch to test, and have the spec tests working.  It does clean things up quite a bit.   Thanks. 

I'll wait for it to get merged to master and then update this PR.

adrienthebo:
Sounds good. Since I wrote #359 and it does have some functional changes I can't merge it, so it might be a little while until it's merged.

adrienthebo:
#359 was merged so you're good to go on this.

adrienthebo:
Looks good to me, @brettswift could you rebase this into a single commit so that we can merge it? Thanks!

brettswift:
Pretty sure it is the same difference, but I had problems rebasing, so just did a soft reset off the merge-base on my branch.   

Thanks for the help on this, and getting it in quick!

adrienthebo:
No worries, history looks good now. Could you add a little information to the commit message explaining what this is for? You can use something like https://github.com/puppetlabs/r10k/commit/178d4c7d549054fbeb510bf80b10ff048b52969a as a reference.

brettswift:
Does that message make sense?

adrienthebo:
Looks good to me, as soon as Travis passes (which could be two hours given past trends) I'll merge this in. Thanks!

brettswift:
I guess either wait for Travis,  or for Dos Equis hour to come around. :)

adrienthebo:
Merged in 910709a; thanks for the contribution!

PR #296, opened 2015-02-05
(CODEMGMT-63) Automate Tests for Negative "Basic Workflow"

cowofevil:
Add more negative basic workflow user scenario integration tests.
Also, refactor the "01_pe_r10k.rb" pre-suite to use new functions for
creating r10k git sources.

### DIFF
--- a/integration/tests/user_scenario/basic_workflow/negative/neg_invalid_env_name.rb
+++ b/integration/tests/user_scenario/basic_workflow/negative/neg_invalid_env_name.rb
@@ -1,0 +1,34 @@
+require 'git_utils'
+require 'r10k_utils'
+require 'master_manipulator'
+test_name 'CODEMGMT-63 - C62511 - Attempt to Deploy Environment Containing Invalid Character in Name'
+
+#Note: this is not much of a negative test because it actually works.
+#Waiting on CODEMGMT-65 so that this actually fails.
+
+#Init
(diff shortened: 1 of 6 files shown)


integration/tests/user_scenario/basic_workflow/negative/neg_invalid_env_name.rb
```diff
@@ -0,0 +1,34 @@
...
+step "Create \"#{invalid_env_name}\" Branch from \"production\""
+git_on(master, 'checkout production', git_environments_path)
+git_on(master, "checkout -b #{invalid_env_name}", git_environments_path)
+
+step "Push Changes to \"#{invalid_env_name}\" Environment"
+git_push(master, invalid_env_name, git_environments_path)
+
+#Tests
+step 'Attempt to Deploy via r10k'
+on(master, 'r10k deploy environment -v') do |result|
```
zreichert, line 32: Shouldn't you include :acceptable_exit_codes => 1 ?

cowofevil, line 32: @zreichert Currently no. The r10k deployment is currently working because of CODEMGMT-65. I wanted to use the "expected_fail" Beaker DSL, but it isn't available yet. So this test is currently passing, but when CODEMGMT-65 it will start failing which is what I want to happen.

PR #274, opened 2014-12-29
(#273) Emphasize deploy keys for the r10k user in README, faq, and quickstart docs

rnelson0:
See #273

PR #297, opened 2015-02-06
Missing sources

rnelson0:
Acknowledge bug in RK-4 rather than tying to #274


rnelson0:
Ready for review

adrienthebo:
The bug report contains a description of the issue and serves as documentation; we don't need to duplicate the information in the docs themselves.

adrienthebo:
Also, [googling](https://www.google.be/search?q=r10k+sources+map+nil) for this error returns RK-4 as the first link which should point people in the right direction.

PR #298, opened 2015-02-06
(CODEMGMT-72) Update Pre-suite and Add Test Run Scripts

cowofevil:
The pre-suites for integration testing do not have TestRail test case ID in
the "test_name" so we can't automatically record test results using Telly.
First I had to break the pre-suites into smaller chunks and then add
the appropriate TestRail test case ID. Also, I added a new batch of test run
scripts to assist in test execution.


cowofevil:
@zreichert Please review #299 and #300 first before this one. After those two get merged then I can rebase this PR and you should be able to get passing runs.

cowofevil:
@zreichert Okay, I rebased this PR. Force pull and you should be good to go.

PR #307, opened 2015-02-13
Don't link to the librarian-puppet Puppetfile docs

juniorsysadmin:
The acceptable Puppetfile format defers slightly for r10k compared to librarian-puppet.
eg. puppetlabs/ntp compared to puppetlabs-ntp
This patch links to the r10k docs instead to avoid confusion.


adrienthebo:
Thanks!

PR #309, opened 2015-02-18
(RK-23) Update for the use of PMT metadata library

andersonmills:
Before this commit, R10K used it's own metadata parsing algorithms for
modules from the Forge. After this commit, we are using shared code from
Puppet to parse the metadata files.

WARNING: This commit has two big problems that will be resolved in future PRs.
1 - Semantic should be a gem which R10K depends upon. That should be
handled soon.
2 - The shared code from puppet is currently in lib/shared/puppet.


lib/r10k/module/metadata_file.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  # Does the metadata file itself exist?
+  def exist?
+    @metadata_file_path.file? and @metadata_file_path.readable?
+  end
+
+  # Attempt to read the metadata file and return the Puppet::ModuleTool::Metadata object
+  def read(metadata_file_path = @metadata_file_path)
+    if self.exist?
+      File.open(metadata_file_path) do |f|
```
adrienthebo: This can use `@metadata_file_path.open` instead of `File.open(path)` since `@metadata_file_path` is a Pathname object.

lib/r10k/module/metadata_file.rb
```diff
@@ -0,0 +1,31 @@
...
+        begin
+          metadata = Puppet::ModuleTool::Metadata.new
+          #returns Metadata.self
+          metadata.update(JSON.load(f))
+        rescue JSON::ParserError
+          #raise ArgumentError, "Could not parse JSON #{metadata_path}", ex.backtrace
+          false
```
adrienthebo: This seems to indicate that `#read` has a return value, if so it would be good to document that.

lib/r10k/module/metadata_file.rb
```diff
@@ -0,0 +1,31 @@
...
+  # Attempt to read the metadata file and return the Puppet::ModuleTool::Metadata object
+  def read(metadata_file_path = @metadata_file_path)
+    if self.exist?
+      File.open(metadata_file_path) do |f|
+        begin
+          metadata = Puppet::ModuleTool::Metadata.new
+          #returns Metadata.self
+          metadata.update(JSON.load(f))
+        rescue JSON::ParserError
+          #raise ArgumentError, "Could not parse JSON #{metadata_path}", ex.backtrace
```
adrienthebo: This is swallowing an exception, so it might be good to log a message at debug or debug1 to indicate that the load failed.

lib/semantic.rb
```diff
@@ -0,0 +1,7 @@
+module Semantic
+  autoload :Version, 'semantic/version'
```
adrienthebo, line 2: Since this is Semantic code this is an FYI and not a "this needs to be fixed", but Matz indicated that autoload is [deprecated](https://www.ruby-forum.com/topic/3036681) and should be avoided.

lib/r10k/module/forge.rb
```diff
@@ -61,10 +63,9 @@ def expected_version
     @expected_version
   end
 
-
   # @return [R10K::SemVer] The version of the currently installed module
   def current_version
-    @metadata.read
+    #@metadata.read(@path + 'metadata.json') #Why would it not be already read?
```
adrienthebo: https://github.com/puppetlabs/r10k/commit/a20870fdd551e7b0cfb9b3c23a0cb42393c4f201 - the metadata file would not automatically be read when instantiated, which meant that the version field would be nil until that read. Not great behavior, and I think that's an artifact of how I implemented the metadata file. As long as the metadata fields will be automatically populated when we read this, we can omit the explicit read.

lib/r10k/module/forge.rb
```diff
@@ -89,7 +90,7 @@ def status
     if not self.exist?
       # The module is not installed
       return :absent
-    elsif not @metadata.exist?
+    elsif not File.exist?(@path + 'metadata.json')
```
adrienthebo, line 92: Since `@path + 'metadata.json'` will create a Pathname object, we could do `(@path + 'metadata.json').exist?` or create the object and invoke exist on it. Stylistic change, do what you think is more readable.

lib/r10k/module/metadata_file.rb
adrienthebo: Looking more closely at the previous code indicates that this might just be a remnant of that, feel free to remove the `false`.

lib/r10k/module/forge.rb
andersonmills: Gotcha. Certainly for the forge class here, metadata is now read on initialization.

lib/r10k/module/forge.rb
andersonmills, line 92: I'm still learning when Ruby-ism is correct and when it's over done, but the calling a method on the object result of an addition feels a little weird.

lib/r10k/module/metadata_file.rb
andersonmills: Happy to change this one.

lib/r10k/module/metadata_file.rb
andersonmills: Happy to log. Chose debug.

lib/r10k/module/metadata_file.rb
andersonmills: Removed. forge.rb is currently the only place reading Metadata. That will be reworked in CODEMGMT-5, which is up next.

lib/semantic.rb
andersonmills, line 2: Noted and added to personal notes about PE-3118.

lib/r10k/module/forge.rb
adrienthebo, line 92: Fair point, disregard this comment.

adrienthebo:
+1 from me, LMK if you want to merge or if you want to keep this as a long running integration branch.

andersonmills:
I think for now, that I don't want to merge this until I get a dependent Forge integration PR ready which keep this from having so many failing tests in forge_spec.rb.

andersonmills:
Okay, based on a planning meeting today, I DO need to get the forge_spec.rb tests passing.

adrienthebo:
Tested installing, upgrading, downgrading, switching to and from Git modules, all is good. Will merge when Travis completes.

adrienthebo:
Travis-ci is stalled; I'm testing and merging this locally.

adrienthebo:
Merged in 128537c.

PR #312, opened 2015-02-18
Don't include spec/integration in gemspec files

adrienthebo:
The unicode files in the `integration` directory are causing bundler to complain about the gemspec being invalid, and specs aren't really used as part of the installed gem. We can ignore both.


cowofevil:
@adrienthebo FYI, we are no longer placing Unicode files in the repo. Fixed in #314.

PR #314, opened 2015-02-18
(CODEMGMT-87) Fix Unicode Tests

cowofevil:
The Unicode file stored in the Git repository causes bundler to fail.
The file has been renamed to ASCII to prevent the bundler failure. The
associated tests has been updated to rename the file on copy to Unicode.


cowofevil:
@andersonmills Ping.

PR #317, opened 2015-02-23
(RK-22) Add feature detection classes

adrienthebo:
This pull request adds a registry of sorts for detection features. This will be used to determine if things like libgit2 is available, or the git binary is present on the system, in order to select the correct implementation to use.

### DIFF
--- a/lib/r10k/features.rb
+++ b/lib/r10k/features.rb
@@ -1,0 +1,13 @@
+require 'r10k/feature/collection'
+require 'forwardable'
+
+module R10K
+  module Features
+    @features = R10K::Feature::Collection.new
+
+    class << self
+      extend Forwardable
+      def_delegators :@features, :add, :available?
+    end
+  end
+end
--- a/spec/unit/feature_spec.rb
+++ b/spec/unit/feature_spec.rb
@@ -13,4 +13,38 @@ describe R10K::Feature do
       expect(feature.available?).to be_falsey
     end
   end
+
+  describe "confining a feature to a block" do
+    it "is available if the block is true" do
+      feature = described_class.new(:blockfeature) { true }
+      expect(feature.available?).to be_truthy
+    end
+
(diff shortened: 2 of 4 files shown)


lib/r10k/features.rb
```diff
@@ -0,0 +1,5 @@
+require 'r10k/feature/collection'
+
+module R10K
+  Features = R10K::Feature::Collection.new
```
adrienthebo: In the long term we need a better way to handle singletons like this, but I don't think that now is the time to implement something more complex. Is this sufficent?

 For reference, this would be used like so:

``` ruby
# Adding features

R10K::Features.add(:rugged, :libraries => 'rugged')

R10K::Features.add(:shellgit) do
  !!R10K::Util::Commands.which('git')
end

# Checking for features

R10K::Features.available?(:rugged)
```

andersonmills: This is clever. I'm not sure we want clever?

Yeah, not crazy about using a constant like that.

adrienthebo:
@andersonmills updated to use a more common method of handling singletons, per our conversation.

spec/unit/feature_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      expect(feature.available?).to be_falsey
+    end
+
+    it "is unavailable if the block returns false and libraries are present" do
+      feature = described_class.new(:nope, :libraries => 'r10k') { false }
+      expect(feature.available?).to be_falsey
+    end
+
+    it "is available if the block returns true and libraries are present" do
+      feature = described_class.new(:nope, :libraries => 'r10k') { true }
```
andersonmills: LOL. The feature name is this case is misleading. :)

adrienthebo: /cough fixing

andersonmills:
:+1: I'm happy with this, how it works, and the testing of it.

PR #318, opened 2015-02-23
Convert Git integration specs to shared examples, deduplicate spec setup

adrienthebo:
This pull request extracts all of the Git integration specs to shared examples for reuse later, and removes duplicate setup steps from the integration specs into a single shared context.

PR #319, opened 2015-02-23
(RK-17) Move Shell based git implementation to namespace

adrienthebo:
This moves all Git classes that are based on shelling out to Git into a separate namespace, pending the implementation of a Rugged based Git implementation. It does not add any new behavior and just renames/renamespaces files.

This pull request is based on GH-318.


andersonmills:
Yep. I like it. :+1:

PR #323, opened 2015-02-24
Add Settings::Container#reset!, set up inheritance relationship

adrienthebo:
In order to add a rugged based Git implementation we need to subclass R10K::Git::Cache, which relies on having settings applied to the subclass. This adds adding some cleanup functionality for the tests for those examples, and reconfigures the settings mixin to properly handle inheritance of settings values.


andersonmills:
Tested and happy with it.  :+1:

PR #324, opened 2015-02-24
RK-17 Make R10K::Git::Cache an abstract class

adrienthebo:
In order to make sure that separate Git implementations don't accidentally intermix by sharing Cache objects, we need to subclass Git::Cache and use the appropriate concrete implementation. When the libgit2 implementation is merged we can switch from instantiating a specific subclass to a factory method that'll select the appropriate implementation.

### DIFF
--- a/lib/r10k/git/cache.rb
+++ b/lib/r10k/git/cache.rb
@@ -1,5 +1,4 @@
 require 'r10k/git'
-require 'r10k/git/shellgit/bare_repository'
 
 require 'r10k/settings'
 require 'r10k/instance_cache'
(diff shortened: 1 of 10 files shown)


adrienthebo:
This is based on GH-323.

adrienthebo:
Since GH-323 has been merged this has been rebased on master; @andersonmills ready for your review.

andersonmills:
:+1: I am happy with it. Git::Cache now calls out to the Git::ShellGit::Cache implementation in the ShellGit specific repositories.

PR #327, opened 2015-02-26
(CODEMGMT-95) Fix Use of "production" Environment

cowofevil:
Currently, when integration tests run against the "production" environment it
is assumed the default environment is automatically set on the agent. Because
of PE-7069 this assumption is no longer valid and all tests referencing the
"production" environment implicitly need to be changed to be explicit.


zreichert:
Should this line be updated as well?
https://github.com/puppetlabs/r10k/pull/327/files#diff-28c36dbf27c43d0a2a0eb7fa60ec6484L84

cowofevil:
Ahh, I missed one! Yes, I'll need to submit another commit.

cowofevil:
@zreichert Fixed.

PR #330, opened 2015-03-02
(RK-27) Move request to find latest release to v3 forge api.

adreyer:
What should happen when r10k gets a 404 from forge. I believe the old API would return an empty release set for modules that don't exit? This will now raise a R10K::Error

Before this r10k made a request to the deprecated unversioned forge
API to find the latest release of a module. After this it will get
that information from the v3 API. If a user has forge.puppetlabs.com
as the forge in a Puppetfile it will be changes to
forgeapi.puppetlabs.com and the user will be warned.


adrienthebo:
:+1:

andersonmills:
Any reason not to change the default forge in Puppetfiles to forgeapi.puppetlabs.com? See lib/r10k/puppetfile.rb line 40.

https://github.com/puppetlabs/r10k/blob/master/lib/r10k/puppetfile.rb#L40

adrienthebo:
Ooh. Missed that. :+1: to that. Those code paths aren't active but I think we should reconcile that now rather than leave the code in an inconsistent state. That, or remove `@forge` since it's not used right now.

adreyer:
fixed

andersonmills:
I believe in it! :+1:

PR #326, opened 2015-02-24
(RK-17) Add rugged based Git implementation

adrienthebo:
This pull request adds a rugged based implementation of all of the Git classes. The rugged and shellgit based implementations should have identical interfaces and run the same integration specs, so hopefully transitioning from one to another should be seamless.

TODO: figure out how rugged runs ssh, because that might not be seamless.

This code will crash and burn on Ruby 1.8.7 because rugged doesn't support that; disabling the rugged classes on 1.8.7 will come later.

This PR is based on #322, #323, and #324.

PR #332, opened 2015-03-04
(RK-17) Detect presence of Git binary and use ShellGit implementation.

adrienthebo:
This pull request uses the feature detection code to check for the presence of the Git binary and use the appropriate Git provider. Since there's only one implementation this is pretty limited in scope but provides the base for adding additional implementations.

This is required for GH-326 so that the rugged based Git implementation is bypassed on Ruby 1.8.7.

### DIFF
--- a/lib/r10k/util/commands.rb
+++ b/lib/r10k/util/commands.rb
@@ -1,0 +1,31 @@
+module R10K
+  module Util
+    module Commands
+      module_function
+
+      # Find the full path of a shell command.
+      #
+      # On POSIX platforms, the PATHEXT environment variable will be unset, so
+      # the first command named 'cmd' will be returned.
+      #
+      # On Windows platforms, the PATHEXT environment variable will contain a
+      # semicolon delimited list of executable file extensions, so the first
+      # command with a matching path extension will be returned.
+      #
+      # @param cmd [String] The name of the command to search for
+      # @return [String, nil] The path to the file if found, nil otherwise
+      def which(cmd)
+        exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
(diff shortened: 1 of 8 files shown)


lib/r10k/util/commands.rb
```diff
@@ -0,0 +1,30 @@
...
+    module Commands
+      module_function
+
+      # Find the full path of a shell command.
+      #
+      # On POSIX platforms, the PATHEXT environment variable will be unset, so
+      # the first command named 'cmd' will be returned.
+      #
+      # On Windows platforms, the PATHEXT environment variable will contain a
+      # semicolon delimited list of executable file extensions, so
```
andersonmills: Was this sentence cut off?

adrienthebo: I accidentally the rest of the sentence. Fixed.

andersonmills:
Is there something that can be done with the crazy overloading of environment here? lib/r10k/util/environment.rb

And actually, can you explain to me why environment.rb is in this pull request?

adrienthebo:
'environment' is up there with 'root' as one of the most overloaded terms in POSIX; the class is for dealing with process environment variables so there's a collision. `Process` could work.

As for why, I need it to test the `.which` function and expect that I'll need it down the road for other things; it's a simple function that I thought I would just nab and implement now.

andersonmills:
Ah, I didn't see the use of it in the tests. I was looking for it being used in the main code.

I'm not too worried about the name, but noted the collision with Puppet's "environment."

adrienthebo:
Naming is hard, let's go ride bikes. s/environment/exec_env/

andersonmills:
I believe in it and believe the tests are testing what we want it to.
:+1:

PR #333, opened 2015-03-06
(CODEMGMT-92) Add Tests for Git Source Protocols

cowofevil:
Create integration tests for Git source protocols. Fix bug in git pre-suite
where the default branch was not set to "production". Fix bug in r10k utils
where "agents" host array was specified instead of "master" host.

### DIFF
--- a/integration/tests/git_source/git_source_submodule.rb
+++ b/integration/tests/git_source/git_source_submodule.rb
@@ -1,0 +1,69 @@
+require 'git_utils'
+require 'r10k_utils'
+require 'master_manipulator'
+test_name 'CODEMGMT-92 - C59238 - Single Git Source with Git Sub-module'
+
+#Init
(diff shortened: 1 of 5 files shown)


integration/tests/git_source/git_source_submodule.rb
```diff
@@ -0,0 +1,69 @@
+require 'git_utils'
+require 'r10k_utils'
+require 'master_manipulator'
+test_name 'CODEMGMT-92 - C59238 - Single Git Source with Git Sub-module'
```
zreichert, line 4: Is this the correct test?  The ticket lists 59233 as a test that is included in this PR.

cowofevil:
@zreichert The ticket got out of sync with the actual work that needed to be done. I've updated the ticket to reflect the actual work that needed to be done.

PR #334, opened 2015-03-06
[cli/version] (RK-37) Add --verbose flag to r10k version

elyscape:
This adds support for the `-v`/`--verbose` flag to `r10k version`:

Without the flag:

``` shellsession
[root@puppet cli]# r10k version
r10k 1.4.1
```

Note that this now prints `r10k 1.4.1` instead of `1.4.1`.  This is because it was simpler to always print `r10k 1.4.1` than to print `1.4.1` normally and `r10k 1.4.1` if passed `-v`/`--verbose`.

With the flag on a 1.8.7 box:

``` shellsession
[root@puppet-eli cli]# r10k version --verbose
r10k 1.4.1
ruby 1.8.7 (2013-06-27 patchlevel 374) [x86_64-linux]
Command path: /usr/bin/r10k
Interpreter path: /usr/bin/ruby
```

With the flag on a 1.9.3 box:

``` shellsession
[root@puppet cli]# r10k version -v
r10k 1.4.1
ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]
Command path: /usr/bin/r10k
Interpreter path: /opt/puppet/bin/ruby
Default encoding: UTF-8
```

This change is documented in [RK-37](https://tickets.puppetlabs.com/browse/RK-37).

### DIFF
--- a/lib/r10k/cli/version.rb
+++ b/lib/r10k/cli/version.rb
@@ -12,7 +12,16 @@ module R10K::CLI
         summary 'Print the version of r10k'
 
         run do |opts, args, cmd|
-          puts R10K::VERSION
+          puts "r10k #{R10K::VERSION}"
+          if opts[:verbose]
+            puts RUBY_DESCRIPTION
+            cmdpath = caller.last.slice(/\A.*#{$PROGRAM_NAME}/)
+            puts "Command path: #{cmdpath}"
+            puts "Interpreter path: #{Gem.ruby}"
+            if RUBY_VERSION >= '1.9'
+              puts "Default encoding: #{Encoding.default_external.name}"
+            end
+          end
           exit 0
         end
       end


lib/r10k/cli/version.rb
```diff
@@ -12,7 +12,16 @@ def self.command
...
 
         run do |opts, args, cmd|
-          puts R10K::VERSION
+          puts "r10k #{R10K::VERSION}"
+          if opts[:verbose]
+            puts RUBY_DESCRIPTION
+            cmdpath = caller.last.match(/\A(.*#{$PROGRAM_NAME}):/).captures.first
+            puts "Command path: #{cmdpath}"
+            puts "Interpreter path: #{Gem.ruby}"
+            if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('1.9.3')
```
adrienthebo: If we're running on Ruby 1.8.7 we don't have a guarantee that rubygems was loaded; I would prefer that we do a naive string compare (`RUBY_VERSION >= '1.9.3'`) and skip the `Gem::Version`.

elyscape: Change made.

lib/r10k/cli/version.rb
```diff
@@ -12,7 +12,16 @@ def self.command
...
         run do |opts, args, cmd|
-          puts R10K::VERSION
+          puts "r10k #{R10K::VERSION}"
+          if opts[:verbose]
+            puts RUBY_DESCRIPTION
+            cmdpath = caller.last.match(/\A(.*#{$PROGRAM_NAME}):/).captures.first
+            puts "Command path: #{cmdpath}"
+            puts "Interpreter path: #{Gem.ruby}"
+            if RUBY_VERSION >= '1.9.3'
+              puts "Default encoding: #{Encoding.default_external.name}"
```
adrienthebo, line 22: Do we need to explicitly call `default_external.name`? Will `default_external` work?

lib/r10k/cli/version.rb
```diff
@@ -12,7 +12,16 @@ def self.command
         summary 'Print the version of r10k'
 
         run do |opts, args, cmd|
-          puts R10K::VERSION
+          puts "r10k #{R10K::VERSION}"
+          if opts[:verbose]
+            puts RUBY_DESCRIPTION
+            cmdpath = caller.last.match(/\A(.*#{$PROGRAM_NAME}):/).captures.first
```
adrienthebo: This packs a lot of logic into one line, could we break it up a bit?

adrienthebo: This could be `caller.last.slice(/\A.*#{$PROGRAM_NAME}/)` since we just want to chop out a substring.

adrienthebo:
:+1: on this in general; could you create an issue at https://tickets.puppetlabs.com/browse/RK so we can track this with the rest of the work in progress?

lib/r10k/cli/version.rb
elyscape: Good call.

lib/r10k/cli/version.rb
elyscape, line 22: We could, but it's uglier:

``` shellsession
irb(main):001:0> Encoding.default_external
=> #<Encoding:UTF-8>
```

adrienthebo, line 22: That's the output of `Encoding#inspect`; `Encoding#to_s` prints a much prettier version:

```
[4] pry(main)> "#{Encoding.default_external}"
=> "UTF-8"
[5] pry(main)> $ Encoding.default_external.to_s

From: encoding.c (C Method):
Owner: Encoding
Visibility: public
Number of lines: 5

static VALUE
enc_name(VALUE self)
{
    return rb_usascii_str_new2(rb_enc_name((rb_encoding*)DATA_PTR(self)));
}
```

I'm debating whether the more terse form is more or less clear.

elyscape:
[RK-37 created](https://tickets.puppetlabs.com/browse/RK-37).

adrienthebo:
Merged into master in 5017bc5; this should be released in 1.5.0. Thanks for the contribution!

PR #335, opened 2015-03-07
[deployment] (RK-4) Raise meaningful error on missing sources

adrienthebo:
If the 'sources' key in r10k.yaml was left unset, was misspelled, or was
empty, r10k try to blindly iterate through it as a hash and would
subsequently raise an error. This commit updates the source enumeration
code to check for these error case and raise a meaningful error.

Per the comment on RK-4 there are other issues with config loading but they're out of scope for this issue; the broader issue is tracked in RK-34.

### DIFF
--- a/lib/r10k/deployment.rb
+++ b/lib/r10k/deployment.rb
@@ -114,6 +114,9 @@ module R10K
 
     def load_sources
       sources = @config.setting(:sources)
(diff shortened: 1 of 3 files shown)


adrienthebo:
Retargeting at 1.4.x.

PR #336, opened 2015-03-07
[deployment] (RK-4) Raise meaningful error on missing sources

adrienthebo:
If the 'sources' key in r10k.yaml was left unset, was misspelled, or was
empty, r10k try to blindly iterate through it as a hash and would
subsequently raise an error. This commit updates the source enumeration
code to check for these error case and raise a meaningful error.


andersonmills:
:+1: Tested and approved.

PR #337, opened 2015-03-07
[deploy/display] (RK-36) Monkey patch Symbol#<=> on Ruby 1.8.7

adrienthebo:
Ruby 1.8.7 didn't implement Symbol#<=>, which broke sort operations on
arrays of symbols. This commit backports the 1.9 symbol comparison method
to remove the inconsistency.

Thanks to Eli Young elyscape@gmail.com for suggesting the monkey patch
instead of inserting 1.8.7 specific hacks into random methods.

### DIFF
--- a/lib/r10k/action/deploy/display.rb
+++ b/lib/r10k/action/deploy/display.rb
@@ -1,3 +1,4 @@
+require 'r10k/util/monkey_patches'
 require 'r10k/util/setopts'
 require 'r10k/deployment'
 require 'r10k/logging'
--- a/lib/r10k/util/monkey_patches.rb
+++ b/lib/r10k/util/monkey_patches.rb
@@ -1,0 +1,11 @@
+if !Symbol.instance_methods.include?(:<=>)
+  class Symbol
+    # Ruby 1.8.7 does not define #<=>, which subsequently breaks Enumerable#sort
+    # when sorting an array of symbols.
+    #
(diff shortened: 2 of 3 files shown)


elyscape:
Why not add an implementation for `Symbol#<=>` when it isn't defined?

``` ruby
unless :sources.respond_to?(:<=>)
  class Symbol
    def <=> other
      return unless Symbol === other
      return self.to_s <=> other.to_s
    end
  end
end
```

Alternative implementation that can be called inside other modules:

``` ruby
unless :sources.respond_to?(:<=>)
  Symbol.class_eval do
    def <=> other
      return unless Symbol === other
      return self.to_s <=> other.to_s
    end
  end
end
```

adrienthebo:
Because I hate monkey patching with a great fiery passion. I could, but I also want to drop 1.8.7 support sooner rather than later.

elyscape:
Makes sense.  Carry on!

adrienthebo:
On the other hand, monkey patching wouldn't stuff version specific logic all over the place; I think you may be onto something. I'll revisit it later.

elyscape:
Yeah, that was my reasoning.  You could put all the Ruby 1.8 fixup monkeypatching in `make_ruby_18_suck_less.rb` or something and then just remove it when you stop supporting it.

andersonmills:
:+1: I believe in a thing called love!

adrienthebo:
Just listen to the rhythm of my heart!

PR #339, opened 2015-03-09
[git] (RK-32) Use Rugged as the fallback Git provider

adrienthebo:
R10K needs to be able to use the Rugged provider as a fallback in case
the Git executable is not present on the system. This commit updates the
`R10K::Git.default` method to probe for multiple Git providers rather
than hardcode for the presents of shellgit.

### DIFF
--- a/doc/git/providers.mkd
+++ b/doc/git/providers.mkd
@@ -1,0 +1,53 @@
+Git Providers
+=============
+
+As of 1.5.0, r10k can interact with Git repositories using multiple Git
+providers.
+
+Shellgit
+--------
+
+The shellgit provider is the original Git provider that is based on shelling out
+to the `git` binary. It relies on the standard set of Git userland executables
+in order to work.
+
+The shellgit provider is the default Git provider in order to maintain
+compatibility with existing r10k installations.
+
+### Requirements
+
+The shellgit provider requires that `git` can be found on the `PATH` environment
+variable. This can be done by installing the git package via the system package
+manager.
+
+### SSH Configuration
+
+Because the shellgit provider relies on the `git` command which in turn uses the
+`ssh` binary as the SSH transport layer, configuring access to Git repositories
+over SSH is done by configuring the underlying `ssh` command.
+
+Rugged
+------
+
+The rugged provider is based on the [libgit2](https://github.com/libgit2/libgit2)
+library and the Ruby [rugged gem](https://github.com/libgit2/rugged).
+
+### Requirements
+
+The rugged provider relies on the rugged Ruby gem. The rugged gem requires Ruby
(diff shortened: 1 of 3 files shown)


andersonmills:
I believe in this code and give the :+1:. But I want to make sure that I understand correctly. This code is currently not being called at all when r10k runs, correct? It seems that it is only being called by the spec tests.

adrienthebo:
It will be used; any code touching the Git implementation invokes one of the [indirected methods](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/git.rb#L22-L32) to get the correct implementation, which is based on the [current provider](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/git.rb#L18-L20). That value is currently populated by the [`self.default` method](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/git.rb#L10-L16), which this code updates to include Rugged as a possible implementation. If the Git executable is, well, not executable, then Rugged will be used as the next best default value, per the [config docs](https://github.com/puppetlabs/r10k/pull/339/files#diff-0d37eef1cf89ef37336b6ec00756dc20R46).

andersonmills:
I see now that it is being called.

Though I will point out that if git is not executable, then I get

```
$ be r10k deploy environment -vp
There was  Errno::ENOENT while loading r10k.gemspec:
No such file or directory - git ls-files from
  /Users/anderson/puppet/src/r10k/r10k.gemspec:41:in ``'
```

adrienthebo:
Yep; that's a package time issue that we can ignore since when r10k is installed we won't be touching the gemspec, but for my demo I did this:

``` diff
diff --git a/r10k.gemspec b/r10k.gemspec
index 2342a8b..444de5d 100644
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -38,7 +38,11 @@ Gem::Specification.new do |s|

   s.add_development_dependency 'yard', '~> 0.8.7.3'

-  s.files        = %x[git ls-files].split($/)
+  # s.files        = %x[git ls-files].split($/)
+  root = File.expand_path('..' , __FILE__)
+  puts root
+
+  s.files = Dir.glob(File.join(root, 'bin', '*')) + Dir.glob(File.join(root, 'lib', '**', '*'))
   s.require_path = 'lib'
   s.bindir       = 'bin'
   s.executables  = 'r10k'
```

andersonmills:
Okay. After in-person discussion, I see now that Feature#library_available? isn't called if the :libraries key doesn't exist for a Feature, which is where I got the idea that code wasn't being called. Woof!

Also, I see that the above error for missing git is because it doesn't exist at packaging rather than during execution.

:+1: again.

PR #343, opened 2015-03-10
(RK-24) Add the AIO location for r10k.yaml

andersonmills:
Before this commit, the AIO location for r10k.yaml was not searched for
a configuration file. This commit adds /etc/puppetlabs/r10k/r10k.yaml to
the search path, and adds spec tests for the two expected paths to
r10k.yaml

### DIFF
--- a/spec/unit/deployment/config/loader_spec.rb
+++ b/spec/unit/deployment/config/loader_spec.rb
@@ -1,0 +1,18 @@
+require 'spec_helper'
+require 'r10k/deployment/config/loader'
+
+describe R10K::Deployment::Config::Loader do
+
+  it "includes /etc/puppetlabs/r10k/r10k.yaml in the loadpath" do
+    expect(subject.loadpath).to include("/etc/puppetlabs/r10k/r10k.yaml")
+  end
+
+  it "includes /etc/r10k.yaml in the loadpath" do
+    expect(subject.loadpath).to include("/etc/r10k.yaml")
+  end
+
+  it "does not include /some/random/path/atomium/r10k.yaml in the loadpath" do
+    expect(subject.loadpath).not_to include("/some/random/path/atomium/r10k.yaml")
+  end
+
+end
--- a/lib/r10k/deployment/config/loader.rb
+++ b/lib/r10k/deployment/config/loader.rb
@@ -1,43 +1,49 @@
 
 module R10K
-class Deployment
-class Config
-class Loader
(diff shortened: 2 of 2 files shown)


spec/unit/deployment/config/loader_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+describe R10K::Deployment::Config::Loader do
+
+  it "creates R10K::Deployment::Config::Loader instances" do
+    subject = described_class.new()
+    expect(subject).to be_a_kind_of R10K::Deployment::Config::Loader 
+  end
+
+  it "includes /etc/puppetlabs/r10k/r10k.yaml in the loadpath" do
+    subject = described_class.new()
```
adrienthebo: RSpec automatically creates a `subject` helper by calling `described_class.new`, so we have this implicit helper in every spec:

``` ruby
  let(:subject) do # magic!
    described_class.new
  end
```

Unless we need to override the constructor for a specific test, we don't need to explicitly set subject to use it.

For future reference, you can also set an explicit subject:

``` ruby
  subject do
    described_class.new('something', 'with', specific', 'args')
  end
```

spec/unit/deployment/config/loader_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
+require 'r10k/deployment/config/loader'
+
+describe R10K::Deployment::Config::Loader do
+
+  it "creates R10K::Deployment::Config::Loader instances" do
```
adrienthebo: In some cases we may need to test the object type returned by `.new`, specifically when we're overloading new to [redirect a deprecated class](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/deployment/source.rb#L33). But unless that's happening, we can generally assume that `.new` is behaving as expected and skip testing it. If we're not getting back a `R10K::Deployment::Config::Loader` object, either all the tests pass, the desired behavior is still occurring, and there's not a problem, or the tests crash and burn, in which case this test would fail as well. If this covers behavior that wouldn't be caught otherwise then we should keep it; if another test will fail if another test fails then we can omit it.

andersonmills: No problem removing that. It was the first "are these tests working" test. I didn't see a problem with leaving it in, but happy to remove it.

adrienthebo:
♪ The wheels on the Travis-CI go round and round, round and round, round and round! ♪

PR #344, opened 2015-03-10
(RK-25) Warn if old and new default configs exist

andersonmills:
Before this commit, there was no warning when both the old ond new default locations for the configuration file existed. With this commit, if /etc/r10k.yaml and /etc/puppetlabs/r10k/r10k.yaml, a warning will be issued.

NOTE! This PR depends on https://github.com/puppetlabs/r10k/pull/343


adrienthebo:
It doesn't look like GitHub has recognized that this is based on #343; this might need a rebase tomorrow morning.

PR #350, opened 2015-03-13
(RK-50) Fix Broken Integration Tests

cowofevil:
Two tests were failing and now have been fixed.


cowofevil:
Since I can't know what the error message will look like when the bug is fixed I have to throw a bogus one in that will cause the "expect_failure" statement to detect a failure. The limitations of "expect_failure" keep me from doing something more intelligent.

carlosvelarde:
ah, got it. Both tests passing, merged

PR #351, opened 2015-03-13
(maint) Add static `.symbolize_keys!` method, remove use of symbolize keys mixin

adrienthebo:
The existing #symbolize_keys! method needed callers to monkey patch themselves in order to use the method, which was less than ideal. doing monkey patching like this at runtime invalidates the Ruby internal method cache which can cause a non-trivial performance hit, and the monkey patching was done selectively in the first place.

This pull request defines a standalone `symbolize_keys!` method that can be invoked without monkey patching, removes the old monkey patched method, and removes existing invocations of the monkey patching approach with calls to the static method.


andersonmills:
:+1:

I believe in the code, and that the tests are testing what they need to.

PR #352, opened 2015-03-13
(CODEMGMT-104) Update Tests to Use Acceptance Forge

cowofevil:
Currently we run our tests against production Forge which could present problems
with test stability. All Beaker configs have been updated to reference the
acceptance forge. All tests that use modules from the Forge have been updated
to use the "stub_forge_on" function.

### DIFF
--- a/integration/configs/pe/centos-6-64mda
+++ b/integration/configs/pe/centos-6-64mda
@@ -22,3 +22,4 @@ CONFIG:
   folder: Delivery/Quality Assurance/Enterprise/Dynamic
   resourcepool: delivery/Quality Assurance/Enterprise/Dynamic
   pooling_api: http://vmpooler.delivery.puppetlabs.net/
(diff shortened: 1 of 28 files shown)


cowofevil:
@carlosvelarde I found a lot of errors in the tests that were causing a variety of failures. I fixed them all up and this PR is ready for review again.

cowofevil:
@carlosvelarde I have updated the PR again to finally fix all the brokeness! You will need to force pull this branch because I did a rebase on master. The reason things started breaking was because of changes in the "stdlib" Forge module. I also needed to prevent a test from running on anything but the EL platforms.

cowofevil:
@carlosvelarde Okay, I ran all the tests against all platforms and they passed. Please review again.

carlosvelarde:
Only test that fails is neg_duplicate_module_names.rb with a network failure to https://forgeapi.puppetlabs.com. Passes when I run it solo.

PR #354, opened 2015-03-16
(RK-33) User configurable git provider

adrienthebo:
This pull request allows users to specify which Git provider they want to use. Configuration is handled in r10k.yaml with the following:

``` yaml
git:
  provider: rugged # valid values are rugged, shellgit
[...]
```

This pull request is based on GH-339 and GH-351.

### DIFF
--- a/spec/unit/deployment/config_spec.rb
+++ b/spec/unit/deployment/config_spec.rb
@@ -1,0 +1,19 @@
+require 'spec_helper'
+
+describe R10K::Deployment::Config do
+
+  describe "reading settings" do
+    # Try all permutations of string/symbol values as arguments to #setting and
+    # values in the config file
+    x = [:key, "key"]
+    matrix = x.product(x)
+
+    matrix.each do |(searchvalue, configvalue)|
+      it "treats #{searchvalue.inspect}:#{searchvalue.class} and #{configvalue.inspect}:#{configvalue.class} as equivalent" do
+        expect(YAML).to receive(:load_file).with('foo/bar').and_return(configvalue => 'some/cache')
+        subject = described_class.new('foo/bar')
+        expect(subject.setting(searchvalue)).to eq 'some/cache'
+      end
+    end
+  end
+end
--- a/spec/unit/deployment/config_spec.rb
+++ b/spec/unit/deployment/config_spec.rb
@@ -16,4 +16,22 @@ describe R10K::Deployment::Config do
       end
     end
   end
+
+  describe "applying global settings" do
+    describe "for the cachedir" do
+      it "sets the git cache root when given" do
+        expect(YAML).to receive(:load_file).with('foo/bar').and_return('cachedir' => 'some/cache')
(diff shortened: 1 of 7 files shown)


andersonmills:
Does this need a rebase?

adrienthebo:
Oui, un moment sil vous plait

andersonmills:
Tu ne dois pas vous-er a moi. :)

adrienthebo:
I've made the additional changes that I needed and rebase, this is ready for review/heckling.

spec/unit/deployment/config_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    matrix = x.product(x)
+
+    matrix.each do |(searchvalue, configvalue)|
+      it "treats #{searchvalue.inspect}:#{searchvalue.class} and #{configvalue.inspect}:#{configvalue.class} as equivalent" do
+        expect(YAML).to receive(:load_file).with('foo/bar').and_return(configvalue => 'some/cache')
+        subject = described_class.new('foo/bar')
+        expect(subject.setting(searchvalue)).to eq 'some/cache'
+      end
+    end
+  end
```
andersonmills, line 18: It took me a bit to figure out what this test was doing, but I do believe it is testing what is intended and needs to be tested.

adrienthebo, line 18: Makes sure all permutations of string and symbol keys are equivalent. As part of https://tickets.puppetlabs.com/browse/RK-59 this should be rendered obsolete, so it's there to get test coverage right here, right now, but I would rather not have to do this at all. Kludgy tests for kludgy code.

spec/unit/git_spec.rb
```diff
@@ -21,5 +24,54 @@
...
+      expect(described_class).to receive(:default).and_return :def
+      expect(described_class.provider).to eq :def
+    end
+
+    it "uses an explicitly set provider" do
+      expect(R10K::Features).to receive(:available?).with(:rugged).and_return true
+      described_class.provider = :rugged
+      expect(described_class).to_not receive(:default)
+      expect(described_class.provider).to eq R10K::Git::Rugged
+    end
```
andersonmills, line 72: So, this test is exactly the same as "sets the current provider if the provider exists and is functional" above with the exception of the described_class receiving default. Is it supposed to receive that call or not?

adrienthebo, line 72: Unfortunately testing `.provider=` and `.provider` is tricky because to test one you have to use the other, so there is duplication. In the earlier test I suspect that the `.default` method isn't even being called; I'll check. In the case that it doesn't exist they are entirely identical; I could delete one to resolve the duplication.

adrienthebo, line 72: Yep, that stub isn't used; I'll remove it. I feel weird because it feels like two different cases but they overlap a lot.

andersonmills, line 72: I would say that just removing the stub above does make the purpose of these two different but similar looking tests more explicit.

andersonmills:
:+1: Just waiting on Travis.

PR #356, opened 2015-03-16
[doc] Document config file searching for 1.5.0

adrienthebo:
See also RK-24, RK-25, and RK-31.

/cc @andersonmills @laurenrother


andersonmills:
Late to the party, but :+1:

adrienthebo:
I just got so excited!

PR #359, opened 2015-03-18
[source] Extract environment name class

adrienthebo:
The Git and SVN sources each had separate implementations of the Environment name helper with a fair amount of overlap; this commit removes the separate implementations in favor of a common Environment::Name helper.


andersonmills:
:+1: I think this is good, and I'm happy with it.

Issue #106, opened 2014-03-27
Implement the 'forge' Puppetfile configuration directive


thias:
The current README mentions it's a noop. In situations where there is no Internet access whatsoever, it's really useful to be able to use a local forge.

(note that I've searched existing issues, but found none to track this)

thias:
Just a quick note : I saw that the protocol isn't included in the current example, and that https is hardcoded in parts of the code. I think it'll be required to move the protocol to be part of the "forge" value from the Puppetfile, in order to be able to use plain http.

Looking at librarian-puppet, it seems like the protocol is part of the setting, so it would be the same. They have `forge "http://forge.puppetlabs.com"` in their example file.

walidshaari:
In our workplace, we too have the situation of disconnected and came to this issue in hope to see what possible options are in R10k for local modules, or forge. or  what the forge directive can point to locally if not the official puppet forge. (e.g.  local repo served by http, nfs from previously downloaded modules.)

scotje:
I'm going to go ahead and close this issue. There is now a Forge `baseurl` setting in r10k.yaml (https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#baseurl) which I think makes a lot more sense than controlling this on a per-environment basis in the Puppetfile.

`forge` directives in the Puppetfile will likely be deprecated in a future version of r10k, not that they ever did anything. :)

PR #362, opened 2015-03-19
(#106) use forge directive so r10k works with using local forges set in ...

rjsab:
I needed the ability to utilize a local forge. This patch attempts to add support for the forge directive, which should allow r10k users to use internal forges by setting them in the Puppetfile. If no forge is set, it will still default to https://forgeapi.puppetlabs.com.

### DIFF
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -24,6 +24,10 @@ class R10K::Module::Forge < R10K::Module::Base
 
   include R10K::Logging
 
+  def self.forge(forge)
+    @@forge = forge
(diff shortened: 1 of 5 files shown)


adrienthebo:
This change has broken a large number of specs, especially because this makes backwards incompatible API changes. Could you see if these changes could be made while preserving the current API?

rjsab:
I reset back to master to pick up some of the latest updates you've pushed and made the changes. I think this one preservers the current API. If not, I'm happy to continue working toward a solution. I also updated a couple of the specs in order to account for the new default forges.

PR #371, opened 2015-03-29
(RK-39) Add logging for rugged provider, general logging clarification

adrienthebo:
This pull request adds logging for actions taken by the rugged provider and clarifies a number of existing log messages.


adrienthebo:
This adds quite a bit more logging which makes it run into the issue fixed by #372; that pull request should be merged before this one.

andersonmills:
:+1: Tested it and believe in it. Travis was happy with the last version of this, and the last commit was minor.

PR #372, opened 2015-03-29
[logging] Use log4r logger repository

adrienthebo:
The intention of the `#logger` method was to provide a single logger per
class to add context for the log messages, but reuse a single instance
per class to cut down on object overhead. However this does not happen
automatically, and the existing code was creating many more logger
instances than what was intended, which had a nontrivial overhead.

When a log4r logger is created, it's automatically stored by log4r in an
internal repository for later use. The logger has to be explicitly
created as requesting a logger with a given name will return nil if that
logger hasn't been created, but can be retrieved afterwards. This commit
modifies the `#logger` method to try to retrieve the class logger from
the log4r repository first, and if that fails generate a new logger.

### DIFF
--- a/lib/r10k/logging.rb
+++ b/lib/r10k/logging.rb
@@ -14,9 +14,14 @@ module R10K::Logging
   end
 
   def logger
-    unless @logger
-      @logger = Log4r::Logger.new(self.logger_name)
-      @logger.add R10K::Logging.outputter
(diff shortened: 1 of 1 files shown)


andersonmills:
:+1:  I've tested this and believe in it. I hear that spec tests are in later PRs.

PR #373, opened 2015-03-29
(refactor) Clean up, deduplicate, document, test log configuration

adrienthebo:
This is based on #372.

PR #374, opened 2015-03-29
[logging] Use more appropriate log formatters at different log levels

adrienthebo:
At lower log levels it makes sense to keep log chatter at an appropriate
minimum, while at higher levels log messages benefit from having more
context. This commit adds a debug level formatter which adds more
information to log messages and enables it when the log level is DEBUG
or higher.

Based on #373.


andersonmills:
I BELIEVE! :+1:

adrienthebo:
\o/

PR #375, opened 2015-03-29
[logging] Centralize "default" log level handling

adrienthebo:
If a user specifies '--verbose' without a level, Cri will set the
verbosity as "true:TrueClass" and pass this along. This was originally
handled in the CLI code but it was far removed from logging, and if the
log level was invalid an exception would be thrown during argument
parsing.

This commit adds handling of 'true' as a log level and removes the
extra code path.

This is based on #373.

### DIFF
--- a/lib/r10k/action/cri_runner.rb
+++ b/lib/r10k/action/cri_runner.rb
@@ -41,21 +41,10 @@ module R10K
 
       # @return [Hash] The adapted options for the runner
       def handle_opts(opts)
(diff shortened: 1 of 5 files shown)


andersonmills:
This is funny...

```
$ be r10k deploy environment -pv true

Error while running: #<ArgumentError: Log level must be in 0..10>
```

Happens for `-v false` also.

adrienthebo:
Gr. Fixed.

andersonmills:
Behavior looks correct now. :+1:

PR #382, opened 2015-04-01
(RK-39) Wrap and raise rugged errors

adrienthebo:
Without this PR, if a clone or fetch fails in the rugged provider, the raised error will not report which repository had the issue, making it very hard to debug the failure. This PR adds exception handling to methods invoking network operations to add debugging context to the message.


andersonmills:
:+1: 

I believe it can fly.

PR #383, opened 2015-04-01
(spec) reset log level to default level instead of "off"

adrienthebo:


andersonmills:
Love it. SO glad those errors are gone. :)

:+1:

PR #384, opened 2015-04-01
Add colored log levels, --color command line argument

adrienthebo:
This pull request adds a log4r outputter that adds color for different log levels, in order to help break up the visual output when high volumes of logs are put out and draw out warnings and errors from non-essential messages. Color output is disabled by default and can be enabled by passing the `--color` flag to r10k.


andersonmills:
:+1: Happiness.

adrienthebo:
\o/

PR #388, opened 2015-04-02
[rugged/bare_repository] Wipe/restore branches on fetch

adrienthebo:
Branch pruning on fetch was implemented in libgit2 0.22.0, but the only
version of rugged currently available that uses libgit2 0.22.x is
rugged 0.22.1b1, which is a prerelease. To get around this limitation
all branches are wiped before a fetch, with the expectation that all
branches will be restored by the fetch operation. If an error occurs the
old branches are restored.


andersonmills:
:+1: I believe in this and tested it. The acceptance tests cover this situation.

PR #389, opened 2015-04-02
(RK-48) Ignore deleted releases when choosing latest

adreyer:
Previously r10k would try to install a deleted latest releases. This
would result in r10k asking PMT to install a release that it will error
out on. Now r10k will choose the latest non-deleted release which should
successfully install.(The cassettes will no longer rebuild cleanly)


adrienthebo:
The cassettes won't rebuild because we used the acceptance forge?

adreyer:
The cassettes won't rebuild because different tests use different forges. See RK-77

andersonmills:
:+1:  I'm not nervous; I'm ecstatic!!!

PR #396, opened 2015-04-06
[feature] (RK-73) Add logging for feature detection

adrienthebo:
Right now determining if a feature is available is entirely opaque; if a
feature failed to load there's no indication why.  For instance, the
rugged provider tries to load the 'rugged' gem - if that fails for any
reason (such as a missing shared object) the error is masked and r10k
will pretend that the rugged library is not present.

This commit adds log messages so that if a feature is unavailable the
failure can be divined by checking the debug level logging.

### DIFF
--- a/lib/r10k/feature.rb
+++ b/lib/r10k/feature.rb
@@ -1,6 +1,10 @@
+require 'r10k/logging'
+
 module R10K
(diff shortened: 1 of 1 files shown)


andersonmills:
:+1: Don't go breakin' my heart.

PR #398, opened 2015-04-08
[git] (RK-62) Warn if rugged is missing features

adrienthebo:
If Rugged was compiled without SSH support and the SSH transport is
used, Rugged will throw a rather cryptic error. This commit adds an
`:on_set` hook for Git providers, uses that hook to check for the
features compiled into Rugged, and emits a warning if the https or ssh
feature is missing.


andersonmills:
:+1: Happy!

PR #399, opened 2015-04-09
[git/stateful_repository] (RK-86) Correctly sync branches

adrienthebo:
R10k 1.5.0 added an optimization for avoiding syncing the Git cache
unless necessary, but accidentally treated all refs as commits, causing
branches to not be updated. This only affected Git modules as
environments/sources would always force a sync. This commit resolves the
issue by checking the type of the ref, not the type of the resolved sha,
when determining if the cache needs to be updated.


acidprime:
:+1: tested on a machine that was previously not working and `r10k deploy module foo` now works.

PR #404, opened 2015-04-13
(CODEMGMT-167) Exploratory Testing Helper Scripts

cowofevil:
Added several helper scripts for setting up an environment in the "vmpooler"
for "r10k" exploratory testing. A README has been added to document how to
use the scripts. Also, the Beaker configs have been updated to make it
clear in Beaker output which machine is the Puppet master.

### DIFF
--- a/integration/scripts/README.mkd
+++ b/integration/scripts/README.mkd
@@ -1,0 +1,83 @@
+Scripts
+=======
+
+This folder contains helper scripts for setting up manual testing environments
+using [Beaker](https://github.com/puppetlabs/beaker) configuration files. The
+environments created by these scripts have "r10k" fully configured along with
+a Git repository with a valid "production" environment.
+
+## Prerequisites
+
+To utilize this test scripts you will need to have Ruby 1.9.3 or greater
+installed on your system along with Bundler. Also, the scripts utilize the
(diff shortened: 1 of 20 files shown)


integration/scripts/README.mkd
```diff
@@ -0,0 +1,83 @@
...
+platform and execute the script:
+
+```bash
+bash setup_r10k_env_centos6.sh
+```
+
+## Connecting to Machines
+
+The setup process takes about 10 minutes to complete. Once finished Beaker
+will report that all tests have been ran successfully. The output log will
```
andersonmills: have been _run_, I believe.

integration/scripts/README.mkd
```diff
@@ -0,0 +1,83 @@
+Scripts
+=======
+
```
andersonmills, line 3: Since this in an open source project, I think it'd be worth mentioning here that these are to be used while on Puppet's internal network. If we specifically don't do this on purpose or as policy, that's fine.

andersonmills:
Other than those two comments on the README.mkd, I'm :+1: on this.

cowofevil:
@andersonmills Updated README, please review again.

andersonmills:
Perfecto!

PR #408, opened 2015-04-14
Update/document r10k.yaml.example with new settings

adrienthebo:
This updates the r10k.yaml.example config to include new configuration
settings and adds inline documentation explaining each setting.

### DIFF
--- a/r10k.yaml.example
+++ b/r10k.yaml.example
@@ -1,5 +1,86 @@
-:cachedir: '/var/cache/r10k'
-:sources:
-  :plops:
-    remote: 'git@github.com:my-org/org-shared-modules'
-    basedir: '/etc/puppet/environments'
+---
+# This example configuration details the most commonly used configuration
+# options for the `r10k deploy` command.
+#
+
+# The 'cachedir' setting controls where cached content, such as mirrored Git
+# repositories, are stored on the local machine. This location should be
+# persistent, as environments and modules may rely on these files in order to
+# be updated.
+#
+# See https://github.com/puppetlabs/r10k/blob/1.5.1/doc/git/cloning-and-mirroring.mkd
+# for more information on how r10k caches Git repositories.
+#
+# The default value is "~/.r10k"
+cachedir: '/var/cache/r10k'
+
+# The 'sources' setting determines what repositories r10k will use when creating
+# Puppet environments.
+sources:
+  # Each source should have a unique name, and preferrably should only use
+  # alphanumeric characters and underscores.
+  operations:
+
+    # The 'remote' setting determines the location of the VCS repository used
+    # for creating environments.
+    remote: 'git@github.com:my-org/org-operations-modules'
+
+    # The 'basedir' setting specifies where environments will be created for
+    # this source. This setting should match the Puppet 'environmentpath'
+    # setting in puppet.conf.
+    #
+    # See http://docs.puppetlabs.com/references/latest/configuration.html#environmentpath
+    # for more information about the Puppet 'environmentpath' setting.
+    basedir: '/etc/puppetlabs/puppet/environments'
+
+  # One or more sources can be specified; each source is simple another entry
+  # in the sources map.
+  qa:
+    remote: 'git@github.com:my-org/org-qa-modules'
+
+    # Multiple sources can have the same 'basedir' setting, in case there is a
+    # reason to push different sets of environments to the same Puppet master.
+    # If two different sources have the same basedir care should be exercised
+    # to make sure that two sources don't try create multiple environments with
+    # the same name. See the 'prefix' setting for more information.
+    basedir: '/etc/puppetlabs/puppet/environments'
+
+    # Sources can use the 'prefix' setting to avoid collisions between multiple
+    # sources being created in the same directory. Setting prefix to true
+    # prepends the source name and an underscore to each environment, so for
+    # the qa source the "production" branch will create the "qa_production"
+    # branch.
+    #
+    # The default value is 'false'.
+    #
+    # See https://github.com/puppetlabs/r10k/blob/1.5.1/doc/dynamic-environments/configuration.mkd#prefix
+    # for more information about the prefix setting.
+    prefix: true
+
+# Additional configuration can be supplied to configure how r10k uses Git
+# and what version of Git it uses.
+#
+# See https://github.com/puppetlabs/r10k/tree/1.5.1/doc/git for more
+# information on how r10k can configure Git.
+git:
+
+  # As of 1.5.0 r10k can interact with Git repositories in two ways - by
+  # shelling out to the 'git' executable, and by using libgit2 through the
+  # 'rugged' library.
+  #
+  # See https://github.com/puppetlabs/r10k/blob/1.5.1/doc/git/providers.mkd
+  # for more information on the Git providers.
+  provider: 'shellgit' # Either 'shellgit' or 'rugged', defaults to 'shellgit'
+
+  # The 'private_key' setting sets the the SSH private key to use for remote
+  # Git repositories accessed over SSH.
+  #
+  # The 'private_key' setting is only used by the 'rugged' Git provider.
+  #private_key: '/root/.ssh/id_rsa'
+
+  # The 'username' setting sets the SSH username for remote Git repositories
+  # when a username is not provided in the URL. The username defaults to the
+  # current user.
+  #
+  # The 'username' setting is only used by the 'rugged' Git provider.
+  #username: 'git'


r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
+
+    # The 'remote' setting determines the location of the VCS repository used
+    # for creating environments.
+    remote: 'git@github.com:my-org/org-operations-modules'
+
+    # The 'basedir' setting specifies where environments will be created for
+    # this source. This setting should match the Puppet 'environmentpath'
+    # setting in puppet.conf.
+    #
+    # See http://docs.puppetlabs.com/references/3.7.latest/configuration.html#environmentpath
```
cowofevil: This link points to 3.7, you may want to change it to 3.8.

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
+    # information.
+    basedir: '/etc/puppetlabs/puppet/environments'
+
+    # To avoid collisions between multiple sources being created in the same
+    # directory sources can use the prefix setting. Setting prefix to true
+    # prepends the source name and an underscore to each environment, so for
+    # the qa source the "production" branch will create the "qa_production"
+    # branch.
+    #
+    # See https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#prefix
```
cowofevil: Are you sure you want to point to master? What if the documentation changes in the FOSS version, but a PE 3.8 user is reading this example file?

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
+# and what version of Git it uses.
+#
+# See https://github.com/puppetlabs/r10k/tree/master/doc/git for more
+# information on how r10k can configure Git.
+git:
+
+  # As of 1.5.0 r10k can interact with Git repositories in two ways - by
+  # shelling out to the 'git' executable, and by using libgit2.
+  #
+  # See https://github.com/puppetlabs/r10k/blob/master/doc/git/providers.mkd
```
cowofevil: Ditto from above comment.

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
-  :plops:
-    remote: 'git@github.com:my-org/org-shared-modules'
-    basedir: '/etc/puppet/environments'
+---
+
+# The cachedir setting controls where cached content, such as mirrored Git
+# repositories, are stored. This location should be persistent, as environments
+# and modules may rely on these files in order to be updated.
+#
+# See https://github.com/puppetlabs/r10k/blob/master/doc/git/cloning-and-mirroring.mkd
```
cowofevil: Ditto from above comment.

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
-:cachedir: '/var/cache/r10k'
-:sources:
-  :plops:
-    remote: 'git@github.com:my-org/org-shared-modules'
-    basedir: '/etc/puppet/environments'
+---
+
```
cowofevil, line 5: I suggest adding a link at the top pointing to the quickstart guide for r10k.

r10k.yaml.example
adrienthebo: This is for Puppet itself, not all of PE - are we shipping a Puppet 3.8?

r10k.yaml.example
adrienthebo: I can point it to a tagged version.

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
+    # branch.
+    #
+    # See https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#prefix
+    # for more information about the prefix setting.
+    prefix: true
+
+# Additional configuration can be supplied to configure how r10k uses Git
+# and what version of Git it uses.
+#
+# See https://github.com/puppetlabs/r10k/tree/master/doc/git for more
```
cowofevil: Are you sure you want to point to master? What if the documentation changes in the FOSS version, but a PE 3.8 user is reading this example file?

adrienthebo:
r10k documentation links updated to pin the 1.5.1 tree.

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
-:cachedir: '/var/cache/r10k'
-:sources:
-  :plops:
-    remote: 'git@github.com:my-org/org-shared-modules'
-    basedir: '/etc/puppet/environments'
+---
```
andersonmills, line 1: I think a first line that says something like, "This is r10k.yaml. This is used to configure the behavior of r10k when run with the "deploy" action.

r10k.yaml.example
cowofevil: My bad, I'm always in PE mode, you're right the link is for Puppet so it doesn't matter.

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
+
+# The cachedir setting controls where cached content, such as mirrored Git
+# repositories, are stored. This location should be persistent, as environments
```
andersonmills: Maybe stored should be "stored on the local machine"?

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
+# and modules may rely on these files in order to be updated.
+#
+# See https://github.com/puppetlabs/r10k/blob/1.5.1/doc/git/cloning-and-mirroring.mkd
+# for more information on how r10k caches Git repositories.
+#
+# The default value is "~/.r10k"
+cachedir: '/var/cache/r10k'
+
+# The sources setting determines what repositories r10k will use when creating
+# Puppet environments. The structure is a map of source names and a map of the
```
andersonmills: I find "a map and a map" a little confusing. Maybe something like "is a map of source names to a map of that source's settings" would be better? "to each source's settings"?

r10k.yaml.example
adrienthebo: Done.

r10k.yaml.example
adrienthebo, line 1: Done.

r10k.yaml.example
```diff
@@ -1,5 +1,79 @@
...
+    # for more information about the Puppet 'environmentpath' setting.
+    basedir: '/etc/puppetlabs/puppet/environments'
+
+  # One or more sources can be specified; each source is simple another entry
+  # in the sources map.
+  qa:
+    remote: 'git@github.com:my-org/org-qa-modules'
+
+    # Multiple sources can have the same 'basedir' setting, in case two
+    # different groups want to push different sets of environments to the same
```
andersonmills: Maybe change "in case two different groups want" to "in case there is a reason to"

r10k.yaml.example
adrienthebo: "a map and a map" is confusing, but I don't know how to clearly document that it's a nested structure.

r10k.yaml.example
```diff
@@ -1,5 +1,83 @@
...
+    # Puppet master. If two different sources have the same basedir care should
+    # be exercised to make sure that two sources don't try create multiple
+    # environments with the same name. See the 'prefix' setting for more
+    # information.
+    basedir: '/etc/puppetlabs/puppet/environments'
+
+    # To avoid collisions between multiple sources being created in the same
+    # directory sources can use the prefix setting. Setting prefix to true
```
andersonmills: I think add a comma between "directory" and "sources".

r10k.yaml.example
```diff
@@ -1,5 +1,83 @@
...
+
+# Additional configuration can be supplied to configure how r10k uses Git
+# and what version of Git it uses.
+#
+# See https://github.com/puppetlabs/r10k/tree/1.5.1/doc/git for more
+# information on how r10k can configure Git.
+git:
+
+  # As of 1.5.0 r10k can interact with Git repositories in two ways - by
+  # shelling out to the 'git' executable, and by using libgit2.
```
andersonmills: I think change "libgit2" here to "libgit2 through the rugged library." Otherwise why would "rugged" be the other choice for provider below?

adrienthebo: Done.

r10k.yaml.example
adrienthebo: What about leaving the description of the structure blank and letting people figure it out based on the actual document structure and the following documentation?

r10k.yaml.example
acidprime: @adrienthebo from what I heard from eric0 there will be a 3.8 to match the PE release but be the same as 3.7 code. He said it publicly at a PUG the other night.

adrienthebo: So we can either set this link to documentation that will be out of date by the time PE 3.8 ships, set this link to documentation that hasn't been published, or set this link to the latest documentation that might change the meaning of this setting. Pick your poison.

adrienthebo: Or we can just omit the link entirely and sidestep this issue.

adrienthebo: I've updated this to point to the `/latest/` uri; it'll stay up to date until the semantics of the setting changes, at which point we'll need to change this setting regardless.

adrienthebo:
I think that the last comment needing addressing was with regard to the environmentpath documentation; this has now been addressed so is there anything else we need to do before this can be merged?

cowofevil:
+1 for merge.

andersonmills:
:+1: All the links are live. The content is accurate and well presented.

PR #411, opened 2015-05-04
[forge] (RK-96) forge module status is incorrect when previously a git checkout

thlapin:
Return mismatched status when a forge module was previsously a git
module.

### DIFF
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -83,6 +83,7 @@ class R10K::Module::Forge < R10K::Module::Base
   # @return [Symbol] :absent If the directory doesn't exist
   # @return [Symbol] :mismatched If the module is not a forge module, or
   #   isn't the right forge module
+  # @return [Symbol] :mismatched If the module was previously a git checkout
   # @return [Symbol] :outdated If the installed module is older than expected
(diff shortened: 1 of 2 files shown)


adrienthebo:
@thlapin could you clarify where this issue came up and how it's a problem? If you switch from having a version of a module cloned via Git pointing to a tag to one installed from the Forge pointing to the same release, the code should be identical. I'm somewhat concerned about this change as it leaks details about Git into the Forge module class, which I'm not thrilled by.

thlapin:
This was noticed with a workflow where you create a fork of an existing module to test out a change/fix. The Puppetfile is updated and r10k deploys the forked git repo with the applicable ref in this case a branch. Then if you switch back to using a forge modulel, if the version in the metadata hasn't changed then the module is not seen as needing an update and the git checkout remains.

adrienthebo:
And with this, the important thing is that you may have a Git based module with a number of changes that don't get reverted when you try to switch back to the Forge module?

thlapin:
Exactly. The work around has been to delete the module reference and then push another commit adding the module back as a forge module.

I agree that having a check for git in the forge module is not ideal. So if this can be done elsewhere, then that's fine.

adrienthebo:
Thanks for the contribution!

PR #412, opened 2015-05-05
[shared/puppet_forge] Add support for downloading forge modules

adrienthebo:
This commit adds the scaffording for direct integration with the Puppet
Forge. Right now this is limited to downloading modules but this will
expand as more functionality is needed.

### DIFF
--- a/lib/shared/puppet_forge/v3/module.rb
+++ b/lib/shared/puppet_forge/v3/module.rb
@@ -1,0 +1,31 @@
+require 'shared/puppet_forge/v3'
+require 'shared/puppet_forge/v3/module_release'
+require 'shared/puppet_forge/connection'
+
+module PuppetForge
+  module V3
+    # Represents metadata for a single Forge module and provides access to
+    # the releases of a module.
+    class Module
+
+      include PuppetForge::Connection
+
+      # @!attribute [r] full_name
+      #   @return [String] The hyphen separated full name of this module
+      attr_reader :full_name
+
+      # @param full_name [String] The name of this module, will be normalized to
+      #   a hyphen separated name.
+      def initialize(full_name)
+        @full_name = PuppetForge::V3.normalize_name(full_name)
+      end
+
+      # Get a specific release of this module off of the forge.
+      #
+      # @return [ModuleRelease] a release object of the given version for this module.
+      def release(version)
+        PuppetForge::V3::ModuleRelease.new(@full_name, version).tap { |mr| mr.conn = conn }
+      end
+    end
+  end
+end
--- a/lib/shared/puppet_forge/v3/module_release.rb
+++ b/lib/shared/puppet_forge/v3/module_release.rb
@@ -1,0 +1,73 @@
+require 'shared/puppet_forge/v3'
+require 'shared/puppet_forge/connection'
+
+module PuppetForge
+  module V3
+    # Access metadata and downloads for a specific module release.
+    class ModuleRelease
+
+      include PuppetForge::Connection
+
+      # @!attribute [r] full_name
+      #   @return [String] The hyphen delimited name of this module
+      attr_reader :full_name
+
+      # @!attribute [r] version
+      #   @return [String] The version of this module
+      attr_reader :version
+
+      # @param full_name [String] The name of the module, will be normalized
+      #   to a hyphen delimited name.
+      # @param version [String]
+      def initialize(full_name, version)
+        @full_name = PuppetForge::V3.normalize_name(full_name)
+        @version   = version
+      end
+
+      # @return [Hash] The complete Forge resposne for this release.
+      def data
+        @data ||= conn.get(resource_url).body
+      end
+
+      # @return [String] The unique identifier for this module release.
+      def slug
+        "#{full_name}-#{version}"
+      end
+
+      # Download this module release to the specified path.
+      #
+      # @param path [Pathname]
+      # @return [void]
+      def download(path)
+        resp = conn.get(file_url)
+        path.open('wb') { |fh| fh.write(resp.body) }
+      end
+
+      # Verify that a downloaded module matches the checksum in the metadata for this release.
+      #
+      # @param path [Pathname]
+      # @return [void]
+      def verify(path)
+        expected_md5 = data['file_md5']
+        file_md5     = Digest::MD5.file(path).hexdigest
+        if expected_md5 != file_md5
+          raise ChecksumMismatch.new("Expected #{path} checksum to be #{expected_md5}, got #{file_md5}")
+        end
+      end
+
+      private
+
+      def file_url
+        "/v3/files/#{slug}.tar.gz"
+      end
+
+      def resource_url
+        "/v3/releases/#{slug}"
+      end
+
+      class ChecksumMismatch < StandardError
+
+      end
+    end
+  end
+end
--- a/lib/shared/puppet_forge/version.rb
+++ b/lib/shared/puppet_forge/version.rb
@@ -1,0 +1,3 @@
+module PuppetForge
+  VERSION = '2.0.0.alpha'
+end
--- a/spec/unit/puppet_forge/connection_spec.rb
+++ b/spec/unit/puppet_forge/connection_spec.rb
@@ -1,0 +1,31 @@
+require 'shared/puppet_forge/connection'
+
+describe PuppetForge::Connection do
+
+  let(:extended) { Object.new.extend(described_class) }
+
+  describe 'creating a new connection' do
+
+    let(:faraday_stubs) { Faraday::Adapter::Test::Stubs.new }
+
+    subject { extended.make_connection('https://some.site/url', [:test, faraday_stubs]) }
+
+    it 'parses response bodies with a JSON content-type into a hash' do
+      faraday_stubs.get('/json') { [200, {'Content-Type' => 'application/json'}, '{"hello": "world"}'] }
+      expect(subject.get('/json').body).to eq('hello' => 'world')
+    end
+
+    it 'returns the response body as-is when the content-type is not JSON' do
+      faraday_stubs.get('/binary') { [200, {'Content-Type' => 'application/octet-stream'}, 'I am a big bucket of binary data'] }
+      expect(subject.get('/binary').body).to eq('I am a big bucket of binary data')
+    end
+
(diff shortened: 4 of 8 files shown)


adrienthebo:
Pulled VCR removal out into #414

lib/shared/puppet_forge/module_name.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def initialize(*args)
+      super
+      tr!('/', '-')
+    end
+
+    # @return [String] The slash separated module name.
+    def forge_name
+      String.new(self).tr('-', '/')
+    end
```
scotje: Since this subclasses string can't we just do `self.tr('-', '/')` here? 

But bigger picture, do we even need this method here? The fact that the Forge website uses "/" is really just an implementation detail that I'm not sure consumers of this library need to be concerned with. If we decide to keep it, something like "to_path" might be a more appropriately generic name for it.

lib/shared/puppet_forge/v3/module.rb
```diff
@@ -0,0 +1,31 @@
...
+require 'shared/puppet_forge/connection'
+require 'shared/puppet_forge/v3/module_release'
+
+module PuppetForge
+  module V3
+    # Represents metadata for a single Forge module and provides access to
+    # the releases of a module.
+    class Module
+
+      include Connection
```
scotje: Does this need to include the namespace?

lib/shared/puppet_forge/module_name.rb
```diff
@@ -0,0 +1,23 @@
...
+  # Normalize Forge module names
+  #
+  # Forge modules use '-' as the separator between the username and modulename
+  # in the module slug, but the Puppet Forge uses '/' as the separator on module
+  # pages and in the URL. This class hides the distinction by always using '-' as
+  # the separator for normal use. If the name needs to use the slash separated name
+  # then the #forge_name method can be used.
+  #
+  # @api private
+  class ModuleName < String
```
scotje: I'm wondering if ModuleName should also be namespaced under V3, it seems unlikely but not impossible that a future Forge API might change the syntax of module names...

lib/shared/puppet_forge/module_name.rb
adrienthebo: That's what I originally did, but if we do `self.tr` then we'll create another ModuleName object (I think) that will normalize it back the other way. I really need to write tests to verify this though.

I think we should have this method provided in some fashion, but I think that `#to_path` is a better name.

But in general, what do you think of the approach of wrapping interactions with module names in this object to remove the `#tr` calls all over the place?

lib/shared/puppet_forge/v3/module.rb
adrienthebo: Through the power of Ruby and dynamic scoping we don't _need_ it, but we shouldn't rely on said dynamic scoping. Will fix.

lib/shared/puppet_forge/v3/module_release.rb
```diff
@@ -0,0 +1,72 @@
...
+      # @param version [String]
+      def initialize(full_name, version)
+        @full_name = PuppetForge::ModuleName.new(full_name)
+        @version   = version
+      end
+
+      # @return [Hash] The metadata for this module
+      def data
+        @data ||= conn.get(resource_url).body
+      end
```
scotje, line 30: Long term we should have a more structured way of accessing the response, but this is probably fine for now. The comment should probably not claim this is the module metadata though since that is an actual specific part of the response. :)

lib/shared/puppet_forge/v3/module_release.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      # Download this module release to the specified path.
+      #
+      # @param path [Pathname]
+      # @return [void]
+      def download(path)
+        resp = conn.get(file_url)
+        path.open('wb') { |fh| fh.write(resp.body) }
+      end
```
scotje, line 44: Are we planning on more specific error handling at some point? It seems like we probably shouldn't even write to the file unless it was a 2xx response. (`resp.success?`)

Edit: or does Faraday already raise on 4xx responses here?

lib/shared/puppet_forge/module_name.rb
adrienthebo: Doing a `#tr` directly will create a new ModuleName object that has a forward slash separator, which isn't want I would expect since the class tries to reverse that on creation - so I think returning an explicit string is a bit more complexity but less surprising in behavior.

```
[1] pry(#<RSpec::ExampleGroups::PuppetForgeModuleName>)> subject.to_path
=> "authorname/modulename"
[2] pry(#<RSpec::ExampleGroups::PuppetForgeModuleName>)> subject.to_path.class
=> PuppetForge::ModuleName
```

What do you think? Go with `#tr` or do the explicit clone/modify?

lib/shared/puppet_forge/version.rb
```diff
@@ -0,0 +1,3 @@
+module PuppetForge
+  VERSION = '2.0.0dev'
```
scotje: Let's call it 2.0.0-alpha

lib/shared/puppet_forge/module_name.rb
adrienthebo: Sounds good to me.

lib/shared/puppet_forge/v3/module_release.rb
adrienthebo, line 30: Ah yes - the "metadata" was a holdover.

spec/unit/module_repository/forge_spec.rb
```diff
@@ -19,15 +34,33 @@
...
+      }
+      JSON
+
+      spotty_body = <<-JSON
+      {"releases": [
+          {"version": "0.4.0", "deleted_at": "some point"},
+          {"version": "0.3.0"}
+        ]
+      }
+      JSON
```
scotje: Might be cleaner to define these as hashes and just call to_json on them? Heredoc is going to be full of leading whitespace.

lib/shared/puppet_forge/v3/module_release.rb
adrienthebo, line 44: Faraday itself is handling errors with the `raise_error` middleware - if a response has an HTTP status code in `(400....600)` an exception is thrown so we don't need to explicitly add error handling for each HTTP request unless we need it.

lib/shared/puppet_forge/version.rb
adrienthebo: Gem versioning doesn't allow a hyphen between the patch release and the prerelease but I can do `2.0.0alpha`

scotje: Looks like the rubygems pattern is a dot seperator: http://guides.rubygems.org/patterns/#prerelease-gems

lib/shared/puppet_forge/v3/module_release.rb
adrienthebo, line 30: And in the longer term I intend to split out attr_readers or similar for each field - but for now I'm aiming for enough functionality to get things working without trying to build in more functionality than we need.

lib/shared/puppet_forge/version.rb
adrienthebo: Rubyyyyyyyyyy

spec/unit/module_repository/forge_spec.rb
adrienthebo: The original intent was to use the same faraday middleware as the actual application which would parse the JSON and ignore any whitespace, but that was causing issues with Ruby 1.8.7 so I replaced this with the actual data structures.

lib/shared/puppet_forge/module_name.rb
scotje: Ah yes, I agree that returning a string is preferable there.

As for the larger question of whether the class should exist, it seemed reasonable to me but this article leads me to believe there are a lot of dragons to watch out for when subclassing Ruby core: http://words.steveklabnik.com/beware-subclassing-ruby-core-classes

Specifically there are circumstances where Ruby will not call the initializer of your custom String subclass which seems like it might defeat the whole purpose of this class. :/ Something like his approach using Forwardable might be preferable so that we can explicitly control what gets delegated to String.

I think this is OK to merge but we probably should revisit at some point.

scotje:
:+1: with the caveat that we should probably explore the edge cases of ModuleName subclassing String at some point and decide if we're comfortable with it.

lib/shared/puppet_forge/module_name.rb
adrienthebo: That's a good point actually - the oddness that I pointed out above that required the explicit cast to string was a demonstration of this quirk. I'll give the Forwardable approach a once over to see if it's something we can stuff in really quickly.

adrienthebo:
Replaced `ModuleName` class with `PuppetForge::V3.module_name`

scotje:
:+1: with no caveats!

adrienthebo:
We should make sure to merge #414 before this.

adrienthebo:
Rebased onto master following the merge of #414.

lib/shared/puppet_forge/v3/module.rb
```diff
@@ -0,0 +1,31 @@
...
+  module V3
+    # Represents metadata for a single Forge module and provides access to
+    # the releases of a module.
+    class Module
+
+      include PuppetForge::Connection
+
+      # @!attribute [r] full_name
+      #   @return [String] The hyphen separated full name of this module
+      attr_reader :full_name
```
andersonmills: Shouldn't this ALSO be canonical_name?

spec/unit/puppet_forge/connection_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'shared/puppet_forge/connection'
+
+describe PuppetForge::Connection, :vcr => false do
```
andersonmills: Why is :vcr => false here?
Is this left over from removing it?

adrienthebo: Yep, leftover from integrating the Faraday::Adapter::Test based tests with the existing vcr tests when I tried to use both; I've removed this.

andersonmills:
:+1: I just tested this final single-variable name change, and I'm happy with this.

PR #414, opened 2015-05-05
[spec] Remove vcr

adrienthebo:
VCR is effectively unmaintained at this point and the faraday test
adapter can serve as an effective replacement.


andersonmills:
:+1:  This works, is a good idea, and passes tests (well _is_ tests which pass).

PR #415, opened 2015-05-06
(CODEMGMT-12) Install modules from the forge

adrienthebo:
This adds a class that orchestrates downloading, verifying, unpacking, and installing a module. Still needs the code for downloading/verifying modules, unpacking modules, and a better settings system.


andersonmills:
:+1: from me. This code will change with the upcoming r10k/forge integration ticket, but I think that's absolutely fine.

PR #416, opened 2015-05-06
[CODEMGMT-27] Unpack/untar in r10k as library

andersonmills:
This commit adds the library code which can unpack/untar module
tarballs. This unpacking code is not completely generic in that at a low
level it warns about symlinks in tarballs.

### DIFF
--- a/lib/shared/puppet_forge/tar/mini.rb
+++ b/lib/shared/puppet_forge/tar/mini.rb
@@ -1,0 +1,75 @@
+require 'zlib'
+require 'archive/tar/minitar'
+
+module PuppetForge
+  class Tar
+    class Mini
+
+      SYMLINK_FLAG = 2
+      VALID_TAR_FLAGS = (0..7)
+
+      def unpack(sourcefile, destdir)
+        dirlist = []
+        Zlib::GzipReader.open(sourcefile) do |reader|
+          Archive::Tar::Minitar.unpack(reader, destdir, find_valid_files(reader)) do |action, name, stats|
+            case action
+            when :file_done
+              FileUtils.chmod('u+rw,g+r,a-st', "#{destdir}/#{name}")
+            when :file_start
+              validate_entry(destdir, name)
+            when :dir
+              validate_entry(destdir, name)
+              dirlist << "#{destdir}/#{name}"
+            end
+          end
+        end
+        dirlist.each {|d| File.chmod(0755, d)}
+      end
+
+      def pack(sourcedir, destfile)
+        Zlib::GzipWriter.open(destfile) do |writer|
+          Archive::Tar::Minitar.pack(sourcedir, writer)
+        end
+      end
+
+      private
+
+      # Find all the valid files in tarfile.
+      #
+      # Symlinks are not supported in modules
+      #
+      # This check was mainly added to ignore 'x' and 'g' flags from the PAX
+      # standard but will also ignore any other non-standard tar flags.
+      # tar format info: http://pic.dhe.ibm.com/infocenter/zos/v1r13/index.jsp?topic=%2Fcom.ibm.zos.r13.bpxa500%2Ftaf.htm
+      # pax format info: http://pic.dhe.ibm.com/infocenter/zos/v1r13/index.jsp?topic=%2Fcom.ibm.zos.r13.bpxa500%2Fpxarchfm.htm
+      def find_valid_files(tarfile)
+        Archive::Tar::Minitar.open(tarfile).collect do |entry|
+          flag = entry.typeflag
+          # Symlinks are not supported in modules and a warning is issued (flag '2' is a symlink)
+          if flag.nil? || flag =~ /[[:digit:]]/ && SYMLINK_FLAG == flag.to_i
+            # TODO: Error/Warn about symlinks here.
+            entry.name
+          elsif flag.nil? || flag =~ /[[:digit:]]/ && VALID_TAR_FLAGS.include?(flag.to_i)
+            entry.name
+          else
+            next
+          end
+        end
+      end
+
+      def validate_entry(destdir, path)
+        if Pathname.new(path).absolute?
+          raise PuppetForge::InvalidPathInPackageError, :entry_path => path, :directory => destdir
+        end
+
+        path = File.expand_path File.join(destdir, path)
+
+        if path !~ /\A#{Regexp.escape destdir}/
+          raise PuppetForge::InvalidPathInPackageError, :entry_path => path, :directory => destdir
+        end
+      end
+    end
+  end
+
+end
+
--- a/lib/shared/puppet_forge/unpacker.rb
+++ b/lib/shared/puppet_forge/unpacker.rb
@@ -1,0 +1,65 @@
+require 'pathname'
+require 'shared/puppet_forge/error'
+require 'shared/puppet_forge/tar'
+
+module PuppetForge
+  class Unpacker
+    # Unpack a tar file into a specified directory
+    #
+    # @param filename [String] the file to unpack
+    # @param target [String] the target directory to unpack into
+    def self.unpack(filename, target, tmpdir)
+      inst = self.new(filename, target, tmpdir)
+      inst.unpack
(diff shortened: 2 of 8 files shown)


lib/shared/puppet_forge/tar/mini.rb
```diff
@@ -0,0 +1,74 @@
...
+      #
+      # This check was mainly added to ignore 'x' and 'g' flags from the PAX
+      # standard but will also ignore any other non-standard tar flags.
+      # tar format info: http://pic.dhe.ibm.com/infocenter/zos/v1r13/index.jsp?topic=%2Fcom.ibm.zos.r13.bpxa500%2Ftaf.htm
+      # pax format info: http://pic.dhe.ibm.com/infocenter/zos/v1r13/index.jsp?topic=%2Fcom.ibm.zos.r13.bpxa500%2Fpxarchfm.htm
+      def find_valid_files(tarfile)
+        Archive::Tar::Minitar.open(tarfile).collect do |entry|
+          flag = entry.typeflag
+          if flag.nil? || flag =~ /[[:digit:]]/ && 2 == flag.to_i
+            @logger.warning("Symlinks in modules are unsupported. Please investigate symlink #{entry.name}.")
```
adrienthebo: I found a hitch in this - Puppet uses `logger.warning` while r10k uses `logger.warn` so there's a mismatch in which method we should call. I'm not sure how viable injecting a logger will be if we have discrepancies like this.

lib/shared/puppet_forge/unpacker.rb
```diff
@@ -0,0 +1,78 @@
...
+      inst = self.new(filename, target, tmpdir)
+      inst.unpack
+      inst.move_into(target)
+    end
+
+    # Set the owner/group of the target directory to those of the source
+    #
+    # @param source [Pathname] source of the permissions
+    # @param target [Pathname] target of the permissions change
+    # @param microsoft_windows boolean flag for whether this system is windows or not
```
adrienthebo: It looks like nothing in this codebase invokes `.harmonize_ownership`, can we make callers make the check to see if the platform is Windows and conditionally call this method instead of having this method do the check?

lib/shared/puppet_forge/unpacker.rb
```diff
@@ -0,0 +1,78 @@
...
+    def unpack
+      begin
+        PuppetForge::Tar.instance.unpack(@filename.to_s, @tmpdir)
+      rescue PuppetForge::ExecutionFailure => e
+        raise RuntimeError, "Could not extract contents of module archive: #{e.message}"
+      end
+    end
+
+    # @api private
+    def move_into(dir)
```
adrienthebo, line 44: Is it intended for `#move_into` to move the unpacked module into the final location?

andersonmills, line 44: Yessir.

lib/shared/puppet_forge/unpacker.rb
andersonmills: Yep. That works for me. Working on it.

lib/shared/puppet_forge/unpacker.rb
adrienthebo, line 44: This is problematic because the directory that will be moved is in the format `username-modulename-version`, which means that we can't move this from the temp location to the final location without stuffing another rename into the mix.

```
    52: def move_into(dir)
    53:   dir = Pathname.new(dir)
    54:   dir.rmtree if dir.exist?
 => 55:   require 'pry'; binding.pry
    56:   FileUtils.mv(root_dir, dir)
    57: ensure
    58:   FileUtils.rmtree(tmpdir)
    59: end

[1] pry(#<PuppetForge::Unpacker>)> dir
=> #<Pathname:/tmp/d20150513-13668-10i1lta>
[2] pry(#<PuppetForge::Unpacker>)> root_dir
=> #<Pathname:/tmp/tmp-unpacker20150513-13668-1x49wtg/branan-eight_hundred-8.0.0>
```

adrienthebo, line 44: That being said, r10k can just skip this call and move the contents of `#root_dir` and skip this secondary move.

lib/shared/puppet_forge/unpacker.rb
```diff
@@ -0,0 +1,78 @@
...
+    def self.harmonize_ownership(source, target, microsoft_windows = false)
+      if !microsoft_windows
+        FileUtils.chown_R(source.stat.uid, source.stat.gid, target)
+      end
+    end
+
+    # @param filename [String] the file to unpack
+    # @param target [String] the target directory to unpack into
+    def initialize(filename, target, tmpdir)
+      @filename = Pathname.new(filename)
```
adrienthebo: We convert `filename` to a `Pathname` here, and on line 40 we call `@filename.to_s` which means we're round tripping through Pathname to a string. Normally I really prefer using Pathname objects over Strings because it's easier to test and stub, but it doesn't look like we're actually taking advantage of any functionality on Pathname. Does it make sense to do this for consistency or some other reason?

lib/shared/puppet_forge/unpacker.rb
```diff
@@ -0,0 +1,78 @@
...
+      @target = target
+      @tmpdir = tmpdir
```
adrienthebo, line 31: The `#tmpdir` method memoizes the `@tmpdir` value, but here we assign it - so this assignment looks redundant.

lib/shared/puppet_forge/unpacker.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    # @api private
+    def unpack
+      begin
+        PuppetForge::Tar.instance.unpack(@filename.to_s, @tmpdir)
```
adrienthebo: Wait, here we use `@tmpdir` but in other places we use the `#tmpdir` attr_reader.

lib/shared/puppet_forge/unpacker.rb
adrienthebo, line 31: Wait. I'm insane. We only use `@tmpdir` in this class but the `#tmpdir` method does memoization, so it looks like the memoization is unnecessary.

lib/shared/puppet_forge/unpacker.rb
```diff
@@ -0,0 +1,78 @@
...
+      rescue PuppetForge::ExecutionFailure => e
+        raise RuntimeError, "Could not extract contents of module archive: #{e.message}"
+      end
+    end
+
+    # @api private
+    def move_into(dir)
+      dir = Pathname.new(dir)
```
adrienthebo: Can we require that the caller pass a `Pathname` object instead of expecting a string and casting to a `Pathname`?

lib/shared/puppet_forge/unpacker.rb
andersonmills, line 44: The behavior of FileUtils::mv is correct for the copying of the contents of the root_dir to the dir, so this doesn't need to change.

lib/shared/puppet_forge/unpacker.rb
andersonmills: Not that is still a part of the library. Changed to keeping it a string.

lib/shared/puppet_forge/unpacker.rb
andersonmills, line 31: Sorry, I left the #tmpdir method in there by accident. Removed.

lib/shared/puppet_forge/unpacker.rb
andersonmills: This has now been made consistent as @tmpdir

lib/shared/puppet_forge/unpacker.rb
andersonmills: Sure. It's private. We're in charge!

lib/shared/puppet_forge/unpacker.rb
adrienthebo, line 44: Yep, this was the same thing that the PMT did with two tempdirs. I've fixed this on my end.

Issue #54, opened 2013-08-14
Feature: Parallel checkout of modules


lindenle:
It would be great if r10k could take a parallel job parameter to speed up checkout of modules.

rbowlby:
+1

adrienthebo:
I've been planning on adding this feature, it's a bit tricky to do correctly but I think this is a key feature. I'm initially targeting it at 1.2.0.

lindenle:
http://jessewolgamott.com/blog/2013/02/07/the-one-where-i-introduce-celluloid-pmap/

Could you use that in the task runner? 
In this code (https://github.com/adrienthebo/r10k/blob/master/lib/r10k/task_runner.rb#L23):

``` javascript
  until @tasks.empty?
        current = @tasks.first
        current.task_runner = self
        begin
          current.call
        rescue Interrupt => e
          logger.error "Aborted!"
          $stderr.puts e.backtrace.join("\n").red if @trace
          @succeeded = false
          throw :abort
        rescue => e
          logger.error "Task #{current} failed while running: #{e.message}"
          $stderr.puts e.backtrace.join("\n").red if @trace

          @errors[current] = e
          @succeeded = false
        end
        @tasks.shift
```

luisfdez:
@scotje connected to this issue, is there any roadmap to add some kind of parallelism to `r10k`. I think it's a must to scale it properly.

lvlie:
+1

alex-harvey-z3q:
+1

alexharv074:
@adrienthebo  Is this on the roadmap yet?  I've just parallelised all of the Rspec builds for my customer and now I have to tell them "Um, however I can't parallelise it in production".

lpciuser:
+1

maxadamo:
@adrienthebo you only need to implement threaded execution for the methods. Is it really tricky to implement? Or is there something that I am overlooking, underestimating?

theody-magill-amp:
+1

maxadamo:
Of course,  an option to limit the number of executions should be mandatory. In my case I have 80 remote repositories.

dominics:
@maxadamo see the discussion in https://github.com/puppetlabs/r10k/pull/417 for some of the difficulties (although the naive approach in that PR still works great for my use case, and I haven't run into any real-world parallelism issues)

maxadamo:
[...]

PR #417, opened 2015-05-07
Add --parallel option to `r10k puppetfile install`

dominics:
This is a quick proof-of-concept for fixing #54 - it adds a `-p`/`--parallel` option, which is the number of modules from the Puppetfile to process in parallel.

The underlying implementation is the `pmap` convenience library (https://rubygems.org/gems/pmap), but could just as easily be something else (or an r10k-specific implementation).

Remaining work includes:
- Tests for the CLI interface additions?
- A default value for the CLI option itself? (rather than via a default parameter on `R10K::Puppetfile.accept`)
## Performance

The Puppetfile I'm most often using has somewhere around 100 modules listed. The install operation is completely IO bound, and doing it all serially takes almost four minutes(!)

``` sh
$ time bundle exec r10k puppetfile install

real  3m43.955s
user  0m4.142s
sys   0m4.789s
```

With a modest amount of parallelism, we get a **20 times faster** operation (11 seconds).

``` sh
$ time bundle exec r10k puppetfile install -p 32

real  0m11.055s
user  0m4.265s
sys   0m5.395s
```

I understand if we want to delay this fix until we cover things like the `deploy` subcommand. But #54 has been open for a couple of years now. And I'd describe this performance improvement as necessary, not a nice-to-have, however we implement it.

### DIFF
--- a/lib/r10k/action/puppetfile/install.rb
+++ b/lib/r10k/action/puppetfile/install.rb
@@ -21,12 +21,13 @@ module R10K
             :moduledir  => :self,
             :puppetfile => :path,
             :trace      => :self,
+            :parallel   => :self
           })
         end
 
         def call
           pf = R10K::Puppetfile.new(@root, @moduledir, @path)
-          pf.accept(self)
+          pf.accept(self, @parallel.to_i)
           @ok
         end
 
--- a/lib/r10k/cli/puppetfile.rb
+++ b/lib/r10k/cli/puppetfile.rb
@@ -31,6 +31,8 @@ Puppetfile (http://bombasticmonkey.com/librarian-puppet/).
           usage   'install'
           summary 'Install all modules from a Puppetfile'
 
+          required :p, :parallel, 'Number of module installations to conduct in parallel', argument: :required
+
           # @todo add --moduledir option
           # @todo add --puppetfile option
           # @todo add --no-purge option
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -2,6 +2,7 @@ require 'pathname'
 require 'r10k/module'
 require 'r10k/util/purgeable'
 require 'r10k/errors'
+require 'pmap'
 
 module R10K
 class Puppetfile
@@ -88,10 +89,16 @@ class Puppetfile
     @modules.map { |mod| mod.name }
   end
 
-  def accept(visitor)
+  def accept(visitor, parallel=1)
     visitor.visit(:puppetfile, self) do
-      modules.each do |mod|
-        mod.accept(visitor)
+      if parallel > 1
+        modules.peach(parallel) do |mod|
+          mod.accept(visitor)
+        end
+      else
+        modules.each do |mod|
+          mod.accept(visitor)
+        end
       end
     end
   end
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'colored',   '1.2'
   s.add_dependency 'cri',       '~> 2.6.1'
+  s.add_dependency 'pmap',     '~> 1.0.2'
 
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'


adrienthebo:
Parallelism can be deceptively tricky, because while the base case of "do different things at once" seems straightforward it also brings along a host of issues. For instance, what happens if a module fails to install? How do we report that? If we have multiple environments that have the same modules, how do we make sure that things like cached Git repos aren't updated at the same time? How do we make sure that multiple threads don't try to access the same resources?

In addition, you said that it takes 4 minutes to install 100 modules - are these modules all installed from the Puppet Forge or are these modules coming from Git?

dominics:
> For instance, what happens if a module fails to install? How do we report that?

In both the parallel and non-parallel cases:
- We log the error (luckily, with a single log statement, so there's no risk of the threading causing interpolated output from different threads)
- We continue to process other modules
- We eventually return an exit status of 1 if any of the sub-processes failed, and 0 if they didn't (this works exactly the same)

So, I can't see any behaviour differences around error reporting.

> If we have multiple environments that have the same modules, how do we make sure that things like cached Git repos aren't updated at the same time?

In the specific case of Git repos, objects are immutable, and references are updated atomically. Where these don't apply (`git gc`, `git commit`), there are things like `.git/index.lock`. So, basically by relying on Git to do the right thing.

> How do we make sure that multiple threads don't try to access the same resources?

Well, in this PR, by only spawning threads to exec sub-processes. That way, the mutable resources are in a completely different address space, and the threading is just being used to wait/join on the results.

> In addition, you said that it takes 4 minutes to install 100 modules - are these modules all installed from the Puppet Forge or are these modules coming from Git?

About 20% from the Forge, 80% from Git. But many of the Git ones are private/roles/profiles, so we won't be able to publish them (unless we use Pulp or some other sort of self-hosted Forge). The assumption was that pretty much any form of module install would be similarly IO bound, often on networking (remote packages are the norm.)

lazyfrosch:
Sounds very interesting

adrienthebo:
If you're using a lot of Git modules you should be getting a lot of benefit from the caching of Git repositories that r10k does under the hood; are you somehow not reusing the cached Git repositories? In environments with cached Git repositories I've seen 100 modules created in ~15 seconds, so if you're seeing r10k take 4 minutes to take 100 repositories then I suspect that caching isn't working.

With respect to accessing the same resource, I'm not worried about the address space itself - I'm worried about accessing filesystem resources. If two threads try to run a `git fetch` on the same repository, weird things happen - and since r10k does cache repos this is a real risk.

What I'm getting at is that 1) I'm interested in doing this, but 2) incorporating parallelism generally creates a slew of hard to resolve bugs, and I don't have the time or resources right now to check for potential races and fix resulting bugs. I'm happy to leave this pull request open but I won't be able to merge this until parallelism can be fully incorporated into the codebase rather than adding it in an incremental manner.

dominics:
> I won't be able to merge this until parallelism can be fully incorporated into the codebase rather than adding it in an incremental manner

Understood, and a fine rationale.

> If two threads try to run a git fetch on the same repository, weird things happen

What weird things specifically? Like I said, references are updated atomically (ultimately by the `rename` syscall, which is atomic as long as you're not doing something weird like hosting half a repo on a different filesystem), and objects are referenced by their content, so I don't see how running two fetches on the same repository causes anything weird to happen at all.

Besides that, can you explain how two threads fetching for a local repo is at all different from two processes pushing to a remote repo at the same time? In both situations, there are objects being transferred, and then references updated. If Git couldn't do that in a concurrency-safe manner, a bare repo on disk wouldn't be able to act as a reasonable origin/upstream. (Yet, we know it does.)

> are you somehow not reusing the cached Git repositories?

No, we are definitely using the cached Git repositories. In fact, the time is spent _updating the cached repos_, not updating the working copies.

> I've seen 100 modules created in ~15 seconds

That's certainly not possible with our Git remotes.

```
$ time git --git-dir /home/<some-user>/.r10k/git/<cache-repo> fetch --prune

real    0m2.572s
user    0m0.019s
sys 0m0.012s
```

(`fetch --prune` on the cache repo is shown by `--verbose=debug2` if anyone else is following along.)

I can't see that referring to a specific commit or tag will change the fetch behaviour, but I guess it could (I could only see references to a `*/*` refspec). If that's the case, it should be prominently mentioned.

Other things that might influence this would be provider (I guess I should be looking into libgit2/rugged?) and protocol (SSH vs HTTPS)

adrienthebo:
> Besides that, can you explain how two threads fetching for a local repo is at all different from two processes pushing to a remote repo at the same time? In both situations, there are objects being transferred, and then references updated. If Git couldn't do that in a concurrency-safe manner, a bare repo on disk wouldn't be able to act as a reasonable origin/upstream. (Yet, we know it does.)

Git handles that by locking the repository; processes that lose the race condition will error out, which means that users may see spurious errors all the time. So yes - Git can act as a reasonable upstream that's "safe" for concurrent writes but it does that by using pessimistic locking and forcing users to handle delaying and retrying. This means that r10k will have to take on the onus of said delaying and retrying for locks.

In addition r10k also supports libgit2, and while I would assume that libgit2 has some sort of similar pessimistic locking I would have to explore how that would behave.

Do most/all of your Git based modules need to track branches, or are some of those modules pointing to commits or tags?

dominics:
> Do most/all of your Git based modules need to track branches, or are some of those modules pointing to commits or tags?

Most of them need to track branches, and yeah, some of them are pointed at tags/commits. Exact breakdown is:
- 14 forge modules with semvers
- 3 git with specific branches
- 1 git with a commit hash
- 87 git with no ref specified

I guess the last category is treated the same as a branch `ref`, because they're effectively following the ref in `~/.r10k/git/ssh---git<blah>/HEAD`.

It's going to be a mighty pain if we have to semver/tag our private modules to gain better performance. (And a lock file avoids any difference between the treatment of branches and tags - or you can look at it as the fetch call for the branch ref being cached behind the lock file. So, we'd probably look for a more suitable tool before considering semverring roles/profiles)

adrienthebo:
Recently we've been working on different implementations of parallelism, and unfortunately we're discovering a great deal of race conditions between different processes - running things in threads would introduce these issues as well. I appreciate your effort in this but we're currently not in a place where we can implement parallelism for installing modules and doing so in a partial manner would introduce a slew of bugs that we wouldn't be able to stay on top of right now. Apologies for not being able to get this merged and for the time it took to get a response on this.

PR #419, opened 2015-05-11
[module/forge] (RK-83) Allow '-' as a module name separator

adrienthebo:
Hyphens and forward slashes are used mostly interchangeably as the
module name separator, but the existing name matching code only allowed
forward slashes. Right now the hyphen is the preferred character, which
meant that users couldn't use the preferred pattern for Forge modules.
This commit adds hyphens as a valid separator character to remedy this.


adrienthebo:
Superceded by #420.

PR #422, opened 2015-05-19
(CODEMGMT-187) Report file status back from unpacking

andersonmills:
Before this pull request, there were no errors raised and no logging possible from unpacking a module. This commit adds a hash of file lists which is returned from the unpacker giving the lists of vaild, invalid, and symlink files.

### DIFF
--- a/lib/r10k/forge/module_release.rb
+++ b/lib/r10k/forge/module_release.rb
@@ -89,7 +89,17 @@ module R10K
       # @return [void]
       def unpack(target_dir)
         logger.debug1 "Unpacking #{download_path} to #{target_dir} (with tmpdir #{unpack_path})"
-        PuppetForge::Unpacker.unpack(download_path.to_s, target_dir.to_s, unpack_path.to_s)
+        file_lists = PuppetForge::Unpacker.unpack(download_path.to_s, target_dir.to_s, unpack_path.to_s)
+        logger.debug2 "Valid files unpacked: #{file_lists[:valid]}"
+        if !file_lists[:symlinks].empty?
+          logger.warn "Symlinks in modules are unsupported. These files were created as copies of the original " + 
+                      "files, but are no longer symlinks: #{file_lists[:symlinks]}"
+        end
+        if !file_lists[:invalid].empty?
+          logger.warn "These files existed in the module's tar file, but are invalid filetypes and were not " +
+                      "unpacked: #{file_lists[:invalid]}"
+        end
+
       end
 
       # Remove all files created while downloading and unpacking the module.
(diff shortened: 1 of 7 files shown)


lib/r10k/forge/module_release.rb
```diff
@@ -89,7 +89,17 @@ def verify
       # @return [void]
       def unpack(target_dir)
         logger.debug1 "Unpacking #{download_path} to #{target_dir} (with tmpdir #{unpack_path})"
-        PuppetForge::Unpacker.unpack(download_path.to_s, target_dir.to_s, unpack_path.to_s)
+        file_lists = PuppetForge::Unpacker.unpack(download_path.to_s, target_dir.to_s, unpack_path.to_s)
+        logger.debug2 "Valid files unpacked: #{file_lists[:valid]}"
+        if !file_lists[:symlinks].empty?
+          logger.warn "Symlinks in modules are unsupported. These files were created as copies of the original files, "\
```
adrienthebo: I would rather that we either use string concatenation or just have line rather than escaping newlines, right now I'm not actually sure how this works but it seems like it's using some sort of implicit concatenation .

lib/shared/puppet_forge/error.rb
```diff
@@ -26,5 +26,18 @@ def multiline
     end
   end
 
+  class SymlinkInModuleError < PuppetForge::Error
```
adrienthebo: Is this class used anywhere?

andersonmills: No, it isn't. This was left over from a previous mechanism for moving the symlink files back out. Removing it.

lib/r10k/forge/module_release.rb
```diff
@@ -89,7 +89,17 @@ def verify
       # @return [void]
       def unpack(target_dir)
         logger.debug1 "Unpacking #{download_path} to #{target_dir} (with tmpdir #{unpack_path})"
-        PuppetForge::Unpacker.unpack(download_path.to_s, target_dir.to_s, unpack_path.to_s)
+        file_lists = PuppetForge::Unpacker.unpack(download_path.to_s, target_dir.to_s, unpack_path.to_s)
+        logger.debug2 "Valid files unpacked: #{file_lists[:valid]}"
+        if !file_lists[:symlinks].empty?
+          logger.warn "Symlinks in modules are unsupported. These files were created as copies of the original " + 
+                      "files, but are no longer symlinks: #{file_lists[:symlinks]}"
```
adrienthebo, line 96: Minitar is making a full copy of the file under the hood, correct?

andersonmills, line 96: Correct.

PR #423, opened 2015-05-19
(CODEMGMT-179) Add Authorization header support to Forge connections

scotje:


lib/r10k/action/runner.rb
```diff
@@ -30,6 +37,19 @@ def setup_logging
...
 
       def setup_settings
+        setup_authorization
+      end
+
+      def setup_authorization
+        begin
+          license = R10K::Util::License.load
+
+          if license.respond_to?(:authorization_token)
```
adrienthebo, line 47: Why are we checking to see if a method on `license` was defined rather than seeing if the returned license was/was not nil?

scotje, line 47: Older versions of PE will return a license object that doesn't have that method.

adrienthebo, line 47: ಥ_ಥ

Okay.

PR #432, opened 2015-06-02
(RK-108) Add configuration option for forge proxy

adrienthebo:
This pull request allows r10k to set an HTTP proxy for connecting to the Puppet Forge, either via a setting in r10k.yaml or via one of the commonly used http proxy environment variables.


andersonmills:
:+1: This works! I love it! It's so squishy!

PR #433, opened 2015-06-03
(RK-92) Remove documentation links from example config

adrienthebo:
The example r10k.yaml configuration is being packed in PE and had
references to GitHub for documentation, and linking to open source
documentation from a PE example configuration is not desired. This
commit removes the relevant documentation links.


andersonmills:
Documentation change. No testing. Removing these links is approved.

:+1:

PR #435, opened 2015-06-04
(RK-53) Remove /etc/r10k.yaml from config file search path.

scotje:


adrienthebo:
:+1:, will merge after travis-ci passes.

adrienthebo:
Well, one thought - do we want to warn if '/etc/r10k.yaml' is present and indicate that it exists but isn't being used? One less surprise for users.

adrienthebo:
After discussion we decided to skip emitting warnings.

PR #436, opened 2015-06-04
(RK-47) Remove Ruby 1.8.7 compatibility

adrienthebo:
This pull request removes all Ruby 1.8.7 specific compatibility. This only drops Ruby 1.8.7 and doesn't go as far as dropping 1.9 and 2.0 support; that's another discussion that we'll need to work out at some point.


andersonmills:
Looks like you need to change the README.mkd.

adrienthebo:
Whoops - fixed the README

andersonmills:
:+1:  Zero function says, "everything looks like me!"

PR #437, opened 2015-06-04
(maint) Correct error message when no config file present.

andersonmills:
Before this, the error message for no config file present specified the parent directory. With this commit, that error message references the current directory.


adrienthebo:
Waiting for travis-ci.

PR #438, opened 2015-06-04
(RK-54) Remove deprecated r10k subcommands

adrienthebo:
These subcommands were used in r10k 0.0.9 and were deprecated over two
years ago, it's time to say goodbye to them.


adreyer:
+1 grepping only shows these mentioned in the changelog

PR #440, opened 2015-06-08
(Rk-144) Remove deprecated git classes

adrienthebo:
R10k 1.5.0 restructured the R10K::Git namespace and obsoleted a number
of classes in the Git namespaces; this commit removes those classes for
r10k 2.0.0.


adreyer:
It looks like shared_examples may still have some references to the deprecated classes

adrienthebo:
Good catch, I've removed that.

PR #446, opened 2015-06-09
(RK-117) Add minitar as runtime dependency

adrienthebo:
Minitar is required to install modules from the Puppet Forge, and is
required at application startup so it is a hard requirement. This commit
moves the dependency from a development dependency to a runtime
dependency.


andersonmills:
:+1: That's exactly what I did.

PR #448, opened 2015-06-09
(RK-70) add Forge baseurl setting

adrienthebo:
This allows users to specify a custom Forge URL for downloading and installing modules.


andersonmills:
:+1: I approve this. I even tested it against the On-Prem Forge with the help of @zreichert .  Yay!

PR #453, opened 2015-06-15
(Doc) Add invalid_branches setting to source options list

vinzent:
Add the invalid_branches setting to the list of source options. Copy/paste from git-environments.md.

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -188,6 +188,27 @@ sources:
 * if `false` (default) environment folder will not be prefixed
 * if `String` environment folder will be prefixed with the `prefix` value.
 
+### invalid_branches:
+
+This setting specifies how Git branch names that cannot be cleanly mapped to
+Puppet environments will be handled.
+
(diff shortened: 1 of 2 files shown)


adrienthebo:
Since this setting is already documented and is Git specific, could you clarify for me how duplicating this documentation provides value?

vinzent:
Isn't the "configuration.md" file the "r10k.yaml" man page? it should list all the possible options. I was searching for a while until i've found the reference on the git-environments.md file.

adrienthebo:
Per https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#source-options each source type can have different options; I don't want all options to necessarily be packed into a single file. That being said this is less than clear, so if you want to update the change to document the setting, indicate that it's git specific, and point to the git configuration document instead of copy pasting that I'll merge it.

vinzent:
i do know the option now. i don't need the documentation anymore. I think for new users, it would be very helpfull to have all options listed on one page. but it's up to you.

PR #456, opened 2015-06-17
(RK-109) Add context to connection failure errors

adrienthebo:
When Faraday encounters an issue establishing a connection to a host or
to a proxy, it raises an error indicating that the connection failed
but does not include the name of the host. This commit adds a middleware
class that rescues ConnectionFailed exceptions and raises a new error
that includes the host, and optional proxy host.


adreyer:
Looks good.

PR #457, opened 2015-06-18
(Gemspec) Update homepage url

vinzent:
Update url of r10k homepage link on https://rubygems.org/gems/r10k


adrienthebo:
Thanks!

PR #459, opened 2015-06-24
(RK-60) Decouple configuration loading, transformation, validation, and initialization

adrienthebo:


lib/r10k/deployment/config.rb
```diff
@@ -39,66 +39,16 @@ def setting(key)
...
-
-      if git_settings[:private_key]
-        R10K::Git.settings[:private_key] = git_settings[:private_key]
-      end
-
-      if git_settings[:username]
-        R10K::Git.settings[:username] = git_settings[:username]
-      end
-    end
+    initializer = R10K::Initializers::GlobalInitializer.new(@config)
```
adreyer, line 44: nit pick: This probably doesn't deserve it's own method anymore

lib/r10k/initializers.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      include R10K::Logging
+
+      def initialize(settings)
+        @settings = settings
+      end
+
+      private
+
+      def with_setting(key)
```
adreyer, line 20: What do we gain from accessing everything using this below?

lib/r10k/settings/loader.rb
```diff
@@ -0,0 +1,96 @@
...
+          raise ConfigError, "No configuration file given, no config file found in current directory, and no global config present"
+        end
+
+        begin
+          contents = ::YAML.load_file(path)
+        rescue => e
+          raise ConfigError, "Couldn't load config file: #{e.message}"
+        end
+
+        R10K::Util::SymbolizeKeys.symbolize_keys!(contents, true)
```
scotje, line 72: We discovered in pe-license that YAML::load_file can return nil and raise no errors for YAML files that are blank or contain only comments and whitespace. It looks like this would cause an exception when symbolize_keys! tries to call #keys on nil.

scotje:
Looks good overall, I added a note about some unexpected behavior from YAML::load_file that we should probably work around but that's it.

lib/r10k/settings/loader.rb
adrienthebo, line 72: Yep - in fact it's even more weird than that. I originally handled this case, but there are a slew of issues with config handling that I've broken out into tickets, and I'm trying to avoid letting this issue/PR spin into "FIX (allthethings"). This issue already affects r10k, and this call will just change where the exception is thrown. I have this issue documented as https://tickets.puppetlabs.com/browse/RK-34.

Shall we fix it now or just pull RK-34 into the next sprint?

lib/r10k/initializers.rb
adrienthebo, line 20: Syntax sugar, I can do

``` ruby
if value = @settings[:somekeytolookup]
  dothingwith(value)
```

Would you prefer that?

lib/r10k/settings/loader.rb
scotje, line 72: Since it's already tracked I think it's fine to fix it whenever. Just something I look for whenever I see YAML::load_file now :)

lib/r10k/initializers.rb
adreyer, line 20: That seems clearer? I don't care much.

lib/r10k/settings/definition.rb
```diff
@@ -0,0 +1,112 @@
...
+          @default.is_a?(Proc) ? @default.call : @default
+        end
+      end
+
+      private
+
+      # Subclasses may define additional params that are accepted at
+      # initialization; they should override this method to add any
+      # additional fields that should be respected.
+      def allowed_initialize_opts
```
adreyer, line 102: Out of curiosity why is this a map instead of a set?

adreyer:
otherwise it looks good. Tell me when you want it merged(ie if you want to fix any of the nitpicky stuff)

adrienthebo:
:+1: for merge.

PR #460, opened 2015-06-25
(RK-123) remove dependancy on facter -p

zreichert:


integration/pre-suite/01_git_config.rb
```diff
@@ -19,12 +19,9 @@
 git_manifest = ERB.new(File.read(git_manifest_template_path)).result(binding)
 
 step 'Get PE Version'
-pe_major = on(master, 'facter -p pe_major_version').stdout.rstrip
-pe_minor = on(master, 'facter -p pe_minor_version').stdout.rstrip
-pe_version = "#{pe_major}.#{pe_minor}".to_f
-
-if pe_version < 3.7
-  fail_test('This pre-suite requires PE 3.7 or above!')
+on(master, puppet('--version')) do |r|
```
cowofevil, line 22: Small quibble, the rest of the tests use "result" when referring to the Beaker result object instead of "r".

PR #461, opened 2015-06-26
Added support to actually use forge directive in Puppetfiles

hickey:
While the forge directive in a Puppet file was being read and stored, it was not being referenced when module was being downloaded. Instead the connection would default to _always_ download from https://forgeapi.puppetlabs.com/. 

Now when a module is being added to the internal data structures, it is check to see if it is an R10K::Module::Forge class and if so, the connection will be setup to use the specified forge. If the forge directive is not specified in the Puppetfile, it defaults as expected to http://forgeapi.puppetlabs.com/.

### DIFF
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -31,7 +31,7 @@ class R10K::Module::Forge < R10K::Module::Base
   # @!attribute [rw] forge
   #   @api private
   #   @return [String] The Puppet Forge where the module can be found
-  attr :forge
+  attr_accessor :forge
   
   include R10K::Logging
 
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -5,6 +5,7 @@ require 'r10k/module/metadata_file'
 
 require 'r10k/forge/module_release'
 require 'shared/puppet_forge/v3/module'
+require 'shared/puppet_forge/connection'
 
 require 'pathname'
 require 'fileutils'
@@ -27,6 +28,11 @@ class R10K::Module::Forge < R10K::Module::Base
   #   @return [PuppetForge::V3::Module] The Puppet Forge module metadata
   attr_reader :v3_module
 
+  # @!attribute [rw] forge
+  #   @api private
+  #   @return [String] The Puppet Forge where the module can be found
+  attr :forge
+  
   include R10K::Logging
 
   def initialize(title, dirname, expected_version)
@@ -36,6 +42,8 @@ class R10K::Module::Forge < R10K::Module::Base
 
     @expected_version = expected_version || current_version || :latest
     @v3_module = PuppetForge::V3::Module.new(@title)
+    
+    @forge = 'https://forgeapi.puppetlabs.com'
   end
 
   def sync(options = {})
@@ -126,6 +134,7 @@ class R10K::Module::Forge < R10K::Module::Base
       parent_path.mkpath
     end
     module_release = R10K::Forge::ModuleRelease.new(@title, expected_version)
+    module_release.conn = PuppetForge::Connection.make_connection(@forge)
     module_release.install(@path)
   end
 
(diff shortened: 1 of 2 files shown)


adrienthebo:
All things considered this is a pretty decent approach to fixing this issue. However, we were talking about how to allow users to set a custom forge and we ran into a snag - when using `r10k deploy` it makes sense to have a single Forge URL that applies to all Puppetfiles, so that all environments fetch modules from the same forge. Enabling this directive breaks that approach, which is weird. It does make more sense for `r10k puppetfile` but there's a bit of a schism in approaches here.

In addition, if we enable this setting then it'll be a (fairly big) breaking change - a lot of people have set this to something like `http://forge.puppetlabs.com`, which is not a valid Forge URL. Making this setting  functional at all will suddenly mean that existing and previously valid Puppetfiles will no longer be able to install modules from the Forge. We could add a flag somewhere to control whether the `forge` directive is respective, so it defaults to off but people can enable it. What do you think of this approach?

hickey:
I agree that the second problem is an issue, albeit a self imposed one. If people should be using forgeapi.puppetlabs.com in their Puppetfiles instead of forge.puppetlabs.com, then it would really be good for Puppetlabs to be promoting the use of forgeapi.puppetlabs.com. Hint: a good place to start is to update the r10k documentation about this issue and not continue to ignore and hope it will resolve itself. I would also coordinate with Tim on getting librarian-puppet updated so that there is not conflicting information. 

Luckily this is an easy fix for r10k. Since it is using the v3 api, anytime it sees a forge statement using the old address, internally change it to the new address and print a warning to prompt people to update their Puppetfiles. Do what you can to entice people to follow good behavior. 

I am not sure I fully understand the issues you see with your first point. If the intent is to allow someone to override the forge statements in Puppetfiles so that an alternative or internal Forge server, then I am not sure I see the use case here. Besides are these people not controlling their own Puppetfiles? The only time I can see something like this to be useful is if one is bringing up a new Forge server and wanted to test pulling modules against the new Forge server. Seems that this would be a command line arg that would allow one to force all forge access through a specified server. 

What I really don't understand is the statement, "Enabling this directive breaks that approach, which is weird." Because this precisely what I am doing now. I have an internal Forge server that I point the Puppetfiles to and pull the modules from it (and also proxy to forgeapi.puppetlabs.com) instead of forgeapi.puppetlabs.com. This is exactly how I found this issue and why I made the code changes to allow the use of the internal Forge server. So I must be missing something.....

Ideally, one of the things that I would rather see (which may fix a couple of the issues you are anticipating) is to allow multiple forge statements in a Puppetfile. Each Forge would be tried in order until it finds the requested module. Seems like a pretty straight forward implementation. Again, this would be something I would coordinate with Tim with respect to librarian-puppet in case there is an edge case that might break librarian and allow him to be proactive before librarian breaks for people.

adreyer:
In resolving RK-70 it was decided to configure which module repository to use via the config file instead of the Puppetfile. This decision was made specifically to avoid the issues and resulting complex work arounds librarian-puppet has implemented to ensure Puppetfiles are compatible across versions. The 'forge' directive existed in r10k to make it easier for the majority of users who don't care about it to share use librarian-puppet Puppetfiles. For the immediate future that is probably all it will be since honoring it would break most peoples installations or result in confusing errors when it's misconfigured.

We do need to do a better job documenting that(https://tickets.puppetlabs.com/browse/RK-138). I'm sorry this wasn't documented sooner but it's a fairly recent decision made after seeing the problems the forge directive caused librarian.

adrienthebo:
I'm not sure this has been fully discussed - @hickey you are aware that the Forge URL does have a configuration option for the Forge URL in r10k (https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#baseurl) ? This hasn't yet been released yet but will be available in r10k 2.1.0.

adrienthebo:
r10k 2.1.0 has been released and supports using a custom Puppet Forge (https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#baseurl). Per @adreyer's comment about RK-138 we're unlikely to implement the `forge` directive since it has strange interactions with having a per-environment Forge set via the Puppetfile. We can continue the discussion about that on the ticket; in the mean time I'm closing this ticket.

PR #462, opened 2015-06-26
(CODEMGMT-224) Update tests for SG compatibility

zreichert:
*Updated all calls to 'r10k' to use the fully qualified path
*Updated module install to use a dynamic module path


cowofevil:
The "test_run_scripts" reference 3.8.1 and should be updated to point to 4.0.

cowofevil:
@zreichert Also, the "setup" scripts need to be updated as well. All tests are passing so if you address comments I'll merge.

PR #464, opened 2015-07-06
Maintenance

zreichert:


.gitignore
```diff
@@ -4,3 +4,6 @@ Gemfile.lock
 .bundle
 bundle
 coverage
+.idea/
```
adrienthebo, line 7: I would rather we not put a bunch of IDE specific ignores in the project .gitignore; you can use a global gitignore (https://help.github.com/articles/ignoring-files/) to ignore files specific to your IDE of choice, or have local ignores in `.git/info/exclude` to ignore files on a per-repository basis without modifying the project .gitignore.

adrienthebo:
@zreichert ping per the above comment?

adrienthebo:
@cowofevil @zreichert so what's going on with this PR?

PR #471, opened 2015-07-10
Update test run scripts to use 2015.2

zreichert:


adrienthebo:
There's a huge amount of duplication in these files, is it worth trying to clean these up or are you okay with having these duplicate changes?

PR #473, opened 2015-07-13
(RK-146) Use settings class for normalization

adrienthebo:
The settings classes added in RK-59 weren't being used which meant that
we weren't getting input normalization; this commit resolves this by
using the settings layer when loading a config in
R10K::Deployment::Config. This commit also tears out some internal
string/symbol normalization that is handled by the config loading
process and thus is no longer needed.


andersonmills:
:+1: I believe in this. Yay! I'm so happy. Just waiting for Travis to be able to merge.

PR #475, opened 2015-07-14
(RK-142) Add machine readable format options to "deploy display"

scotje:
This may merit some test coverage at this point...

### DIFF
--- a/lib/r10k/cli/deploy.rb
+++ b/lib/r10k/cli/deploy.rb
@@ -88,6 +88,7 @@ try to deploy the given module names in all environments.
 
           flag :p, :puppetfile, 'Display Puppetfile modules'
           flag nil, :detail, 'Display detailed information'
+          required nil, :format, 'Display output in a specific format. Valid values: json, yaml. Default: yaml'
 
           runner R10K::Action::CriRunner.wrap(R10K::Action::Deploy::Display)
         end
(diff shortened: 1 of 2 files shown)


scotje:
Thinking about just removing the "legacy" output format entirely and making "yaml" the default. Only question is whether anyone has integrated against the legacy pseudo-yaml output already and if it's worth preserving for that case.

lib/r10k/cli/deploy.rb
```diff
@@ -88,6 +88,7 @@ def self.command
 
           flag :p, :puppetfile, 'Display Puppetfile modules'
           flag nil, :detail, 'Display detailed information'
+          optional nil, :format, 'Display output in a specific format. Valid values: json, yaml', :argument => :required
```
adrienthebo: This can be `required` instead of `optional`, which drops the need to have `:argument => :required`

scotje: Ah ok, yeah I found the documentation around that part of Cri a little thin....

adrienthebo:
I think we should drop the legacy format; it's really crufty and the YAML/JSON outputters give people a real, parseable format.

scotje:
Pushed a new commit, if that looks good I can squash it so we can merge.

adrienthebo:
:+1:, squash it and I'll merge it.

scotje:
squashed

PR #479, opened 2015-07-27
cli/puppetfile: add moduledir and puppetfile

rowleyaj:
These options appear to already be handled by the action.
Surfacing them at the command level so they can be used

### DIFF
--- a/lib/r10k/cli/puppetfile.rb
+++ b/lib/r10k/cli/puppetfile.rb
@@ -31,8 +31,8 @@ Puppetfile (http://bombasticmonkey.com/librarian-puppet/).
           usage   'install'
           summary 'Install all modules from a Puppetfile'
 
-          # @todo add --moduledir option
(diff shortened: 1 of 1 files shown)


rowleyaj:
@adrienthebo is anything else required from me to pose this for review? It looked like it was a wanted change as there was todo items in the code for it. I think the deprecated part was pulling this settings from env vars, but I may have misunderstood and the functionality as a whole was deprecated.

adrienthebo:
@rowleyaj sorry for the delay - this looks good, and I should be able to merge this as-is and just got bogged down. With luck I'll be able to get to it shortly. Again, sorry for the delay and thanks for both the contribution and your patience!

adrienthebo:
Merged into master in 221c7df; this should be released in 2.1.0. Thanks for the contribution!

rowleyaj:
Thanks @adrienthebo :)

PR #480, opened 2015-07-30
Update the quickstart for Puppet 4

branan:
This brings the quickstart up-to-date for PC1/puppet 4.

It tosses the documentation for puppet 3 entirely - I'd be happy to bring parts of that back as a separate "getting r10k working on puppet 3" document, but I think the quickstart needs to stay focused and simple. There are enough differences between the versions that trying to talk about them both at each step made the guide hard to follow.


doc/dynamic-environments/quickstart.mkd
```diff
@@ -3,22 +3,21 @@
 This intent of this document is to serve as a basic guide for getting started with r10k and a fresh Puppet installation, including the following:
 
 * Installing Puppet and its dependencies
```
adrienthebo, line 5: This document implicitly assumes that we're installing Puppet 4, could you call that out here?

branan, line 5: How about changing two lines up to "getting started with r10k and the latest Puppet version", so we don't have to keep bumping a number?

adrienthebo, line 5: I'm fine with specifying "Puppet 4 and greater", but it's worthwhile saying that this won't work at all for Puppet 3.x.

adrienthebo, line 5: Branannnnnnn

hey

hey

BRANAN

Wanna fix this?

adrienthebo, line 5: @branan please don't make me fix this one up by hand :(

doc/dynamic-environments/quickstart.mkd
```diff
@@ -3,22 +3,23 @@
...
-* Installing Hiera and its dependencies.
 * Installing r10k and its dependencies
 * Configuring all components to support r10k
 * Configuring your git repository and initial files
 
+This document is based around Puppet 4+, specifically using the Puppet Collection repositories provided by Puppet Labs.
```
branan, line 10: How's this look?

PR #482, opened 2015-08-12
(RK-156) Check for pe-license lib via `require`, improve module release logging

adrienthebo:
If the pe_license module is installed into the ruby library path via something like rpm or deb and is not installed into the rubygems library path, it will not be locatable via `Gem::Specification.find_all_by_name`.  This commit switches the library detection to use the r10k feature system, which handles detection by actually requiring the library. While requiring the library as part of detection is a fairly blunt approach and has side effects, it's a more reliable detection method (and if we're checking to see if the library is present, we're going to want to require it anyways).

This also adds improved logging when PE only modules can't be downloaded.

### DIFF
--- a/lib/r10k/features.rb
+++ b/lib/r10k/features.rb
@@ -16,3 +16,5 @@ end
 R10K::Features.add(:shellgit) { R10K::Util::Commands.which('git') }
 
 R10K::Features.add(:rugged, :libraries => 'rugged')
+
+R10K::Features.add(:pe_license, :libraries => 'pe_license')
--- a/lib/r10k/util/license.rb
+++ b/lib/r10k/util/license.rb
@@ -1,19 +1,23 @@
 require 'r10k/errors'
-require 'rubygems'
+require 'r10k/features'
 
 module R10K
   module Util
     module License
+      extend R10K::Logging
+
       def self.load
-        if Gem::Specification::find_all_by_name('pe-license').any?
+        if R10K::Features.available?(:pe_license)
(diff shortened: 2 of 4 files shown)


adrienthebo:
Verified that this works by building a new gem and installing it on top of the PE 2015.2.0 gem; the installation proceeds as normal.

scotje:
Seems fine. Is the default behavior of the PE packages to install pe-license to somewhere on the load path that isn't one of the gem paths?

Also, does that additional 403 behavior in the vendored puppet_forge code need to be ported to puppet_forge standalone?

adrienthebo:
```
[vagrant@master ~]$ find /opt -name pe_license.rb 2>&-
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/pe_license.rb
```

```
[vagrant@master ~]$ find /opt -name r10k.rb 2>&-
/opt/puppetlabs/puppet/lib/ruby/gems/2.1.0/gems/r10k-2.0.2/lib/r10k.rb
```

Since it's installed outside of lib/ruby/gems, it's invisible to Rubygems but can still be required.

andersonmills:
@scotje I mentioned that additional port to puppet_forge to @austb

andersonmills:
Tested this on PE 2015.2.1 candidate by building this as a gem and then installing on top of PE. With a valid license file, it works, and PE only modules install.

andersonmills:
:+1: I believe. @adrienthebo feel free to merge.

PR #487, opened 2015-08-27
(CODEMGMT-307) Update r10k for puppet_forge gem

andersonmills:
Before this commit, r10k was still in the form that used library code
from a shared library directory included in the r10k code base. With
this commit, r10k is updated to use the newer functions provided by the
puppet_forge gem including some new error handling, because errors were
changed and moved around.

(DO NOT MERGE) until
- ~~puppet_forge gem v2.1.0 is released~~ RELEASED


lib/r10k/module/base.rb
```diff
@@ -36,10 +36,11 @@ class R10K::Module::Base
   # @param dirname [String]
   # @param args [Array]
   def initialize(title, dirname, args)
-    @title   = title
+    normalized_title = PuppetForge::V3.normalize_name(title)
+    @title   = normalized_title
     @dirname = dirname
     @args    = args
-    @owner, @name = parse_title(title)
+    @owner, @name = parse_title(normalized_title)
```
scotje: Any reason this couldn't be:

``` ruby
@title = PuppetForge::V3.normalize_name(title)
...
@owner, @name = parse_title(@title)
```

scotje:
Overall looks good, there seems to be a little inconsistency about whether or not we `require 'puppet_forge'` stuff explicitly in each file that uses it...

andersonmills:
@scotje Cleaned up puppet_forge requires to the files they're actually being used in. Cleaned up module/base title in initialize as you suggested. Looking for a +1.

spec/unit/forge/module_release_spec.rb
```diff
@@ -5,6 +5,8 @@
 describe R10K::Forge::ModuleRelease do
   subject { described_class.new('branan-eight_hundred', '8.0.0') }
 
+  let(:forge_release_class) { PuppetForge::V3::Release }
```
scotje, line 9: this spec file should probably also require puppet_forge

scotje:
One more missing require in the tests but +1 when that is fixed.

andersonmills:
@scotje Turned out there was an extra on in metadata_file_spec.rb and a missing one in runner_spec.rb, too. All of those fixed now.

scotje:
:fireworks: 

It would be nice if there was some sort of pedantic require warnings you could turn on. :)

:+1:

PR #494, opened 2015-09-10
(maint) Remove use of def_delegators for module_release

adrienthebo:
The extraction of the puppet_forge into the v2 gem changed around some
APIs, and seems to break under certain conditions. This commit removes
the use of the Forwardable module and calls the appropriate methods on
the PuppetForge::V3::Release class.


andersonmills:
:+1:

Issue #67, opened 2013-09-18
Some level of reporting


zachfi:
I would like some ability to get reporting information out of r10k.
- Machine consumable (json, etc)
- How many branches were deployed
- How much time was taken in total

This should at least be able to get us something intelligent to send to graphite so that we know the state of the deploy system.

jlambert121:
I like to keep my modules up to date, but don't want them to necessarily want to implement the #51 solution of setting module versions to 'latest'.  It would be great if the reports could include modules that were no longer current so they could be evaluated for upgrade.

PR #498, opened 2015-09-14
(CODEMGMT-345) Make r10k write deployed code signature to file.

scotje:


adrienthebo:
Feedback from IRC:

```
[11:52]      Elyscape I'd be cool with that but .r10k_deploy might make more sense
[11:54]      Elyscape deploy.json in theory might conflict with something else
[11:55]      Elyscape also, more importantly, it's standard practice to contain the app name in the filename for that sorta thing
[11:55]      Elyscape and having it be a .dotfile means it won't be visible by default, as that could potentially confuse people
```

adrienthebo:
Perhaps `.r10k-deploy.json` ?

scotje:
Ha, yeah it was originally .r10k-info.json but I changed it to match the ticket. I'll switch it to .r10k-deploy.json, that seems to make the most sense.

scotje:
Updated.

lib/r10k/action/deploy/environment.rb
```diff
@@ -86,6 +93,16 @@ def visit_module(mod)
           logger.info "Deploying module #{mod.path}"
           mod.sync
         end
+
+        def write_environment_info!(environment)
+          File.open("#{environment.path}/.r10k-deploy.json", 'w') do |f|
+            deploy_info = environment.info.merge({
+              :deployed_at => Time.new,
```
adrienthebo: `:deployed_at` indicates that this was the start time (at least to me) but this would run at the end of the environment deploy. In addition I have these wild fantasies of including timing information (for #67). This a bit bike-sheddy, but could we use a name like `:finished_at` or something like that?

scotje:
Updated, now with `started_at` and `finished_at` timestamps. Do you think we need sub-second resolution on the timings?

adrienthebo:
Second level resolution is totally fine. Thanks for making that change!

PR #500, opened 2015-09-15
(CODEMGMT-339) Allow cachedir to be overridden with a CLI option.

scotje:


lib/r10k/deployment/config.rb
```diff
@@ -30,7 +31,8 @@ def setting(key)
   def load_config
     loader = R10K::Settings::Loader.new
     hash = loader.read(@configfile)
-    @config = R10K::Settings.global_settings.evaluate(hash)
+    with_overrides = hash.merge(@overrides)
```
adrienthebo: I would like for us to log when config values are overridden; there's the block form of `Hash#merge` that could be used to log when overrides replace values - perhaps like so?

``` ruby
with_overrides = hash.merge(@overrides) do |key, oldval, newval|
  logger.debug2 "Overriding config setting #{key} with old value #{oldval} to new value #{newval}"
end
```

I'm thinking that having this logged may make it easier to debug support requests.

Thoughts?

scotje: That seems like a good idea. I'll add it.

scotje:
Updated with logging of overrides.

lib/r10k/cli/deploy.rb
```diff
@@ -23,6 +23,8 @@ def self.command
 
         required :c, :config, 'Specify a configuration file'
 
+        required nil, :cachedir, 'Specify a cachedir, overriding the value in config'
```
adrienthebo, line 26: Adding the `cachedir` command line flag here means that this will be a one off for the `r10k deploy environment` command, meaning that it won't work with `r10k deploy module` or `r10k puppetfile`. I know that this feature is being implemented for a specific use case but I dislike having one offs or inconsistencies unless there's a good reason for them. Does it make sense to keep this option limited to the single command?

scotje, line 26: Probably makes sense to be available more globally...

scotje, line 26: This location makes it available to all the "deploy" subcommands if they are updated to accept and pass it through, is there a place we can put the acceptance and override functionality that will just apply to everything without updating it per action?

adrienthebo, line 26: https://tickets.puppetlabs.com/browse/RK-120 covers making `--config` global, and all that really entails is moving the settings loading/initialization from the `Action::Deploy::*` classes to https://github.com/puppetlabs/r10k/blob/master/lib/r10k/action/runner.rb#L39-L41, and then passing the loaded configuration to the called action classes.

What we could do is merge this as-is so that we can continue work on the things that depend on this PR, and once this is in we can also work on RK-120. Does that work for you?

scotje, line 26: That seems reasonable. It wouldn't take much work to update the other deploy subcommands to honor the option as well for now if you think that would be helpful.

adrienthebo, line 26: If we're focusing on getting the required functionality now and expanding things later, let's not worry about `r10k deploy module` right now.

PR #501, opened 2015-09-16
Fix "git environment guide" internal link

apazga:
Bad URL (now): https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/doc/dynamic-environments/git-environments.mkd
Good URL (proposed): https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/git-environments.mkd
(these are full URLs, fix proposed using internal links)


adrienthebo:
Thanks!

PR #502, opened 2015-09-22
Mention potential issues with rugged gem.

dpash:
There's a number of issues that can trip people over when using Rugged. This should document a number of potential problems that people might discover.

### DIFF
--- a/doc/git/providers.mkd
+++ b/doc/git/providers.mkd
@@ -50,6 +50,43 @@ git:
   username: 'git'
 ```
 
+#### Supported transports with Rugged
+
+Rugged compiles libgit2 and and the Ruby bindings when the gem is installed. You
+may need libraries installed before you install the gem to use certain protocols
+to access git remote repositories.
+
+For ssh support, you need to have libssh2 installed (along with the relevant dev
+package/headers) before you install the Rugged gem.
+
+For https support on Linux, you need to have OpenSSL installed (along with the
+relevant dev package/headers) before you install the Rugged gem. OS X and
+Windows support should automatically include https support.
+
+You can check whether https or ssh support is included in your Rugged
+installation by using the following in irb and making sure the required feature
+is listed:
+
+```ruby
+irb(main):001:0> require('rugged')
+=> true
+irb(main):002:0> Rugged.features
+=> [:threads, :https, :ssh]
+irb(main):003:0> 
+```
+You will require the ':https' or ':ssh' features to use the respective protocols
+in your Puppetfile module references or in r10k.yaml. R10K 2.0.0 and later will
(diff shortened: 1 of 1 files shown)


doc/git/providers.mkd
```diff
@@ -50,6 +50,24 @@ git:
...
+
+For https support on Linux, you need to have OpenSSL installed (along with the relevant dev package/headers) before you install the Rugged gem. OS X and Windows support should automatically include https support.
+
+You can check which features your Rugged supports by using the following in irb:
+
+```ruby
+irb(main):001:0> require('rugged')
+=> true
+irb(main):002:0> Rugged.features
+=> [:threads, :https, :ssh]
```
rnelson0, line 74: Which features are required?

doc/git/providers.mkd
```diff
@@ -50,6 +50,24 @@ git:
...
+irb(main):003:0> 
+```
+
+libssh on Debian and Ubuntu is compiled against libgcrypto instead of OpenSSL [due to licensing reasons](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=668271), and unfortunately libgcrypto does not support a number of required operations, including reading from a private key file. You will need to either use shellgit or recompile your own libssh2-1 package to use OpenSSL on these distributions.
```
adrienthebo: libssh2, libssh is a different library.

doc/git/providers.mkd
adrienthebo, line 74: r10k 2.0.0 will automatically check for the :https and :ssh features and will warn if libgit2/rugged was compiled without them: https://github.com/puppetlabs/r10k/commit/94bfeebc1e8f4a723187c60c78ab5fcde2e1bd39

dpash, line 74: Is it worth mentioning that?

rnelson0:
Looks good to me now!

doc/git/providers.mkd
adrienthebo, line 74: I would like for us to document that r10k 2.0.0 will probe for those features, but in r10k 1.5.1 users can use irb to make that check themselves.

adrienthebo:
I have one last thing that I would like to have addressed, then I'm happy with this. Before it gets merged, could you rebase this into a single commit with `git rebase master` and update the pull request with `git push -f` ?

dpash:
Hopefully this commit should be one squashed commit. The changes between this and previous commits is that we mention the warnings in r10k 2.0.0 and also wraps lines at 80 characters to match the rest of the document. Seems github's web editor doesn't do that for you. :)

adrienthebo:
@dpash looks great, thanks for the contribution!

PR #507, opened 2015-10-01
Fix documentation so local_mod/local_module name is consistent

mmckinst:
Sometimes it is called 'local_mod', other times it is called 'local_module'.

Now it will be called 'local_module' all of the time.


adrienthebo:
Thanks!

PR #509, opened 2015-10-05
(RK-80) properly supply credentials to HTTPS connections

adrienthebo:
Previously, r10k would always supply SSH key based authentication credentials when authentication was requested, which would fail poorly when a HTTPS based URL required authentication. This refactors and extracts the logic for supplying credentials to Rugged and adds support for user/password based authentication.

### DIFF
--- a/lib/r10k/git/rugged/base_repository.rb
+++ b/lib/r10k/git/rugged/base_repository.rb
@@ -62,7 +62,13 @@ class R10K::Git::Rugged::BaseRepository
   # @return [Proc]
   def credentials
     Proc.new do |url, username_from_url, allowed_types|
-      get_ssh_credentials(url, username_from_url)
+      if allowed_types.include?(:ssh_key)
+        get_ssh_key_credentials(url, username_from_url)
+      elsif allowed_types.include?(:plaintext)
+        get_plaintext_credentials(url, username_from_url)
+      else
+        get_default_credentials(url, username_from_url)
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/r10k/git/rugged/credentials.rb
```diff
@@ -0,0 +1,69 @@
...
+    @repository = repository
+  end
+
+  def call(url, username_from_url, allowed_types)
+    if allowed_types.include?(:ssh_key)
+      get_ssh_key_credentials(url, username_from_url)
+    elsif allowed_types.include?(:plaintext)
+      get_plaintext_credentials(url, username_from_url)
+    else
+      get_default_credentials(url, username_from_url)
```
adreyer, line 23: where is this coming from?

spec/unit/git/rugged/credentials_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      expect(subject).to receive(:get_ssh_key_credentials).with("https://tessier-ashpool.freeside/repo.git", nil).and_return :sshkey
+      expect(subject.call("https://tessier-ashpool.freeside/repo.git", nil, [:ssh_key])).to eq :sshkey
+    end
+
+    it "creates user/password credentials for the default allowed type" do
+      expect(subject).to receive(:get_plaintext_credentials).with("https://tessier-ashpool.freeside/repo.git", nil).and_return :userpassword
+      expect(subject.call("https://tessier-ashpool.freeside/repo.git", nil, [:plaintext])).to eq :userpassword
+    end
+
+    it "creates default credentials when no other types are allowed" do
```
adreyer, line 70: perhaps these tests are mocking too much out since they hid the lack of this method.

lib/r10k/git/rugged/base_repository.rb
```diff
@@ -1,10 +1,14 @@
 require 'r10k/git/rugged'
+require 'r10k/git/rugged/credentials'
 require 'r10k/logging'
 
 class R10K::Git::Rugged::BaseRepository
 
   include R10K::Logging
 
+  #  @return [Pathname] The path to this repository.
+  attr_reader :path
```
adreyer, line 11: nitpick: It's slightly confusing to have @path defined here but set up in the initialization of subclasses. Putting an initialize here seems like spaghetti though since it won't do anything. Perhaps a comment reminding us to set @path in any new subclasses?

spec/unit/git/rugged/credentials_spec.rb
adrienthebo, line 70: Agree, fixing that now.

lib/r10k/git/rugged/base_repository.rb
adrienthebo, line 11: Will do.

PR #510, opened 2015-10-06
(refactor) Remove boilerplate and deduplicate code in actions

adrienthebo:
There's a lot of boilerplate that's crept into the action classes over time; this commit adds a base class that handles a number of shared concerns. This pull request also adds in tests in areas that were lacking to validate that the refactoring had no adverse affects.

### DIFF
--- a/lib/r10k/action/deploy/display.rb
+++ b/lib/r10k/action/deploy/display.rb
@@ -1,30 +1,10 @@
-require 'r10k/util/setopts'
 require 'r10k/deployment'
-require 'r10k/logging'
+require 'r10k/action/base'
 
 module R10K
   module Action
     module Deploy
-      class Display
-
-        include R10K::Util::Setopts
-        include R10K::Logging
-
-        def initialize(opts, argv)
-          @opts = opts
-          @argv = argv
-          setopts(opts, {
-            :config     => :self,
-            :puppetfile => :self,
-            :detail     => :self,
-            :format     => :self,
-            :fetch      => :self,
-            :trace      => :self
-          })
-
-          @level  = 4
-          @indent = 0
-        end
+      class Display < R10K::Action::Base
 
         def call
           deployment = R10K::Deployment.load_config(@config)
@@ -91,6 +71,10 @@ module R10K
           else
             mod.title
           end
+        end
+
+        def allowed_initialize_opts
+          super.merge(puppetfile: :self, detail: :self, format: :self, fetch: :self)
         end
       end
     end
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -2,29 +2,20 @@ require 'r10k/util/setopts'
 require 'r10k/deployment'
 require 'r10k/logging'
 require 'r10k/action/visitor'
+require 'r10k/action/base'
 require 'r10k/deployment/write_lock'
 require 'json'
 
 module R10K
   module Action
     module Deploy
-      class Environment
+      class Environment < R10K::Action::Base
 
-        include R10K::Util::Setopts
-        include R10K::Logging
         include R10K::Deployment::WriteLock
 
         def initialize(opts, argv)
-          @opts = opts
-          @argv = argv.map { |arg| arg.gsub(/\W/,'_') }
-          setopts(opts, {
-            :config     => :self,
-            :cachedir   => :self,
-            :puppetfile => :self,
-            :purge      => :self,
-            :trace      => :self
-          })
-
+          super
+          @argv = @argv.map { |arg| arg.gsub(/\W/,'_') }
           @purge = true
         end
 
@@ -111,6 +102,10 @@ module R10K
 
             f.puts(JSON.pretty_generate(deploy_info))
           end
+        end
+
+        def allowed_initialize_opts
+          super.merge(puppetfile: :self, cachedir: :self)
         end
       end
     end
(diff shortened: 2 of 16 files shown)


lib/r10k/action/deploy/environment.rb
```diff
@@ -112,6 +103,10 @@ def write_environment_info!(environment, started_at)
             f.puts(JSON.pretty_generate(deploy_info))
           end
         end
+
+        def allowed_initialize_opts
+          super.merge(puppetfile: true, cachedir: true)
```
adreyer: :purge?

lib/r10k/action/puppetfile/check.rb
```diff
@@ -34,6 +18,12 @@ def call
             false
           end
         end
+
+        private
+
+        def allowed_initialize_opts
+          super.merge(root: :self, puppetfile: :self, moduledir: :self)
```
adreyer, line 25: The previous classes had these converted to booleans is this intentional?

lib/r10k/action/puppetfile/install.rb
```diff
@@ -47,6 +29,10 @@ def visit_module(mod)
           logger.info "Updating module #{mod.path}"
           mod.sync
         end
+
+        def allowed_initialize_opts
+          super.merge(root: :self, puppetfile: :self, moduledir: :self)
```
adreyer, line 34: :self not true?

lib/r10k/action/puppetfile/purge.rb
```diff
@@ -31,6 +16,12 @@ def call
           logger.error R10K::Errors::Formatting.format_exception(e, @trace)
           false
         end
+
+        private
+
+        def allowed_initialize_opts
+          super.merge(root: :self, puppetfile: :self, moduledir: :self)
```
adreyer, line 23: I guess there are more with :self

lib/r10k/action/deploy/display.rb
```diff
@@ -1,30 +1,10 @@
...
-          setopts(opts, {
-            :config     => :self,
-            :puppetfile => :self,
-            :detail     => :self,
-            :format     => :self,
-            :fetch      => :self,
-            :trace      => :self
-          })
-
-          @level  = 4
```
adreyer, line 25: it looks like didn't get moved anywhere. is it important?

lib/r10k/action/deploy/environment.rb
```diff
@@ -2,29 +2,20 @@
...
-          setopts(opts, {
-            :config     => :self,
-            :cachedir   => :self,
-            :puppetfile => :self,
-            :purge      => :self,
-            :trace      => :self
-          })
-
+          super
+          @argv = @argv.map { |arg| arg.gsub(/\W/,'_') }
```
adreyer, line 18: What is this for? Do the options get stripped off before we do this?

lib/r10k/action/deploy/module.rb
```diff
@@ -1,30 +1,15 @@
...
-        def initialize(opts, argv)
-          @opts = opts
-          @argv = argv
-          setopts(opts, {
-            :config      => :self,
-            :environment => nil,
-            :trace       => :self
-          })
-
-          @purge = true
```
adreyer, line 25: Was this important or is it handled in the default config anyway?

lib/r10k/action/deploy/display.rb
scotje, line 25: I think those were used for the old way the output was formatted and I probably should have removed them when I converted to yaml/json output.

lib/r10k/action/deploy/environment.rb
adrienthebo: `:purge` is a bit strange in that I meant to make it a tunable and the old code supported it, but I never got around to passing through the option from the CLI to the Action. Good catch, I'll add this.

lib/r10k/action/puppetfile/purge.rb
adrienthebo, line 23: I'll centralize around `:self`, and at some point I should yank :true.

In the future I should think very seriously about dropping 1.9 support so that we can use Ruby keyword args instead of this workaround.

lib/r10k/action/deploy/environment.rb
adrienthebo, line 18: Options are handled through `super`; this transformation needs to be done so that normalized environment names are handled consistently when specified on the CLI and generated from `R10K::Deployment`. It was originally a community contributed fix and isn't the most elegant thing, but it works and I don't want to tackle this cleanup at the moment.

lib/r10k/action/deploy/module.rb
adrienthebo, line 25: `:config` and `:trace` are handled by the parent `#initialize`, which we can rely on directly.

lib/r10k/action/puppetfile/purge.rb
adreyer, line 23: make a 3.0 ticket?

lib/r10k/action/deploy/module.rb
adreyer, line 25: I meant the @purge true

adrienthebo, line 25: I believe that `@purge` was more boilerplate that I copy/pasted; it's not used in this file.

adrienthebo:
Converted the setopts `true` instances to `:self`, and I'll make a ticket for removing `true` as a setopts value in 3.0.

andersonmills:
:+1: From me.

looks good

PR #514, opened 2015-10-13
(RK-21) Treat missing deployed environments as an error

adrienthebo:
This pull request adds checking to r10k to ensure that if environments are supplied to the `r10k deploy environment` command, r10k ensures they're present, logs an error if any are missing, and exits with a non-zero exit code when environments are missing.


adreyer:
waitting for conflict resolution before reviewing

adrienthebo:
Rebased and ready to go.

PR #515, opened 2015-10-14
(RK-175) [rugged] Don't check out unresolvable refs

adrienthebo:
The rugged working repository always tried to resolve a Git ref to a SHA
when checking out a ref, but while the shellgit implementation could
rely on the underlying git binary to report meaningful errors the rugged
implementation had no such fallback check. This could lead to r10k
trying to check out 'nil' when an unresolvable ref was given, which
would raise an ArgumentError. This commit adds an explicit check for
unresolvable refs and raises a meaningful error instead of an
ArgumentError.

### DIFF
--- a/lib/r10k/git/rugged/working_repository.rb
+++ b/lib/r10k/git/rugged/working_repository.rb
@@ -59,8 +59,13 @@ class R10K::Git::Rugged::WorkingRepository < R10K::Git::Rugged::BaseRepository
   # @param ref [String] The git reference to check out
   # @return [void]
   def checkout(ref)
-    logger.debug1 { "Checking out ref '#{ref}' at #{@path}" }
     sha = resolve(ref)
+
+    if sha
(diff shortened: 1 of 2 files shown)


lib/r10k/git/rugged/working_repository.rb
```diff
@@ -59,9 +59,14 @@ def clone(remote, opts = {})
   # @param ref [String] The git reference to check out
   # @return [void]
   def checkout(ref)
-    logger.debug1 { "Checking out ref '#{ref}' at #{@path}" }
     sha = resolve(ref)
 
+    if sha
+      logger.debug2 { "Checking out ref '#{ref}' (resolved to SHA '#{sha}') in repository #{@path}" }
```
adreyer, line 65: Was this intentionally bumped down. It matches the priority I noticed earlier for fetching so I guess that makes sense.

PR #516, opened 2015-10-15
More specific failure for argument error

patrobinson:
If someone were to provide more than two arguments in a Puppetfile, not that I would ever do that :eyes:, then `r10k puppetfile check` wouldn't output a very helpful error message.

Make this error a little bit more informative by wrapping it.


adrienthebo:
Thanks for this contribution! One small thing before I merge - we generally try to make sure that every commit is able to pass tests, so if we need to run a `git bisect` or something like that we can rely on all existing tests to pass so we can chase down regressions. Could you squash these commits together and force push the changes? If you can do that then I can get this merged right after. Thanks!

patrobinson:
@adrienthebo done.

adrienthebo:
Thanks for the contribution!

PR #517, opened 2015-10-16
(RK-120) Allow all commands/actions to use --config

adrienthebo:
This pull request moves config loading/reading to the application startup so that all r10k subcommands can specify a config.

### DIFF
--- a/spec/unit/action/runner_spec.rb
+++ b/spec/unit/action/runner_spec.rb
@@ -1,7 +1,8 @@
 require 'spec_helper'
-require 'r10k/action/runner'
+require 'r10k/action/base'
 require 'puppet_forge/connection'
 
+require 'r10k/action/runner'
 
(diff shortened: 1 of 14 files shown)


spec/unit/action/runner_spec.rb
```diff
@@ -50,6 +52,11 @@ def call
       runner.call
     end
 
+    it "configures forge authoriation" do
```
scotje: s/authoriation/authorization

scotje:
:+1:

spec/unit/action/runner_spec.rb
adrienthebo: No idea how that typo crept in...

adrienthebo:
Resolved merged conflicts against master and fixed a typo in a spec description.

PR #518, opened 2015-10-19
(RK-174) Always sync alternates file when touching thin repos

adrienthebo:
When r10k switches the Git cachedir, it will try to associate existing thin repositories with a new cached repo, but doesn't properly update the alternates file, which means that it will ignore the new cache directory. This can lead to situations where the repos in the new cache directory are updated but existing thin repositories are referencing Git repositories that were not updated, and are missing objects that only exist in the new cached repo.

This commit changes thin repository creation to check and update the alternates file when needed so that cachedir updates are automatically reflected in referenced object dbs.

### DIFF
--- a/lib/r10k/git/alternates.rb
+++ b/lib/r10k/git/alternates.rb
@@ -12,21 +12,34 @@ class R10K::Git::Alternates
   # @param git_dir [Pathname] The path to the git repository
   def initialize(git_dir)
     @file = git_dir + File.join('objects', 'info', 'alternates')
+    @entries = []
   end
 
-  def to_a
-    read()
-  end
-
-  def <<(path)
+  def add(path)
     write(to_a << path)
+  end
+  alias << add
+
+  # Conditionally add path to the alternates file
+  #
+  # @param path [String] The file path to add to the file if not already present
+  # @return [true, false] If the entry was added.
+  def add?(path)
+    paths = read()
+
+    add_entry = !paths.include?(path)
+
+    if add_entry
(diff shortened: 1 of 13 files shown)


adrienthebo:
This should probably include tests for some of the introduced changes, but they would be very much testing the implementation. I might write some integration (spec/integration rather than beaker) level tests for this.

scotje:
Seems reasonable to me. Since we are appending new alternates, does that mean it will potentially be looking across multiple cachedirs to find objects?

adrienthebo:
@scotje yes, but that's pretty harmless and might actually make things EVEN FASTER!!!111one111 It'll be a constant time (mostly, filesystem checks will be constant but packfiles might slow things down) to look up each location so it's cheap to do the check, and if we allow multiple alternates files then it'll allow r10k to work well in environments such as vagrant, where someone is running r10k on the host system and inside of the vagrant VM.

adrienthebo:
Shouldn't have been so hasty to respond - I can yank that functionality, but since it's harmless and has a few nice side effects it would be nice to leave in.

scotje:
My only concern I guess is what might happen in the case that one (or all) of those cachedirs goes away, if the old non-existent ones are retained in the alternates file would that cause any problems?

adrienthebo:
I've got a git repository that's associated with about 9 cached repos that I deleted; zero impact. If the cache directory doesn't exist it's skipped entirely.

scotje:
Cool, :+1: from me then.

PR #521, opened 2015-10-21
(maint) Update log regex for missing r10k config

adrienthebo:


integration/tests/basic_functionality/negative/neg_deploy_with_missing_r10k_yaml.rb
```diff
@@ -9,7 +9,7 @@
 r10k_fqp = get_r10k_fqp(master)
 
 #Verification
-error_message_regex = /No configuration file given, no config file found in current directory, and no global config present/
+error_message_regex = /the supplied configuration does not define the 'sources' key/
```
zreichert, line 12: @adrienthebo Is there a reason that we can't say that we can't find a configuration file?  Since in this test we are specifically not supplying a configuration file.

adrienthebo, line 12: The config file loading happens far before we start picking out values from the config, and supplying a config is no longer a hard requirement for running r10k. It's possible to plumb that through but it'll actually be a bit of a pain.

zreichert, line 12: @adrienthebo Perhaps we should remove or rework this test case if the config is no longer required.

PR #524, opened 2015-10-22
[action/deploy] Raise error on missing config

adrienthebo:
The deploy actions all require information from a config, and when no
config is provided then `r10k deploy` can't run at all. This commit adds
another helper to fail with a useful error message so that users know
how to actually fix r10k when it can't find config file data.


scotje:
Oh yeah, I looked at this yesterday but got distracted by something shiny....

:+1:

PR #529, opened 2015-11-02
(RK-154) Add per-repository config

adrienthebo:
This allows users using the Rugged git provider to set a specific SSH private key on a per-remote basis.

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -68,6 +68,32 @@ git:
 See the [git provider documentation](../git/providers.mkd) for more information
 regarding Git providers.
 
+#### username
+
+The username option sets the username for SSH remotes when the SSH URL does not provide
+a username. When used with a Git hosting service this is most sensibly set to 'git'.
+
+The username defaults to the username of the currently logged in user.
+
+The username setting is only used by the Rugged git provider.
+
+```yaml
+git:
+  username: "git"
+```
+
+#### private_key
+
+The private_key option specifies the path to the default Git SSH private key for Git SSH remotes.
+The private_key setting must be set if SSH remotes are used.
+
+The private_key setting is only used by the Rugged git provider.
+
+```yaml
+git:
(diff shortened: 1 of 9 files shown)


doc/dynamic-environments/configuration.mkd
```diff
@@ -68,6 +68,49 @@ git:
...
 regarding Git providers.
 
+#### username
+
+The username option sets the username for SSH remotes when the SSH URL does not provide
+a username. When used with a Git hosting service this is most sensibly set to 'git'.
+
+The username defaults to the username of the currently logged in user.
+
+The username setting is only used by the Rugged git provider.
```
adreyer: Can this go at the top.

doc/dynamic-environments/configuration.mkd
```diff
@@ -68,6 +68,49 @@ git:
...
+git:
+  username: "git"
+```
+
+#### private_key
+
+The private_key option specifies the path to the default Git SSH private key for Git SSH remotes.
+The private_key setting must be set if SSH remotes are used.
+
+The private_key setting is only used by the Rugged git provider.
```
adreyer: same

doc/dynamic-environments/configuration.mkd
```diff
@@ -68,6 +68,49 @@ git:
...
+git:
+  private_key: "/etc/puppetlabs/r10k/ssh/id_rsa"
+```
+
+#### repositories
+
+The repositories option allows configuration to be set on a per-remote basis. Each entry is a map of
+the repository URL and per-repository configuration for that repo.
+
+##### private_key
```
adreyer, line 102: Should there be a user here too for parity?

adrienthebo, line 102: Since the URL itself can contain the user, it doesn't really provide a lot of value to provide two places to specify a username - the global username is mainly useful for setting a sane default for all URLs for services like GitHub where the username is always `git`.

PR #530, opened 2015-11-02
(CODEMGMT-440) Defer rugged repo alternates setup

adrienthebo:
Attempting to reset the git alternates file when creating Git
repositories means that actions with side effects can occur during
object initialization, which is pretty terrible and can result in some
very odd bugs. This commit defers the rugged repo setup/alternates setup
till the first time the rugged repo is accessed, which should remove
this problem.


adreyer:
Looks like a good start. We should probably do this for shellgit eventually too.

PR #531, opened 2015-11-03
(CODEMGMT-440) Update shellgit thin repo alternates

adrienthebo:
This defers syncing the alternates file till the first time the Git command is run so the file is only updated when needed.

### DIFF
--- a/lib/r10k/git/shellgit/thin_repository.rb
+++ b/lib/r10k/git/shellgit/thin_repository.rb
@@ -11,12 +11,6 @@ class R10K::Git::ShellGit::ThinRepository < R10K::Git::ShellGit::WorkingReposito
   def initialize(basedir, dirname, cache_repo)
     @cache_repo = cache_repo
     super(basedir, dirname)
-    if git_dir.exist?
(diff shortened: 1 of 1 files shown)


lib/r10k/git/shellgit/thin_repository.rb
```diff
@@ -51,4 +45,21 @@ def setup_cache_remote
     git ["remote", "add", "cache", @cache_repo.git_dir.to_s], :path => @path.to_s
     fetch
   end
+
+  def git(cmd, opts = {})
+    if @_synced_alternates
```
scotje: Should this be if-not or unless?

adrienthebo: Yep, good catch.

scotje:
:+1: Looks good to me

adrienthebo:
We're going to ship this in r10k 2.1.2, so holding off on merging it till we ship 2.1.1.

scotje:
We have a dependent PR which needs to go in 2.1.1 so we might as well include this.

PR #532, opened 2015-11-04
(RK-181) Set baseurl/proxy with shared PuppetForge URL

adrienthebo:
The PuppetForge 2.0 release changed how Faraday connections were passed
around, and r10k wasn't passing the Forge baseurl information to all of
the necessary Faraday clients. This pull request fixes the error by
explicitly configuring the PuppetForge host and connection proxy
information for the shared PuppetForge connection.


scotje:
:+1:

scotje:
Don't see why this can't go into 2.1.1 as well.

PR #534, opened 2015-11-06
(RK-187) Consider thin repos with a .git file (not directory) to be mismatched.

scotje:


adreyer:
+1 once it's tested against pe

scotje:
Manually tested in a clean PE install.

PR #535, opened 2015-11-10
Fix modulepath examples in environment.conf

tobiaspal:
If I try to use the modulepath settings as they are documented now I get:

```
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/settings/config_file.rb:37:in `block (2 levels) in parse_file': Could not match line modulepath: modules:external-modules (Puppet::Settings::ParseError)
 at /etc/puppetlabs/code/environments/master/environment.conf:2
```

Also fix the double colon in the second example

### DIFF
--- a/doc/faq.mkd
+++ b/doc/faq.mkd
@@ -99,7 +99,7 @@ environment to indicate which directories contain modules:
 
 ```
 # environment.conf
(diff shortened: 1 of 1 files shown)


adrienthebo:
Thanks!

PR #536, opened 2015-11-10
(CODEMGMT-453) Preliminary JRuby support for r10k/shellgit.

scotje:
This seems to be the minimal set of changes required to get the spec tests running under JRuby 1.7 and JDK8. Everything apart from public Forge interactions _should_ work under JDK7, public Forge doesn't work due to JDK7 not supporting DH keys longer than 1024 bytes.

### DIFF
--- a/lib/r10k/git.rb
+++ b/lib/r10k/git.rb
@@ -2,6 +2,7 @@ require 'r10k/features'
 require 'r10k/errors'
 require 'r10k/settings'
 require 'r10k/logging'
+require 'r10k/util/platform'
 
 module R10K
   module Git
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,6 @@ rvm:
   - "2.1.0"
   - "2.0.0"
   - "1.9.3"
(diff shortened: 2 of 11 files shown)


lib/r10k/git.rb
```diff
@@ -19,8 +20,12 @@ module Git
           :feature => :shellgit,
           :module  => R10K::Git::ShellGit,
         }
-      ],
-      [ :rugged,
+      ]
+    ]
+
+    unless R10K::Util::Platform.jruby?
+      @providers << [
```
adrienthebo: I don't think we need to gate this here, since the feature detection logic can look for the presence/absence of the gem (and since it can't be installed in JRuby it shouldn't be used.)

scotje: It must have been one of the now-skipped test scenarios that was causing the on-set proc to be evaluated (which blows up if the Rugged module isn't defined)...

adrienthebo:
:+1:, either I can merge or you can - your call.

PR #540, opened 2015-11-27
Add r10k:syntax rake task to check the syntax of the Puppetfile

tampakrap:
In order to use it, add the following in the Rakefile: require 'r10k/tasks'
and run '(bundle exec) rake r10k:syntax'

### DIFF
--- a/lib/r10k/tasks.rb
+++ b/lib/r10k/tasks.rb
@@ -1,0 +1,24 @@
+require 'rake'
+require 'rake/tasklib'
+
+module R10KTasks
+  class RakeTask < ::Rake::TaskLib
+    def initialize(*args)
+      namespace :r10k do
+        desc "Syntax check Puppetfile"
+        task :syntax do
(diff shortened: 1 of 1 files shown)


adrienthebo:
Why not use `r10k puppetfile check` instead of a rake task?

tampakrap:
I added the r10k:syntax task in the :test task, so it's faster to type "bundle exec rake test" than "bundle exec rake r10k puppetfile check && bundle exec rake test" when I want to run validation tests locally

adrienthebo:
Since you've published this at https://rubygems.org/gems/ra10ke I would rather not take on the burden of maintaining this in r10k core while having a separately maintained and published implementation. Would you be willing/interested in adding this to the r10k documentation?

tampakrap:
Sure, I'll file a separate PR to add it to the r10k documentation. Just for the record, I was perfectly happy to drop my gem in case the rake tasks were going to be added to the r10k codebase, but I'm fine either way

PR #545, opened 2015-12-16
Remove vestigial R10K::ModuleRepository::Forge class.

scotje:
The switch to the puppet_forge gem seems to have obviated the need for
this class as well as the direct dependencies on Faraday.


scotje:
Unless I missed the usage somewhere....

PR #547, opened 2015-12-17
(RK-189) Implement R10K::API envmap generation functions.

scotje:
Initial implementations for the basic Puppetfile extraction and parsing functions, as well as inspecting an already deployed environment.

There are a few changes mixed in here to just make it possible to require various bits of R10K in piecemeal since since Ruby's global namespace context makes it easy to forget to require/define various things where they are used.

The testing is not as robust as it should be yet, but there is some basic unit test coverage.

### DIFF
--- a/lib/r10k/api.rb
+++ b/lib/r10k/api.rb
@@ -1,9 +1,16 @@
+require 'r10k/api/envmap_builder'
+
+require 'r10k/logging'
+require 'r10k/git'
+require 'r10k/svn/remote'
+require 'r10k/puppetfile'
+
 module R10K
   # A low-level interface for triggering r10k operations.
   #
   # @example Parse a Puppetfile and deploy the environment it represents:
   #   # Given "ops_source" is an instance of R10K::Source
-  #   puppetfile = R10K::API.get_puppetfile(ops_source.to_hash, "production")
+  #   puppetfile = R10K::API.get_puppetfile(ops_source.type, ops_source.cache.path, "production")
   #   envmap = R10K::API.parse_puppetfile(puppetfile)
   #
   #   R10K::API.module_sources_for_environment(envmap).each do |src|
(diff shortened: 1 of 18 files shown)


lib/r10k/api/envmap_builder.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      # @param [String] name
+      # @param [Hash] args
+      def add_module(name, args)
+        name_parts = name.split(/[\-\/]/, 2)
+        mod_name = name_parts.size > 1 ? name_parts.last : name
+
+        if @modules.collect { |m| m[:name] }.include?(mod_name)
+          raise RuntimeError, "Puppetfile cannot declare the same module name twice: #{mod_name} was already declared"
```
andersonmills, line 33: Wasn't there a discussion about this, and currently r10k doesn't error on this. One of the two declarations wins. I don't remember which. Don't we want to mimic the behavior of r10k? https://tickets.puppetlabs.com/browse/RK-101

lib/r10k/api.rb
```diff
@@ -212,5 +259,54 @@ def deploy_module_into_env(module_name, env_map, path, opts={})
...
+      # Find the deployed version of each module.
+      env_data[:modules].map! do |mod|
+        mod_path = File.join(path, moduledir, mod[:name])
+        mod.merge(parse_deployed_module(mod_path, mod[:type]))
+      end
+
+      return env_data
+    end
+
+    def parse_deployed_svn_env(path, moduledir)
```
andersonmills, line 291: I kind of want to raise a NotImplementedError here, just to be able to trace what happened easier if this ever comes up.

lib/r10k/api/envmap_builder.rb
scotje, line 33: My inclination would be to make this an error going forward and since this code won't impact existing functionality I'm not sure we need to match the current functionality. This is the kind of thing that would make switching the CLI actions to use this code require a major version bump though. Maybe we should just add a comment or something documenting that this varies from existing behavior?

lib/r10k/api.rb
scotje, line 291: Yeah that seems reasonable.

lib/r10k/api/envmap_builder.rb
andersonmills, line 33: I do think that (major version before switching behavior) was the intention with calling out an error on the dual declaration in the current code anyway.
I'm fine with this being different for good reason.
I think the comment will be helpful, but may be unnoticed when this becomes the standard.

andersonmills:
:+1:  Woot!

PR #548, opened 2015-12-21
Work around some bugs in Popen3 that exist in older versions of JRuby.

scotje:


lib/r10k/util/subprocess/runner/jruby.rb
```diff
@@ -14,8 +14,7 @@ def initialize(argv)
   end
 
   def run
-    spawn_opts = @cwd ? {:chdir => @cwd} : {}
-    stdout, stderr, status = Open3.capture3(*@argv, spawn_opts)
+    stdout, stderr, status = Open3.capture3(@argv.join(' '))
```
adrienthebo, line 22: This change will ignore the `cwd` variable, right? This looks like a regression on https://github.com/puppetlabs/r10k/commit/9b48cb27b71abe12646bae601e76f17f6c2d44c1 .

scotje, line 22: my understanding is that JRuby ignores the spawn options anyway :/

I'll look again and see if there is a better way to pass that through

adrienthebo, line 22: Well, the `cwd` value is currently only used by the SVN code, so if this is just JRuby then this is less of a worry.

scotje, line 22: https://github.com/jruby/jruby/issues/2298

Sounds like JRuby 9000 has a much better implementation, but we probably are going to be using a JGit backend in production anyway, the shellgit provider is mostly for testing.

scotje:
Added a log warning if you try to set a CWD under JRuby.

PR #550, opened 2015-12-28
(RK-198) Parse environment dir

andersonmills:
With this commit, the method is added to parse an entire environment
directory. Tests are included to make sure the proper calls are made.

### DIFF
--- a/lib/r10k/api.rb
+++ b/lib/r10k/api.rb
@@ -95,8 +95,19 @@ module R10K
     # of every environment found.
     #
     # @param path [String] Path on disk to search for Puppet environments.
+    # @param moduledir [String] The path, relative to the environment path, where the modules are deployed.
     # @return [Array<Hash>] An array of hashmaps, each representing the actual state of a single environment found in environmentdir.
-    def parse_environmentdir(path)
+    def parse_environmentdir(path, moduledir="modules")
+      deployed_env_states = []
+
+      if path
+        deployed_envs = Dir.glob(File.join(path, '*')).select {|f| File.directory? f}
+        deployed_envs.each do |env_dir|
+          deployed_env_states << parse_deployed_env(env_dir, moduledir)
+        end
+      end
+
+      return deployed_env_states
     end
 
     # Return a single module_source hashmap for the given module_name from the given env_map.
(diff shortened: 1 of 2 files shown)


lib/r10k/api.rb
```diff
@@ -97,6 +97,20 @@ def parse_deployed_env(path, moduledir="modules")
     # @param path [String] Path on disk to search for Puppet environments.
     # @return [Array<Hash>] An array of hashmaps, each representing the actual state of a single environment found in environmentdir.
     def parse_environmentdir(path)
```
scotje: We should probably add the ability to pass a custom moduledir down through as well. (That's an argument that got added after this was mocked out.)

lib/r10k/api.rb
```diff
@@ -97,6 +97,20 @@ def parse_deployed_env(path, moduledir="modules")
...
+      deployed_env_states = []
+
+      if path
+        dir_entries = Dir.glob(File.join(path, '*'))
+
+        if !dir_entries.empty?
```
adrienthebo: We don't need to check to see if the list is empty; calling `Array#select` on an empty list will execute zero times; same with `Array#each`. So all in all, we can omit this check and the code will still behave correctly.

spec/unit/api_spec.rb
```diff
@@ -114,6 +114,30 @@
     end
   end
 
+  describe ".parse_environmentdir" do
+
+    it "calls parse_deployed_env on each dir in the path" do
+      @env_path = '/tmp/environments'
```
adrienthebo: Using instance variables to carry state between tests can lead to order specific test failures; prefer `let` blocks (http://betterspecs.org/#let) and initializing objects with `before` blocks over instance variables.

adrienthebo: Actually, it looks like there's not a lot of reuse in variables between tests; if a variable is only used in a single test it should be a local variable instead of an instance variable so that it can't accidentally leak between tests.

lib/r10k/api.rb
andersonmills: Cool. Will remove.

spec/unit/api_spec.rb
andersonmills: Agreed. Changing to local variables.

PR #551, opened 2016-01-06
(RK-196) Basic implementation of write-to-disk functions for R10K::API

scotje:
Can squash before merge, although for once my history isn't too messy. :)

### DIFF
--- a/lib/r10k/api.rb
+++ b/lib/r10k/api.rb
@@ -1,4 +1,5 @@
 require 'r10k/api/envmap_builder'
+require 'r10k/api/git'
 
 require 'r10k/logging'
 require 'r10k/git'
@@ -25,8 +26,8 @@ module R10K
     extend R10K::Logging
 
     # TODO: split functions up into submodules and extend those instead?
-    extend self
-
+    # Everything that follows is a module_function until we get to some privates at the end.
+    module_function
 
     # Returns the contents of Puppetfile inside the given control repo source at the given version. Assumes source repo cache has already been updated if applicable.
     #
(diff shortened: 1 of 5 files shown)


lib/r10k/api.rb
```diff
@@ -72,6 +77,42 @@ def parse_puppetfile(io_or_content)
...
+    # @return [Hash] A hashmap representing the desired state of the environment matching the given branch in the given source.
+    # @raise [NotImplementedError] Control repo source has a type that is not currently supported.
+    # @raise [RuntimeError] Something bad happened, see exception message.
+    def envmap_from_source(source, branch_name, base_cachedir)
+      # TODO: enforce valid Puppet environment name here?
+      env_name = branch_name
+      env_name = "#{source[:prefix]}_#{env_name}" if source[:prefix]
+
+      case source[:type].to_sym
+      when :git
```
adrienthebo, line 90: Why are we adding a source `type` class, converting it to a symbol, and then doing a comparison of that symbol rather than just checking the class type? Seems like `when R10K::Source::Git ...` will do the same thing with less work.

adrienthebo, line 90: Wait, I see why. Scratch that.

lib/r10k/api/git.rb
```diff
@@ -0,0 +1,30 @@
+require 'r10k/git'
+require 'r10k/logging'
+
+module R10K
+  module API
+    module Git
+      extend R10K::Logging
+      extend self
```
adrienthebo: Why are we using `extend self` over `module_function`?

spec/unit/api_spec.rb
```diff
@@ -139,6 +165,263 @@
...
+      allow(File).to receive(:open).and_yield(mock_fh)
+
+      expect(R10K::API).to receive(:write_env_base).once.with(envmap, path, anything).and_return(true)
+
+      subject.write_environment(envmap, path)
+    end
+
+    it "should invoke write_module once for each module" do
+      path = "/tmp"
+      module_count = modules.size
```
andersonmills: Is module_count used in this test?

andersonmills:
:+1: from me.

PR #552, opened 2016-01-07
Add Mercurial support

pepijnve:
At the moment this PR represents an initial implementation of Mercurial support.
I've created this already to get early feedback. Does this approach look like the proper way to integrate with Mercurial?


pepijnve:
I made this change from the wrong branch. Let me fix that first...

lib/r10k/source/hg.rb
```diff
@@ -0,0 +1,124 @@
...
+require 'r10k/environment/hg'
+require 'r10k/environment/name'
+
+# This class implements a source for Git environments.
+#
+# A Git source generates environments by locally caching the given Git
+# repository and enumerating the branches for the Git repository. Branches
+# are mapped to environments without modification.
+#
+# @since 1.3.0
```
adrienthebo: This version information is incorrect.

lib/r10k/source/hg.rb
```diff
@@ -0,0 +1,124 @@
+require 'r10k/hg'
+require 'r10k/hg/cache'
+require 'r10k/environment/hg'
+require 'r10k/environment/name'
+
+# This class implements a source for Git environments.
```
adrienthebo, line 6: Bit of a nitpick, but this comment will need to be updated.

lib/r10k/source/hg.rb
```diff
@@ -0,0 +1,124 @@
...
+# @since 1.3.0
+class R10K::Source::Hg < R10K::Source::Base
+
+  include R10K::Logging
+
+  R10K::Source.register(:hg, self)
+
+  # @!attribute [r] type
+  #   @return [:hg] The type of this control repo source
+  attr_reader :type
```
adrienthebo: Where is this used?

pepijnve: I had made these changes on the r10k-api branch by accident at first. I rebased to master and this bit got carried over. I'll get remove it since it's not relevant on the master branch yet.

lib/r10k/hg/stateful_repository.rb
```diff
@@ -0,0 +1,86 @@
...
+    sha = @cache.resolve(@rev)
+
+    if sha.nil?
+      raise R10K::Hg::UnknownRevisionError.new("Unable to sync repo to unresolvable revision '#{@rev}'", :hg_dir => @repo.hg_dir)
+    end
+
+    case status
+      when :absent
+        logger.debug { "Cloning #{@repo.path} and checking out #{@rev}" }
+        @repo.clone(@cache.hg_dir.to_s, {:rev => sha})
```
adrienthebo: The git implementation does a lot of work to avoid copying extra objects from the cached repo around; how efficient is cloning a local repo via hg?

pepijnve: Unfortunately Mercurial doesn't allow the same level of sharing between clones that Git does. Local clones use hard links, but that's only for the initial clone. Pulling changes triggers copy-on-write logic that clears the hard links for any changed files. I don't think there's much that can be done about this.

I kept the cached local repo to avoid hitting the network for each environment.

pepijnve: I was digging through the Mercurial man pages just now looking for a way to improve this. One easy improvement is passing the `--branch` parameter when calling `clone` and `pull` for the environment clones. That way we only clone/pull the data that is relevant for each environment rather than cloning everything for each environment.

scotje:
This seems like a reasonable approach, although I need to familiarize myself with some of the Hg nomenclature to understand exactly how it maps to r10k operations.

We will need to develop a suite of unit/spec tests for the new implementation before merging, and we should also consider how to eventually extend the in-progress r10k-api work in a similar fashion.

pepijnve:
Absolutely, writing tests is next on my todo list.

On the nomenclature, https://www.mercurial-scm.org/wiki/GitConcepts correlates concepts on both sides pretty well. Bookmarks are the closest thing to git branches that Mercurial has. Mercurial's named branches don't have an equivalent in Git.

BTW would you prefer to have all the commits squashed or can I keep the individual ones for the time being? Whatever makes it easier to review for you guys is fine by me.

scotje:
Thanks for the link, I'll take a look.

Individual commits is fine for now, we can squash as appropriate when we're ready to merge.

pepijnve:
Ugh, seems like I didn't rebase this the right way... I'll make a new PR.

ghost:
@scotje  Are you considering to merge the PR #569?

pepijnve:
@phononet this PR is only half finished I'm afraid. Mainly missing unit tests. The need for this at work went away and I haven't had sufficient spare time to continue work on it.

PR #554, opened 2016-01-12
Fix accessibility of private helpers from API module functions.

scotje:
Not my favorite solution but should work at least until we re-namespace things.


andersonmills:
:+1:

PR #555, opened 2016-01-12
(CODEMGMT-199) rjgit provider

andersonmills:
This commit adds an rjgit provider to r10k, but only for the r10k-api.

And, there are no tests, yet. That will be put together in RK-202.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,8 @@ gemspec
 
 group :extra do
   gem 'rugged', '~> 0.21.4', :platforms => :ruby
+  gem 'rjgit', :git => 'https://github.com/andersonmills/rjgit.git', :branch => 'add_git_dir_to_repo', :platforms => :jruby
+#  gem 'rjgit', '~> 4.1.1.0', :platforms => :jruby
 end
 
 group :development do
--- a/lib/r10k/api/git.rb
+++ b/lib/r10k/api/git.rb
@@ -23,6 +23,10 @@ module R10K
       private
 
       def self.provider
+        if R10K::Features.available?(:rjgit)
+          R10K::Git.provider = :rjgit
+        end
+
         R10K::Git.provider
       end
       private_class_method :provider
--- a/lib/r10k/git/rjgit.rb
+++ b/lib/r10k/git/rjgit.rb
@@ -1,0 +1,48 @@
+require 'r10k/logging'
+
+if R10K::Features.available?(:rjgit)
+  require 'rjgit'
+end
+
+module R10K
+  module Git
+    module RJGit
+      extend R10K::Logging
+
+      module_function
+
+      def reset(ver, opts = {})
+        reset_mode = opts[:hard] ? "HARD" : "MIXED"
+
+        repo = ::RJGit::Repo.new(opts[:work_tree], git_dir: opts[:git_dir])
+        opts[:repo] = repo
+        commit = resolve_version(ver, opts)
+        repo.git.reset(commit, reset_mode)
+      end
+
+      def clean(opts = {})
+        # TODO: implement excludes:
+        if opts[:excludes]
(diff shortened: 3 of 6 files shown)


Gemfile
```diff
@@ -4,6 +4,8 @@ gemspec
 
 group :extra do
   gem 'rugged', '~> 0.21.4', :platforms => :ruby
+  gem 'rjgit', :git => 'git@github.com:andersonmills/rjgit.git', :branch => 'add_git_dir_to_repo'
```
adreyer: Can this install for the jruby engine only? What happens otherwise?

lib/r10k/git/rjgit.rb
```diff
@@ -0,0 +1,51 @@
+require 'r10k/logging'
+require 'r10k/util/subprocess'
+
+begin
```
adreyer: Should this use the feature?

lib/r10k/git/rjgit.rb
```diff
@@ -0,0 +1,51 @@
...
+  require 'rjgit'
+rescue LoadError
+end
+
+module R10K
+  module Git
+    module RJGit
+      extend R10K::Logging
+      include ::RJGit
```
adreyer: This seems like it muddies the API.

andersonmills: Agreed. Removed.

lib/r10k/git/rjgit.rb
andersonmills: Do you mean like this?

```
if R10K::Features.available?(:rjgit)
  require 'rjgit'
end
```

Gemfile
andersonmills: Ah. I missed that I'd removed the :platform pair when I changed to git.

bin/r10k_api_test
```diff
@@ -0,0 +1,21 @@
...
+version = "fusion"
+puts Git.rev_parse(version, git_dir: cachedir)
+#git.reset(version, hard: true, git_dir: cachedir, work_tree: path)
+version = "master"
+Git.reset(version, hard: true, git_dir: cachedir, work_tree: path)
+puts "after reset Dir = #{Dir.entries(path)}"
+Git.clean(git_dir: cachedir, work_tree: path)
+puts "after clean Dir = #{Dir.entries(path)}"
+Git.clean(excludes: "leave_this", git_dir: cachedir, work_tree: path)
+puts "after clean with excludes Dir = #{Dir.entries(path)}"
```
scotje: Did you mean to commit this file? We do need some sort of acceptance testing for the API, but this would probably need a little work to be cross-developer friendly.

andersonmills: This file will not be in the final commit. I wanted folks to see how I was testing it. I mentioned that in the description, but it's easy to miss.

scotje: Ooops, sorry. I usually just jump straight to the diff. :)

lib/r10k/api/git.rb
```diff
@@ -23,6 +23,7 @@ def rev_parse(rev, opts={})
       private
 
       def self.provider
+        R10K::Git.provider = :rjgit
```
adreyer: this probably shouldn't be hardcoded permanently

bin/r10k_api_test
adreyer: Is there functional testing you can do now at the API level?

spec/unit/git_spec.rb
```diff
@@ -17,14 +17,33 @@
         expect(R10K::Features).to receive(:available?).with(:rugged).and_return true
         expect(described_class.default_name).to eq :rugged
       end
+
+      it 'raises an error when the git executable and rugged library are absent' do
```
scotje: this description should probably be broadened

PR #556, opened 2016-01-12
(RK-190) Implement basic env_map resolution functions

scotje:
This adds the ability to resolve environment maps containing git and Forge sourced module declarations.

Still needs unit testing and there are some bits that will need to be adjusted to work with the jgit provider, but the basic resolution logic is all in place.

### DIFF
--- a/lib/r10k/api.rb
+++ b/lib/r10k/api.rb
@@ -6,6 +6,9 @@ require 'r10k/git'
 require 'r10k/svn/remote'
 require 'r10k/puppetfile'
 
+require 'puppet_forge'
+require 'semantic_puppet'
+
 module R10K
   # A low-level interface for triggering r10k operations.
   #
@@ -24,6 +27,17 @@ module R10K
   #
   module API
     extend R10K::Logging
+
+    class UnresolvableError < StandardError
+      def initialize(message, modules=nil)
+        if modules && modules.respond_to?(:each)
+          modules.each { |m| message << "\n#{m[:name]} could not be resolved: #{m[:error].message}" }
+        end
+
+        super(message)
+      end
+    end
+
 
     # TODO: split functions up into submodules and extend those instead?
     # Everything that follows is a module_function until we get to some privates at the end.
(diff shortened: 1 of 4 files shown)


lib/r10k/api.rb
```diff
@@ -25,6 +28,9 @@ module R10K
   module API
     extend R10K::Logging
 
+    class UnresolvableError < StandardError; end
```
scotje: this needs to be fleshed out a bit to support the way I'm actually using it below

lib/r10k/api/git.rb
```diff
@@ -17,7 +19,13 @@ def clean(opts={})
...
 
       def rev_parse(rev, opts={})
-        provider.rev_parse(rev, opts)
+        result = provider.rev_parse(rev, opts)
+
+        if result.success?
+          return result.stdout.strip
+        else
+          raise CommandFailedError.new(result.stderr)
+        end
```
scotje, line 28: This is the part that is not going to work with the jgit provider. Need to sit down and decide how we want to standardize the return values from the various providers.

lib/r10k/api.rb
```diff
@@ -211,8 +243,74 @@ def resolve_environment(env_map)
...
     # @return [Hash] A copy of env_map with a new :resolved_version key/value pair added for the specified module.
-    def resolve_module(module_name, env_map)
+    # @raise [RuntimeError, R10K::API::UnresolvableError] Something bad happened.
+    def resolve_module(module_name, env_map, opts={})
+      mod_found = false
+
+      env_map[:modules].map! do |mod|
+        if mod[:name] == module_name
+          case mod[:type].to_sym
+          when :git
```
adreyer, line 264: nitpick... these seem big enough to break into their own methods

scotje, line 264: yes, I reached that conclusion as well :) I'll try to get that done in the morning once I've got the unit tests written

lib/r10k/api.rb
```diff
@@ -211,8 +243,74 @@ def resolve_environment(env_map)
...
+            if !opts[:cachedir]
+              raise RuntimeError.new("A value is required for the :cachedir option when resolving module from a git source.")
+            end
+
+            cachedir = cachedir_for_git_remote(mod[:source], opts[:cachedir])
+
+            begin
+              mod[:resolved_version] = git.rev_parse(mod[:version], git_dir: cachedir)
```
adreyer: I wish we wouldn't treat refs as versions... I don't know how feasible that is now

andersonmills:
:+1:

PR #559, opened 2016-01-19
(RK-205) Standardize exception handling between shellgit and rjgit providers.

scotje:


spec/unit/git/rjgit_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    it "raises R10K::Git::GitError with message when unable to resolve ref" do
+      expect(subject).to receive(:resolve_version).with('testref', anything).and_return(nil)
+
+      expect { subject.reset('testref') }.to raise_error(R10K::Git::GitError, /could not resolve/i)
+    end
+
+    it "raises R10K::Git::GitError with message on CheckoutConflictException" do
+      expect(subject).to receive(:resolve_version).with('testref', anything).and_return('testsha')
+      expect(repo_git).to receive(:reset).with('testsha', anything).and_raise(Java::OrgEclipseJgitApiErrors::CheckoutConflictException.new(['/foo/bar'], Java::OrgEclipseJgitErrors::CheckoutConflictException.new('/foo/bar')))
```
scotje, line 29: Oh how I love Java exception hierarchies.

lib/r10k/git/rjgit.rb
```diff
@@ -41,7 +66,6 @@ def self.resolve_version(ver, opts = {})
         repo = opts[:repo] || ::RJGit::Repo.new(opts[:git_dir], is_bare: true)
         ref = repo.commits(ver).first
```
andersonmills: Does this need protection in case commits(ver) returns nil?

spec/unit/git/rjgit_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      expect(subject.clean()).to eq(true)
+    end
+
+    it "raises R10K::Git::GitError with message on NoWorkTreeException" do
+      expect(repo).to receive(:clean).and_raise(Java::OrgEclipseJgitErrors::NoWorkTreeException.new)
+
+      expect { subject.clean() }.to raise_error(R10K::Git::GitError, /neither a working tree.*nor an index/i)
+    end
+
+    it "supports opts[:excludes]" do
```
andersonmills, line 55: I thought we were going to try to rewrite to avoid the need for this?

lib/r10k/git/rjgit.rb
scotje: In most cases it looks like it will return an empty array but it will return nil if you give it a nil repo so we should probably guard against that.

spec/unit/git/rjgit_spec.rb
scotje, line 55: We are (in fact, turns out it's actually already unnecessary for the `.r10k-deploy.json`), it seems fairly harmless to have it plumbed in even if not implemented but I'm happy to just tear everything out if you would prefer.

lib/r10k/git/rjgit.rb
andersonmills: I think that seems totally reasonable to add on this PR.

spec/unit/git/rjgit_spec.rb
andersonmills, line 55: Meh. Feels weird to have you tear it out and feels weird for it to be in. I don't have a strong preference, but was surprised to see it pending since it's not really.

scotje, line 55: If we get to the end and have no forseeable application/need for it, we can take it out.

andersonmills:
:+1: All the way!

PR #560, opened 2016-01-20
(RK-195) Serialize resolved env-map to disk after deploy

scotje:
Also converts the .blob_at functionality to be portable across providers.

### DIFF
--- a/lib/r10k/git/rjgit.rb
+++ b/lib/r10k/git/rjgit.rb
@@ -99,6 +99,18 @@ module R10K
         return true
       end
 
+      def blob_at(treeish, path, opts = {})
+        repo = ::RJGit::Repo.new(opts[:git_dir], is_bare: true)
+
+        begin
+          return repo.blob(path, treeish).data
+        rescue Java::OrgEclipseJgitErrors::LargeObjectException,
+               Java::OrgEclipseJgitErrors::MissingObjectException,
+               Java::JavaIO::IOException => e
+          raise R10K::Git::GitError.new(e.message)
+        end
+      end
+
       private
 
       def self.resolve_version(ver, opts = {})
--- a/lib/r10k/git/shellgit.rb
+++ b/lib/r10k/git/shellgit.rb
@@ -89,7 +89,21 @@ module R10K
         result = git(cmd, opts)
 
         if result.success?
-          return result.stdout.strip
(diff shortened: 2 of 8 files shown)


lib/r10k/git/rjgit.rb
```diff
@@ -60,6 +60,13 @@ def rev_parse(ver, opts = {})
         return commit.id
       end
 
+      def blob_at(ver, path, opts = {})
+        repo = ::RJGit::Repo.new(opts[:git_dir], is_bare: true)
```
adreyer, line 103: should we check for this opt what happens otherwise?

adreyer:
I think blob_at should probably require a git_dir since it's unlikely to behave well without one.

scotje:
It seems like #blob_at doesn't behave any worse than anything else when it doesn't get a `:git_dir` so I'm inclined to leave it be for now and address the option validation more comprehensively in a separate ticket.

lib/r10k/git/rjgit.rb
```diff
@@ -60,6 +60,13 @@ def rev_parse(ver, opts = {})
...
+        blob = repo.blob(path, ver)
+
+        return blob.data
```
andersonmills: Do we want to raise an Error on failure as we're doing in the shellgit version of this?

scotje: Yeah, let me dig down and figure out what hideously long exception classes that can raise....

lib/r10k/git/shellgit.rb
```diff
@@ -67,6 +67,20 @@ def rev_parse(rev, opts = {})
...
 
+      def blob_at(treeish, path, opts = {})
+        opts = { raise_on_fail: false }.merge(opts)
+
+        cmd = ["show", "#{treeish}:#{path}"]
+
+        result = git(cmd, opts)
+
+        if result.success?
+          return result.stdout.strip
```
andersonmills: Should the shellgit and rjgit versions of this return the same thing?

andersonmills: Okay, on further inspection, these only differ by an ending newline. stdout gets chomped in subprocess.

andersonmills:
:+1: I believe!

I believe I can fly!

PR #561, opened 2016-01-26
(RK-191) implement cache update fns

andersonmills:
This commit implements cache updating for modules from git repositories.

### DIFF
--- a/lib/r10k/api.rb
+++ b/lib/r10k/api.rb
@@ -186,22 +186,66 @@ module R10K
     def module_sources_for_environment(env_map)
     end
 
+    # Update local caches represented by the given by module_sources, a collection of module_source hashmaps.
+    #
+    # @param module_sources [Array<Hash>] An array of hashmaps each representing a single remote module source (as produced by {#module_sources_for_environment})
+    # @param base_cachedir [String] Root of where r10k is caching things.
+    # @return [true] Returns true on success, raises on failure.
+    # @raise [RuntimeError] Something bad happened!
+    def update_caches(module_sources, base_cachedir, opts = {})
+      if module_sources.respond_to?(:each)
+        module_sources.each do |module_source|
+          update_cache(module_source, base_cachedir, opts)
+        end
+        return true
+      else
+        raise RuntimeError.new("module_sources must be a collection of source hashes.")
+      end
+
+    end
     # Update local cache represented by the given module_source hashmap.
     #
     # @param module_source [Hash] A hashmap representing a single remote module source (as produced by {#module_sources_for_environment})
+    # @param base_cachedir [String] Root of where r10k is caching things.
     # @return [true] Returns true on success, raises on failure.
     # @raise [RuntimeError] Something bad happened!
-    def update_cache(module_source)
+    def update_cache(module_source, base_cachedir, opts = {})
+      type = module_source[:type]
+      source = module_source[:source]
+
+      raise RuntimeError.new('module source type cannot be nil.') if type.nil?
+
+      case type.to_sym
+      when :git
+        update_git_cache(source, base_cachedir, opts)
+      when :forge
+        raise NotImplementedError
+      when :svn
+        raise NotImplementedError
+      else
+        raise RuntimeError.new("Unrecognized module source type '#{module_source[:type]}'.")
+      end
     end
 
-    # Update local cache of the given remote VCS repository.
+    # Update local cache of the given remote git repository.
     #
     # @param remote [String] URI for the remote repository which should be cached or updated.
+    # @param base_cachedir [String] local directory which contains all specific cachedirs
     # @param opts [Hash] Additional options as defined.
     # @option opts [String] :cachedir Base path where caches are stored.
     # @return [true] Returns true on success, raises on failure.
     # @raise [RuntimeError] Something bad happened!
-    def update_vcs_cache(remote, opts={})
+    def update_git_cache(remote, base_cachedir, opts={})
+      if opts[:path]
+        raise RuntimeError.new("Cannot update a git cache as if it were a worktree with a .git directory at #{opts[:path]} for remote #{remote}")
+      end
+
+      opts[:git_dir] = cachedir_for_git_remote(remote, base_cachedir)
+      if File.directory?(opts[:git_dir])
+        git.fetch(remote, opts)
+      else
+        git.clone(remote, opts[:git_dir], opts.merge({bare: true}))
+      end
     end
 
     # Update local cache of the given module from the Puppet Forge.
@@ -213,7 +257,8 @@ module R10K
     # @option opts [String] :baseurl The URL to the Puppet Forge to use for downloading modules.
     # @return [true] Returns true on success, raises on failure.
     # @raise [RuntimeError] Something bad happened!
-    def update_forge_cache(module_slug, opts={})
+    def update_forge_cache(module_slug, base_cachedir, opts={})
+      raise NotImplementedError.new("Forge cache updating not implemented yet. Attemepted module = #{module_slug}")
     end
 
     # Given an environment map, returns a new environment map with any ambiguous module versions (e.g. branch names, version ranges, etc.)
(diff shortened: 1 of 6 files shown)


lib/r10k/api.rb
```diff
@@ -186,22 +186,68 @@ def module_source_for_module(module_name, env_map)
...
-    # Update local cache of the given remote VCS repository.
+    # Update local cache of the given remote git repository.
     #
     # @param remote [String] URI for the remote repository which should be cached or updated.
     # @param opts [Hash] Additional options as defined.
     # @option opts [String] :cachedir Base path where caches are stored.
     # @return [true] Returns true on success, raises on failure.
     # @raise [RuntimeError] Something bad happened!
-    def update_vcs_cache(remote, opts={})
+    def update_git_cache(remote, opts={})
```
adreyer: does this function make sense without a cache_basedir? It seems like that should just be an argument.

lib/r10k/git/rjgit.rb
```diff
@@ -60,6 +60,41 @@ def rev_parse(ver, opts = {})
         return commit.id
       end
 
+      def fetch(remote, opts={})
+
+        local = opts[:cachedir]
+        repo = ::RJGit::Repo.new(local, is_bare: true)
```
adreyer, line 68: but local could be nil?

lib/r10k/git/rjgit.rb
```diff
@@ -60,6 +60,41 @@ def rev_parse(ver, opts = {})
...
+               Java::OrgEclipseJgitApiErrors::TransportException,
+               Java::OrgEclipseJgitApiErrors::InvalidRemoteException => e
+          raise R10K::Git::GitError.new(e.message)
+        end
+
+        return true
+      end
+
+      def clone(remote, opts={})
+        opts[:branch] = :all
```
adreyer: this is going to alter the state of opts which seems pretty bad given how they're being fed around.

lib/r10k/git/shellgit.rb
```diff
@@ -53,6 +53,33 @@ def clean(opts = {})
         end
       end
 
+      def fetch(remote, opts = {})
+        cmd = ["fetch"]
```
adreyer: since there aren't any conditionals this might be easier to read if you just defined the array

lib/r10k/git/shellgit.rb
```diff
@@ -53,6 +53,33 @@ def clean(opts = {})
...
+
+        if result.success?
+          return result.stdout.strip
+        else
+          raise R10K::Git::GitError.new(result.stderr)
+        end
+      end
+
+      def clone(remote, opts = {})
+        cmd = ["clone"]
```
adreyer: I don't think splitting this across lines improves readability

lib/r10k/api.rb
```diff
@@ -94,18 +94,18 @@ def parse_puppetfile(io_or_content)
     #
     # @param source [Hash] A hashmap representing a control repo source (like would be defined in r10k.yaml)
     # @param branch_name [String] Branch in the control repo to build an envmap for.
-    # @param base_cachedir [String] Root of where r10k is caching things.
+    # @param cache_basedir [String] Root of where r10k is caching things.
     # @return [Hash] A hashmap representing the desired state of the environment matching the given branch in the given source.
     # @raise [NotImplementedError] Control repo source has a type that is not currently supported.
     # @raise [RuntimeError] Something bad happened, see exception message.
-    def envmap_from_source(source, branch_name, base_cachedir)
+    def envmap_from_source(source, branch_name, cache_basedir)
```
adreyer: since there is a setting cachedir I think base_cachedir is probably better. if cachedir previously only had one meaning maybe just use 'cachedir' here and some other variable for the repo specific caches.

lib/r10k/api.rb
```diff
@@ -186,22 +186,68 @@ def module_source_for_module(module_name, env_map)
...
+    # @param cache_basedir [String] Root of where r10k is caching things.
     # @return [true] Returns true on success, raises on failure.
-    # @raise [RuntimeError] Something bad happened!
-    def update_cache(module_source)
+    # @raise [UnresolvableError] Something bad happened!
+    def update_cache(module_source, cache_basedir, opts = {})
+      type = module_source[:type]
+      source = module_source[:source]
+
+      if type
```
scotje: Could probably just `raise` if type is nil...

lib/r10k/api/git.rb
```diff
@@ -20,6 +20,14 @@ def rev_parse(rev, opts={})
         provider.rev_parse(rev, opts)
       end
 
+      def fetch(remote, opts={})
+        provider.fetch(remote, opts)
+      end
+
+      def clone(remote, opts={})
+        provider.clone(remote, opts)
+      end
```
scotje, line 29: It seems like the signature for clone should have a target path argument and an option to control whether or not it's a bare clone. We could probably clean that up with RK-207 though.

lib/r10k/api.rb
andersonmills: I'm going to switch this to your preferred base_cachedir for now. We have a cleanup ticket in the next sprint where we'll switch entirely to repo_cachedir, base_cachedir, etc.

lib/r10k/api.rb
andersonmills: Clearer, thanks.

lib/r10k/api.rb
andersonmills: Agreed.

lib/r10k/git/rjgit.rb
andersonmills, line 68: Protected.

lib/r10k/git/rjgit.rb
andersonmills: Changed to a local variable for new options passed to rjgit.

lib/r10k/git/shellgit.rb
andersonmills: Moved to one line.

lib/r10k/git/shellgit.rb
andersonmills: Moved to one line.

lib/r10k/api/git.rb
andersonmills, line 29: I kind of think this belongs in this PR, so I'll add it.

lib/r10k/git/rjgit.rb
```diff
@@ -60,6 +60,44 @@ def rev_parse(ver, opts = {})
...
+        return true
+      end
+
+      def clone(remote, opts={})
+        local = opts[:git_dir]
+        raise R10K::Git::GitError.new('clone requires that opts[:git_dir] be a valid directory.') unless local && File.directory?(local)
+
+        rjgit_opts = opts.merge({branch: :all, is_bare: true})
+
+        repo = ::RJGit::Repo.new(local, is_bare: true)
```
scotje: Seems kind of weird that you have to instantiate a repo object before cloning, but if that's the way of things then (shrug)....

lib/r10k/git/shellgit.rb
```diff
@@ -53,6 +53,28 @@ def clean(opts = {})
         end
       end
 
+      def fetch(remote, opts = {})
+        cmd = ["fetch", remote, '+refs/*:refs/*']
+        result = git(cmd, opts)
+
+        if result.success?
+          return result.stdout.strip
```
scotje: shellgit fetch and clone should return bools to be consistent with the rjgit provider...

lib/r10k/git/shellgit.rb
```diff
@@ -53,6 +53,28 @@ def clean(opts = {})
...
+          raise R10K::Git::GitError.new(result.stderr)
+        end
+      end
+
+      def clone(remote, opts = {})
+        cmd = ["clone", "--mirror", remote, opts[:git_dir]]
+        result = git(cmd, opts)
+
+        if result.success?
+          return result.stdout.strip
```
scotje: ditto above

lib/r10k/api.rb
```diff
@@ -186,22 +186,66 @@ def module_source_for_module(module_name, env_map)
...
+    # @param base_cachedir [String] Root of where r10k is caching things.
     # @return [true] Returns true on success, raises on failure.
-    # @raise [RuntimeError] Something bad happened!
-    def update_cache(module_source)
+    # @raise [UnresolvableError] Something bad happened!
+    def update_cache(module_source, base_cachedir, opts = {})
+      type = module_source[:type]
+      source = module_source[:source]
+
+      raise UnresolvableError.new('module source type cannot be nil.') if type.nil?
```
scotje: I think UnresolvableError should probably be reserved for the method(s) that specifically attempt envmap resolution

lib/r10k/api.rb
```diff
@@ -186,22 +186,66 @@ def module_source_for_module(module_name, env_map)
...
+
+      case type.to_sym
+      when :git
+        update_git_cache(source, base_cachedir, opts)
+      when :forge
+        raise NotImplementedError
+      when :svn
+        raise NotImplementedError
+      else
+        raise UnresolvableError.new("Unrecognized module source type '#{module_source[:type]}'.")
```
scotje: ditto above comment

lib/r10k/git/shellgit.rb
andersonmills: Sounds good. stdout.strip is always "" anyway. :)

scotje:
:+1:

andersonmills:
Thanks!

Issue #467, opened 2015-07-07
Allow the Puppetfile data to be yaml


nibalizer:
YAML is life. YAML is good. I would like to be able to express the data that is currently in Puppetfile as a yaml file. This is useful if I want to generate the list and consume the list with other tools.

kostecky:
+1

godlikeachilles:
+1

Issue #543, opened 2015-12-11
Allow some form of variable setting in r10k.yaml for interpolation in the Puppetfile


jharlow1:
We have a situation where our internal git repos are replicated out from dev to our production datacenter(s), so the dev/qa puppet masters access via something like:

git@devgit.ourdevdomain.com:awesome/puppet_module.git

But, in production, the masters need to hit:

git@prodgit.ourproddomain.com:awesome/puppet_module.git

We definitely don't want to maintain a completely separate Puppetfile, so the only option I can think of at the moment is some sort of DNS trickery, which I'd rather avoid.

It would be nice if we could do something in r10k.yaml like:

```
sources:
  ops:
    remote: 'git@prodgit.ourproddomain.com:awesome/control_repo.git'
    basedir: '/etc/puppetlabs/puppet/environments'
    vars:
      gitserver: 'git@prodgit.ourproddomain.com'
```

And then in Puppetfile, do something along the lines of:

```
mod 'puppet_module',
  :git => "#{vars['gitserver']}:awesome/puppet_module.git"
```

etc.

That way we could configure vars['gitserver'] differently on our Puppet masters depending on which one they have access to and keep the same Puppetfile across all our environments.

Figured making the variable a generic hash is probably best since people may have multiple sources or other uses for this functionality.

Thoughts?

Cinderhaze:
I would have a similar use case to this.  
I know that I did something similar for rspec-puppet with .fixtures.yaml, but it only works for the symlinks (whose value is in an eval) and not the repository urls

Issue #133, opened 2014-05-15
multiple moduledir in Puppetfile


naevtamarkus:
I understand you don't mind having multiple moduledir statements in a single Puppetfile. How about the following:

```
moduledir 'modules'   # would be the default, no need to specify it
mod x
mod y
moduledir 'manifests'
mod z
```

I understand that having a begin/end looks nicer... but its usefulness is just the same. 
The deploy part works (I think I already mentioned it before) but not the Purge part. I have been looking at the code and it LOOKS relatively easy to solve, without breaking backwards compatibility:
- Modules code does nott need to be modified, since moduledir is part of its properties (that's why the deploy works)
- The Purge task in task/puppetfile.rb would then use, instead of the moduledir to do the cleaning, a similar approach as with the Deploy  (@puppetfile.load and then @puppetfile.modules.each do |mod|). I still have to figure out how to do this in a language I don't know.
- The desired_contents (in r10k/puppetfile.rb) will have to include the moduledir, not only the module.name itself. The same goes with all current_contents, stale_contents and pending_contents (in util/purgeable.rb), so that the mapping includes full paths and not only module names. I am not sure if making all modules "full-pathed" would break any existing rspec test, but that's probably the smart thing to do.

Note that, after this change, the Puppetfile attr :moduledir does not make any sense (it seems to be used only by the purger, anyway, until this change is done) and should be removed (together with the spec checks). I don't think it hurts if we don't remove it just yet, so I would leave this to a second PR to clean up code.

luisfdez:
@adrienthebo , any update on this enhancement? I'm facing a similar scenario. It's more cosmetic than functional but I find useful to support.

PR #564, opened 2016-02-01
Attempt at extending puppetfile parsing to natively support Puppetfile.yaml

gregswift:
During the [CICD talk at PuppetConf'15](https://www.youtube.com/watch?v=_LZ9vNznwJ8) they mentioned moving their Puppetfile contents into yaml and just having their Puppetfile parse the yaml.  Sounded kewl, so [I did the same](https://gist.github.com/gregswift/84a5ba3c303c0c35f9d3). Since we have several control repositories in our organization it occurred to me that it would be handy if r10k supported the concept natively, then we don't have to distribute the Puppetfile across the repositories.

As mentioned in the talk, it can be easier to wrap CICD automation around a YAML file than the native Puppetfile format. That being said, there are also lots of ways to expose the schema in YAML.  The approach I took in my environments was aimed at simplifying the contents rather than mirroring Puppetfile directly. In this PR I am trying to support more of the native argument handling.  I still think sections per source type has the ability to reduce duplicate data.

_How you expect to use this feature._
I would use this feature to remove the Puppetfile I have distributed through our control repos in favor of letting r10k parse the files.

_How important this feature is for your environment_
It makes things easier, but it is not a deal breaker

_How useful this feature would be for other users_
Based on the handful of people I've shown it has potential to be quite handy. It has also been requested by a few in #467 

_Any potential caveats this may have or any issues this feature may have_
Not sure yet.

At this point I'm submitting the PR for the sake of discussion. It definitely needs more:
- [ ] Support for other module types (forge and git only so far)
- [ ] Documentation
- [ ] Full example file
- [ ] Tests

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -29,12 +29,17 @@ class Puppetfile
   #   @return [String] The path to the Puppetfile
   attr_reader :puppetfile_path
 
+  # @!attrbute [r] puppetfile_yaml_path
+  #   @return [String] The path to the Puppetfile.yaml
+  attr_reader :puppetfile_yaml_path
+
   # @param [String] basedir
   # @param [String] puppetfile The path to the Puppetfile, default to #{basedir}/Puppetfile
   def initialize(basedir, moduledir = nil, puppetfile = nil)
     @basedir         = basedir
     @moduledir       = moduledir  || File.join(basedir, 'modules')
     @puppetfile_path = puppetfile || File.join(basedir, 'Puppetfile')
+    @puppetfile_yaml_path = puppetfile || File.join(basedir, 'Puppetfile.yaml')
 
     @modules = []
     @forge   = 'forgeapi.puppetlabs.com'
@@ -43,6 +48,8 @@ class Puppetfile
   def load
     if File.readable? @puppetfile_path
       self.load!
+    elsif File.readable? @puppetfile_yaml_path
+      self.load_yaml!
     else
       logger.debug "Puppetfile #{@puppetfile_path.inspect} missing or unreadable"
     end
@@ -53,6 +60,52 @@ class Puppetfile
     dsl.instance_eval(puppetfile_contents, @puppetfile_path)
   rescue SyntaxError, LoadError, ArgumentError => e
     raise R10K::Error.wrap(e, "Failed to evaluate #{@puppetfile_path}")
+  end
+
+  def load_yaml!
+    dsl = R10K::Puppetfile::DSL.new(self)
+    require 'yaml'
+    conf = YAML::load_file(@puppetfile_yaml_path)
+    forge_conf = conf.fetch('forge', nil)
+    if forge_conf.nil?
+      dsl.forge("")
+    else
+      forge, modules = forge_conf.first
+      dsl.forge(forge)
+      modules.rach { |m|
+        args = Hash.new
+        if m.is_a?(Hash)
+          name, data = m.first
+          if data.is_a?(Hash)
+            args = data
+          elsif data == 'latest'
+            args['version'] = :latest
+          end
+        else
+          args['version'] = nil
+        end
+        dsl.mod(name, args)
+      }
+    end
+    conf['git'].each { |org_base, modules|
+      modules.each { |m|
+        if !(m.is_a?(Hash))
+          m = { m => 'master' }
+        end
+        args = Hash.new
+        repo, data = m.first
+        if data.is_a?(Hash)
+          args = data
+        else
+          name = repo.split('-', 2)[-1].gsub(/-/, '_')
+          args['git'] = "#{org_base}/#{repo}.git"
+          args['ref'] = data
+        end
+        dsl.mod(name, args)
+      }
+    }
+  rescue SyntaxError, LoadError, ArgumentError => e
+    raise R10K::Error.wrap(e, "Failed to evaluate #{@puppetfile_yaml_path}")
   end
 
   # @param [String] forge


adreyer:
Awesome, not executing arbitrary ruby would be great. I think at this point it would probably be more useful to start with the documentation and examples and discuss the format rather than the implementation.

For example I believe you're suggesting a puppetfile.yaml something like:

```
---
"https://forge.puppetlabs.com/":
  puppetlabs-apache:
    version: "1.0.0"
  puppet-iis:   
    git: "https://github.com/voxpupuli/puppet-iis"
    ref: master
```
- I have some reservations about yaml being the best option for large nested hashes.
- The forge directive isn't supported in r10k and only exists for compatibility with librarian. It probably shouldn't be part of a new spec.

gregswift:
I concur on the documentatio nand examples at this point. I started the PR cause I already had the [code and schema](https://gist.github.com/gregswift/84a5ba3c303c0c35f9d3) I was using.

The format I am using currently looks like this:

``` yaml
forge:
  "https://forgeapi.puppetlabs.com":
    - bfraser/grafana
    - bmurt/puppet_agent: 0.0.4
    - camptocamp/augeas: latest

git:
  "https://github.com":
    - jenkinsci/puppet-jenkins: v1.5.0
    - ripienaar/puppet-module-data: 0.0.4
    - rackerlabs/puppet-repose: 1.6.0
  "git@github.example.com:ops":
    - puppet-profile_crm: 0.1.12
    - puppet-profile_docker: 0.3.1
    - puppet-profile_java_app_server: 0.1.1
    - puppet-profile_jenkins: 1.4.0
    - puppet-repo: 0.3.1
  "git@github.example.com:config-management-modules":
    - puppet-passwordsafe: 0.3.5
```

That is the short form.  Even in short form the syntax of the forge modules directly maps to the standard forge module versioning.  I was trying to support the extended form a bit closer to something like this:

``` yaml
forge:
  "https://forgeapi.puppetlabs.com":
    - bfraser/grafana
    - bmurt/puppet_agent: 0.0.4
    - camptocamp/augeas: latest

git:
  "https://github.com":
    - me/puppet-jenkins:
        branch: plugin-handling
  "git@github.example.com:config-management-modules":
    - puppet-passwordsafe: 0.3.5
```

This greatly reduced how we constantly had to provide the same git uri information over and over again.  That being said, I havent started writing the bits for automatically updating the yaml file and i can see where this may not be the greatest path.  This is part of why I stopped making changes and filed the PR. :)

One thing does concern me... If the forge directive is ignored, do that mean there is no way to use Katello as a forge from r10k?  I'm fine with just collapsing out that level of the dictionary.

gregswift:
I am working on adding a `puppetfile_yaml.mkd` but haven't pushed it yet.

luisfdez:
@gregswift , what about `moduledir`? Are you planning to include it as part of the yaml? That maybe could help with https://github.com/puppetlabs/r10k/issues/133.

Or is your proposal an additional layer that will transform yaml inputs to standard Puppetfile internally?

gregswift:
@luisfdez generally my proposal is just transforming yaml inputs internally. However after reading #133 if you could propose a path for supporting that in this schema i would not be opposed.  That being said, I dont track my moduledirs via Puppetfiles so hadn't really looked at supporting that until I was re-reading the docs last night.

adreyer:
For portability yaml should probably become the standard with tooling to transform ruby puppetfiles into yaml.

I think a more verbose but simpler format is probably better than a slightly more efficient complex one. It's easier for new users to understand without a lot of boiler plate for options they don't use and it's more flexible long term as we discover the need for new options.

```
- name: puppetlabs-apache
  version: "1.0.0"
- name: puppet-iis
  git: "https://github.com/voxpupuli/puppet-iis"
  ref: master
  moduledir: "other/modules"
```

gregswift:
I can get on board with that, however i do think that supporting some kind of method for reducing duplication is handy, such as requested in #543.  I have way more internal modules, just in profiles and internal apps than I do public modules.

danzilio:
This would definitely make some CD workflows simpler. Instead of having to parse or templateize a Puppetfile, you could simply load the YAML and access it as a native data structure. I know more than a few people doing something like this inside the Puppetfile.

raphink:
YAML = no Augeas :cry:

claytono:
As one of the people that gave the talk mentioned above, I think this is a good idea, but I'd be sad if Ruby support ever went away.  We use both today, and our YAML format has lots of extra metadata that isn't represented here.

gregswift:
@raphink thats not strictly true https://github.com/hercules-team/augeas/blob/master/lenses/cobblersettings.aug

and on that note I would definitely love your input on making sure that we define something that could be managed with augeas easily. Some of my ideas on simplification for advanced usage may conflict with that goal.

raphink:
@gregswift there is no garantee that Augeas can actually modify that format properly. YAML is based on indentation levels, which Augeas cannot currently manage, however hard you try. The only thing you can do is a lens that manages a YAML file with _fixed_ indentation, on a known number of indentation levels, but nothing else.

gregswift:
@raphink totally understood.  You end up in a scenario where you should not be editing the file manually.  The whole point around introducing the yml syntax is to facilitate machine control of the file, which augeas can help.  

I'd be open to an alternate format, I was just trying to follow the same route those before me were already using.

raphink:
As much as I dislike JSON, it's much easier to edit.

My concern is really stuff like https://github.com/camptocamp/puppetfile-updater that we use to automate modifications to Puppetfile (so long as there's no Ruby code in there). I'm all for a non-executable format, but if the goal is to use a format for machines only, then JSON is probably better (also, you can easily build web interfaces that understand it).

danzilio:
Why use Augeas instead of a YAML parser?

gregswift:
Augeas with a lense specifically for this file could provide schema validation

raphink:
I totally wouldn't mind JSON at all in fact: https://github.com/camptocamp/r10k-dashboard/blob/master/puppet_dashboard.js#L182-L209

raphink:
@danzilio the point is not Augeas vs YAML, but not closing the door to using Augeas for editing. In my case, I use a Puppetfile as a template, and edit it with Augeas (i.e. without losing comments/newlines, etc.). Any YAML parser that I know of will get rid of all that.

gregswift:
@raphink to clarify, this would not deprecate traditional Puppetfile support. Only add an alternate

raphink:
@gregswift right. You see that currently I have an ugly Javascript function in r10k-dashboard in order to parse the Ruby Puppetfile. A JSON format would definitely help here.

gregswift:
I haven't looked at how Ansible does this (cause they do), but isn't it possible to support both yaml and json with the same data schema? Not to say that json shouldnt be the default, one of my co-workers is blind and so strongly prefers json as well (yaml doesn't parse very well in his tooling)

adreyer:
> @raphink to clarify, this would not deprecate traditional Puppetfile support. Only add an alternate

From our point of view we would like to deprecate though not remove traditional Puppetfile support so this format should account for both human editing, automated manipulation and portability.

Both YAML and JSON have shortcomings here. YAML is less portable than JSON especially with complex keys. JSON can be a pain for humans to write and doesn't allow for comments which can be ciritical in a Puppetfile.

This problem was solved elsewhere in puppet ecosystem, like the puppet-server config, with [HOCON](https://puppetlabs.com/blog/managing-hocon-configuration-files-puppet). There isn't an augeus lens yet but there is (puppetlabs-hocon)[https://forge.puppetlabs.com/puppetlabs/hocon/types] with puppet types for managing hocon. JSON is also valid HOCON so you can just limit yourself to writing JSON if you need to for portability.

raphink:
@adreyer I don't think HOCON is a solution, I think it's more of an issue than YAML and JSON combined (from having tried to write a lens for it and pulled my hair more than I dared to admit).

adreyer:
@raphink There is no augeus lens but the ruby hocon library maintains comments and formatting. It seems like that would be easier than maintaining an augeus lens for puppetfile-updater.

If comments and formatting are the issue isn't JSON ruled out anyway?

raphink:
@adreyer true that JSON is not good for comments.

As for HOCON, I still think a strictly typed format for configuration files is kind overkill and looks more like a programming language to me. In the end, it gets apps even further from being easily configurable than ever (consider ideas like 12factor.net, which is almost completely opposite to HOCON's approach).

gregswift:
sooo... i got a puppet test pilot question related to changing the puppetfile format.

logicminds:
This would break at lot of tooling that others and myself have created.  This would also break compatibility with using puppet-librarian's spec of Puppetfile.  

Also the current Puppetfile spec is very similar with that of the bundler Gemfile spec. New users would dislike "another format" to remember.  

What problem are you trying to solve exactly?  Do you just need a different serialization format?

I can marshal to/from Puppet in just anybody any kind of format I want with a few lines of ruby code.

Example:

```ruby
require 'json'
require 'yaml'

class Puppetfile
   attr_reader :modules

	def initialize(file)
	    @file = file
	    @modules = {}
	     instance_eval(File.read(@file))
	end

	def mod(name, *args)
	     @modules[name] = args.first
	end
end

puts "No Puppetfile given"; exit 1 unless ARGV.first
puts JSON.pretty_generate(Puppetfile.new(ARGV.first).modules)
puts Puppetfile.new('puppetfile_git').modules.to_yaml

**Comment by gregswift** [comment]
Nothing about the original proposals would break existing tooling if that existing tooling was used. However, if someone used the the new format, it would prevent them using other tools until support was extended.  But, that is to be expected.  The purpose behind the proposal was to take what some are already doing and standardize it in a manner that other tools could expect and work with.  The original proposal even suggested that the file have an a separate name, which would also make it obvious. Additionally, the file would be optional and not exist until created, also minimizing confusion.

So why exactly is this a bad thing?

**Comment by logicminds** [comment]
I don't like the idea of telling my clients that there are multiple ways to create a Puppetfile.  It just creates mass confusion.   But having something to auto export/import to something other format would be a plus.

Module dependencies are currently already expressed in metadata.json, .fixutres.yml, r10k Puppetfile, and librarian-puppet Puppetfile.   We don't need a different way, unless they replace all three that Puppet has come up with.

Why not just add the ability to export/import Puppetfile to a serialized format of the user's choice?  Make a simple API and allow them to create a plugin of the serialization format they need.  Then you don't have to keep altering the format for everyone every time external needs change.

**Comment by gregswift** [comment]
I'm sorry that you see this as potentially causing conflicts for you.  

None of the dependency management tools you mention work together already and have different purposes, and almost none of them do any real dependency management anyway. Should it be better? definitely.

This is effectively just a supported de-serializer or importer... The existing methods those of us doing this are using just means our `Puppetfile` is the deserializer of the `Puppetfile.yml`.  Since the `Puppetfile` is executable code, that is allowed.  And it already breaks other things when we use it.

Embracing a community model and determining the best way to make it a built in is all we are talking about doing here.

**Comment by logicminds** [comment]
Lets make a simple pluggable Import/Export API to do this stuff.

**Comment by gregswift** [comment]
i'm curious how you see a simple pluggable import/export api working?  What would a workflow be like?  

(edited cause it was a separate incomplete thought)

**Comment by logicminds** [comment]
I created a new issue #710 around the pluggable API with ideas,and thoughts  gathered from this PR.

File: lib/r10k/puppetfile.rb [hunk 1]
```diff
@@ -55,6 +62,52 @@ def load!
     raise R10K::Error.wrap(e, "Failed to evaluate #{@puppetfile_path}")
   end
 
+  def load_yaml!
+    dsl = R10K::Puppetfile::DSL.new(self)
+    require 'yaml'
+    conf = YAML::load_file(@puppetfile_yaml_path)
```
danielparks, line 68: Security: this needs to use `safe_load`.

logicminds:
@gregswift I propsed #710.  But didn't get any feedback.  Assuming this issue is dead now.

gregswift:
Makes me sad to consider this a dead issue.  I would really like to see this moved forward, and have some co-workers actively working on improving our automation around puppetfiles this quarter. They'll be using our current YAML setup.

I did just go give some feedback to #710, which I thought I had already. However, it seems like trying to boil the ocean. Taking a full revamp of how r10k works isnt a bad idea, but is definitely a significant change.

Before I re-state my format opinion, let me prefix it with: If someone official will just make the call I don't care. There is a benefit to having a readily accessible data format for automation tooling.

Considering how the conversation devolved on format I have to say I think the YAML/JSON support with classic Puppetfile support is probably the best path. 
* Any other format still has the option of existing if anyone wants to customize their Puppetfile
* JSON/YAML are fairly well supported in most languages and is considered very accessible
* It has traction in the community from people already implementing a version of it. Heck I'd still love to see @claytononeill 's schema.  At the end of the day verbosity and simplicity is probably the best path. 

As for having too many options, we already do. We can have a classic Puppetfile or write a script to make automated management of the Puppetfile easier.  By embracing a simple YAML/JSON setup we can take the approach of "If we just want to use Puppetfile, just use it. If we need automation or advanced usage, we also support this schema.".  The [r10k Puppetfile doc](https://github.com/puppetlabs/r10k/blob/master/doc/puppetfile.mkd) doesnt even mention the Rakefile aspect.

Although, I guess we can close these tickets and all keep just running a custom internal processor Puppetfile with a DIY as the recommended model.

logicminds:
I have submitted #763 which if/when merged would allow one to implement this PR as a r10k formatter plugin for or as a completely separate gem.

binford2k:
Hi @gregswift! Thanks for the PR and apologies for the long delay here. Using this and other ideas as inspiration, @reidmv has actually implemented a full suite of YAML source options. You can read more about it here, https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#environment-modules

PR #566, opened 2016-02-03
(RK-207) Method cleanup for R10K::API and R10K::API::Git

scotje:
I apologize in advance to anyone who chooses to review these changes.

I will squash before merge.

### DIFF
--- a/lib/r10k/api.rb
+++ b/lib/r10k/api.rb
@@ -1,4 +1,4 @@
-require 'r10k/api/envmap_builder'
+require 'r10k/api/modules_array_builder'
 require 'r10k/api/git'
 
 require 'r10k/logging'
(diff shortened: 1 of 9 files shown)


lib/r10k/api.rb
```diff
@@ -88,72 +91,78 @@ def parse_puppetfile(io_or_content)
...
         begin
-          version = git.rev_parse(branch_name, git_dir: cachedir)
+          commit_sha = git.resolve_commit(git_dir, branch_name)
         rescue R10K::Git::GitError => e
           raise RuntimeError.new("Unable to resolve branch name '#{branch_name}' to a Git commit: #{e.message}")
         end
 
         # TODO: figure out how to pass through base_path option
-        puppetfile = get_puppetfile(:git, cachedir, version)
+        puppetfile = get_puppetfile(source version)
```
andersonmills: Missing comma? We must not have a test for this?

lib/r10k/api.rb
```diff
@@ -88,72 +91,78 @@ def parse_puppetfile(io_or_content)
...
     # Creates a resolved environment hashmap representing the actual state of the Puppet environment found at the given path.
     #
     # @param path [String] Path on disk of the Puppet environment to be inspected.
-    # @param moduledir [String] The path, relative to the environment path, where the modules are deployed.
+    # @options opts [String] :moduledir The path, relative to the environment, where modules are deployed. (Default: "modules")
     # @return [Hash] A hashmap representing the actual state of the environment found at path.
     # @raise RuntimeError
-    def parse_deployed_env(path, moduledir="modules")
+    def parse_deployed_env(path, opts={})
+      moduledir = opts[:moduledir] || "modules"
```
andersonmills: "modules" here feels like it should be a constant, since it's used in a couple of places in the class.

lib/r10k/api.rb
```diff
@@ -379,19 +393,23 @@ def resolve_module(module_name, env_map, opts={})
...
     # @option opts [String] :cachedir Base path where caches are stored.
+    # @option opts [Hash] :forge Additional options to control interaction with a Puppet Forge API implementation.
+    # @option opts [Hash] :git Additional options to control interaction with remote Git repositories.
+    # @option opts [String] :moduledir The path, relative to the environment, where modules are deployed. (Default: "modules")
     # @option opts [Boolean] :clean Remove untracked files after write.
     # @return [true] Returns true on success, raises on failure.
-    # @raise [RuntimeError] Something bad happened!
+    # @raise [RuntimeError]
     def write_environment(env_map, path, opts={})
+      moduledir = opts[:moduledir] || "modules"
```
andersonmills: The other place "modules" should be replaced by a constant.

lib/r10k/api.rb
```diff
@@ -496,44 +503,50 @@ def write_module(module_name, env_map, path, opts={})
...
+    #
+    # @param env_name [String] Name of environment to deploy, including prefix if applicable.
+    # @param basedir [String] Path on disk into which the given environment should be deployed. (E.g. "/etc/puppetlabs/code-staging/environments")
+    # @option opts [String] :cachedir Base path where caches are stored.
+    # @option opts [Hash] :forge Additional options to control interaction with a Puppet Forge API implementation.
+    # @option opts [Hash] :git Additional options to control interaction with remote Git repositories.
+    # @option opts [Boolean] :purge Whether or not to purge unmanaged modules in the given environment path after deploy. Default: false
+    # @return [true] Returns true on success, raises on failure.
+    # @raise [RuntimeError]
     def deploy_environment(env_name, basedir, sources, opts={})
```
andersonmills, line 536: sources is not documented here.

lib/r10k/git/shellgit.rb
```diff
@@ -53,26 +74,22 @@ def clean(opts = {})
...
+        cmd_opts = {
+          raise_on_fail: false,
+        }
+
+        cmd = ["clone", nil, "--mirror", remote, local]
+
         if opts[:bare]
-          cmd << "--bare"
+          cmd[1] = "--bare"
         end
```
andersonmills, line 90: What's the reason for this being different than the others in the command construction?

lib/r10k/api.rb
scotje: kind of surprised that's not a parse error, maybe the commas are optional in Ruby?

scotje: guess it's just an argument error at runtime

lib/r10k/api.rb
```diff
@@ -88,72 +91,78 @@ def parse_puppetfile(io_or_content)
         parser.instance_eval(io_or_content)
       end
 
-      return { :modules => builder.build }
+      return builder.build
```
andersonmills, line 94: This caused me to go look at EnvmapBuilder, and I'll admit that it bothers me that EnvmapBuilder.build does _not_ build an Envmap. It builds an array of modules. That really feels like it needs to be named .build_modules.

lib/r10k/git/shellgit.rb
scotje, line 90: I was trying to cleverly make sure the opts were getting added before the args, although git doesn't really seem to care where they appear. Using the nil value and the numeric index is pretty kludgy though, so I'm switching this to use Array#insert instead. (Even though Array#insert is one of those annoying Ruby methods that both mutates and returns the receiver.)

scotje:
OK, fixes pushed for all the comments.

scotje:
Still planning to squash, but that might get me killed if I did it before the fixup commits could be reviewed individually.

andersonmills:
:+1:

happy dance

PR #568, opened 2016-02-08
(RK-208) Namespace R10K::API methods

scotje:


andersonmills:
:+1: This sucker's big, and to the best of my knowledge it organizes all the things. (I did look for every method in the original api.rb.) As we increase test coverage and usage of this we may find a thing or two that we missed, but I think this looks good, and will allow us to move forward.

PR #570, opened 2016-02-11
(RK-206) add ssh transport to rjgit

andersonmills:
This commit uses the new ssh transport functionality of rjgit to allow
users to use ssh to connect to remote git repositories.

### DIFF
--- a/lib/r10k/git/rjgit.rb
+++ b/lib/r10k/git/rjgit.rb
@@ -66,14 +66,12 @@ module R10K
           raise R10K::Git::GitError.new("CloneCommand requires that local parent directory (#{local_parent}) already exist.")
         end
 
-        if opts[:private_key] || opts[:username]
-          raise NotImplementedError, "RJGit does not support SSH transport."
-        end
-
         clone_opts = {
           branch: :all,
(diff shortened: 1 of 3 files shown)


lib/r10k/git/rjgit.rb
```diff
@@ -66,14 +66,12 @@ def clone(remote, local, opts={})
...
-        if opts[:private_key] || opts[:username]
-          raise NotImplementedError, "RJGit does not support SSH transport."
-        end
-
         clone_opts = {
           branch: :all,
           is_bare: opts[:bare],
         }
+        clone_opts = clone_opts.merge(private_key_file: opts[:private_key]) if opts[:private_key]
+        clone_opts = clone_opts.merge(username: opts[:username]) if opts[:username]
```
scotje: Is the merge approach for consistency with somewhere else in the code? seems like assigning directly into a key of clone_opts (`clone_opts[:private_key_file] = opts[:private_key] if opts[:private_key]`) would be a little simpler (ditto for fetch_opts below)

scotje:
:+1:

PR #571, opened 2016-02-12
(RK-220) Fail early when git private key is unreadable

adrienthebo:
When r10k was using rugged and Git SSH remotes with pubkey
authentication, r10k didn't check if the specified SSH key was readable
and let rugged do that checking. However rugged's error messaging around
this was less than clear when the private key was missing or unreadable
which users found confusing. This commit remedies the error messaging by
doing the check inside of r10k and failing early with a more readable
message.


scotje:
:+1: 

Will merge is Travis ever fixes itself...

scotje:
Manually running the spec tests on all the rubies....

scotje:
OK, tested in JRuby 1.7.22, and MRI 1.9.x, 2.0.0-p353, 2.1.6, and 2.2.0.

PR #572, opened 2016-02-12
(RK-154) Move repository config from a hash to a list.

adreyer:
We may eventually want an ordering to the repository specific configs.
Also it's painful to work with complex YAML keys with some tools.

### DIFF
--- a/lib/r10k/git.rb
+++ b/lib/r10k/git.rb
@@ -135,5 +135,9 @@ module R10K
     def_setting_attr :private_key
     def_setting_attr :username
     def_setting_attr :repositories, {}
+
+    def self.get_repo_settings(remote)
+      self.settings[:repositories].find {|r| r[:remote] == remote }
+    end
   end
 end
--- a/lib/r10k/settings.rb
+++ b/lib/r10k/settings.rb
@@ -30,18 +30,19 @@ module R10K
         }),
 
         Definition.new(:repositories, {
-          :desc => "Repository specific configuration.",
-          :default => {},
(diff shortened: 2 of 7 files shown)


lib/r10k/git.rb
```diff
@@ -135,5 +135,11 @@ def self.reset!
     def_setting_attr :private_key
     def_setting_attr :username
     def_setting_attr :repositories, {}
+
+    def self.get_repo_settings(remote)
+      logger.error self.settings[:repositories]
```
scotje: remove

lib/r10k/git.rb
```diff
@@ -135,5 +135,11 @@ def self.reset!
...
+
+      self.settings[:repositories].select {|r| r[:remote] == remote }.first
```
scotje: can use `detect` or `find` here

lib/r10k/settings.rb
```diff
@@ -30,18 +30,20 @@ def self.git_settings
...
-            # but in this case it makes sense to have string hash keys. This normalization undoes the
-            # hash key symbolization.
-            repositories.inject({}) do |retval, (key, value)|
-              retval[key.to_s] = value
+        :desc => "Repository specific configuration.",
+        :default => [],
+        :normalize => lambda do |repositories|
+        # The config file loading logic recursively converts hash keys that are strings to symbols,
+        # but in this case it makes sense to have string hash keys. This normalization undoes the
+        # hash key symbolization.
```
scotje: update comment to reflect new behavior

PR #573, opened 2016-02-17
(RK-209) Implement forge download, cache and install

andersonmills:
This commit implements download, cache and install for forge modules.

### DIFF
--- a/lib/r10k/api/modules.rb
+++ b/lib/r10k/api/modules.rb
@@ -3,6 +3,8 @@ require 'r10k/api/util'
 
 require 'r10k/logging'
 
+require 'puppet_forge'
+
 module R10K
   module API
     # Namespace containing R10K::API methods that interact primarily with one or more modules.
@@ -123,12 +125,19 @@ module R10K
             git.clean(path, git_dir: git_dir, force: true)
           end
         when :forge
-          tarball_cachedir = cachedir_for_forge_module(mod[:source], opts[:cachedir])
+          release_slug = release_slug_from_module_slug_version(mod[:source], mod[:resolved_version])
+          cachedir = opts[:cachedir] 
+          unpackdir = File.join(cachedir, 'unpack')
+          if !File.directory?(unpackdir)
+            FileUtils.mkdir_p(unpackdir)
+          end
 
-          # TODO: cache tarball
+          forge_opts = opts[:forge] || {}
+          release_tarball = get_cached_forge_release(release_slug, cachedir, false, forge_opts)
 
-          # TODO: use PuppetForge::Unpacker?
-          raise NotImplementedError
+          tmpdir = Dir.mktmpdir(release_slug, unpackdir)
+          PuppetForge::Unpacker.unpack(release_tarball, path, tmpdir)
+
         else
           raise NotImplementedError
         end
@@ -162,12 +171,11 @@ module R10K
           if !mod[:deployed_version]
             resolve_to = :latest
           else
-            resolve_to = mod[:deployed_version]
+            # this is the short circuit where the deployed version and the resolved version are the same and no forge search is required
+            mod[:resolved_version] = mod[:deployed_version]
+            return mod
           end
         end
-
-        # FIXME: if somehow they have a deployed but unpinned module whose version doesn't exist on Forge, this will
-        # currently fail to resolve
 
         begin
           # TODO: Filter out deleted releases?
@@ -195,6 +203,53 @@ module R10K
       end
       private_class_method :resolve_forge_module
 
+      # Cache a specific module release tarball from the Puppet Forge.
+      #
+      # @param release_slug [String] Hyphen separated namespace, module name, and release of the module to be cached or updated. (E.g. "puppetlabs-apache")
+      # @param force [Boolean] Whether or not to overwrite an existing release tarball if present.
+      # @option opts [String] :cachedir Base path where caches are stored.
+      # @option opts [Hash] :forge Additional options to control interaction with a Puppet Forge API implementation.
+      # @return [String] Returns the full path to the tarball on success, raises on failure.
+      # @raise [RuntimeError]
+      # @api private
+      def self.get_cached_forge_release(release_slug, cachedir, force = false, forge_opts = {})
+        module_slug = module_slug_from_release_slug(release_slug)
+        tarball_cachedir = cachedir_for_forge_module(module_slug, cachedir)
+
+        # TODO: Implement forge options, probably with supporting code in puppet_forge gem.
+        if !forge_opts.empty?
+          raise NotImplementedError "Updating the forge cached modules does not support forge options: #{forge_opts}"
+        end
+
+        if !File.directory?(tarball_cachedir)
+          FileUtils.mkdir_p(tarball_cachedir)
+        end
+
+        final_destination = File.join(tarball_cachedir, release_slug + ".tar.gz")
+
+        if !File.exists?(final_destination) || force
+
+          Dir.mktmpdir(release_slug, tarball_cachedir) do |tmpdir|
+            tmp_destination = File.join(tmpdir, release_slug + ".tar.gz")
+
+            begin
+              PuppetForge::Release.find(release_slug).tap do |release|
+                release.download(Pathname(tmp_destination))
+                release.verify(Pathname(tmp_destination))
+              end
+            rescue PuppetForge::ReleaseNotFound, PuppetForge::ReleaseForbidden => e
+              raise RuntimeError.new, e.message, e.backtrace
+            rescue PuppetForge::V3::Release::ChecksumMismatch => e
+              raise RuntimeError.new, "The checksum of the downloaded tarball does not match. #{e.message}", e.backtrace
+            end
+
+            FileUtils.mv(tmp_destination, final_destination, force: true)
+          end
+        end
+
+        return final_destination
+      end
+      private_class_method :get_cached_forge_release
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/r10k/api/modules.rb
```diff
@@ -123,12 +125,22 @@ def write_module(module_name, env_map, path, opts={})
...
 
-          # TODO: cache tarball
+          unpack_tmpdir = cachedir_for_forge_module([release_slug, 'unpack'].join('-'), cachedir)
+
+          release_tarball = get_cached_forge_release(release_slug, cachedir, false, opts[:forge])
+
+          ## Unpack the files to a given directory
+          # @raise RuntimeError if it fails to extract the contents of the release tarball
+          begin
+            Dir.mktmpdir(release_slug, unpack_tmpdir) do |tmpdir|
```
adreyer: Where does this end up putting stuff? Can we simplify it at all.

lib/r10k/api/util.rb
```diff
@@ -31,6 +31,14 @@ def cachedir_for_forge_module(module_slug, cachedir=nil)
         return File.join(cachedir, 'forge', module_slug)
       end
 
+      def module_slug_from_release_slug(release_slug)
```
adreyer, line 34: Are we sure we won't get '/' modules at this point?

lib/r10k/api/modules.rb
```diff
@@ -195,6 +207,53 @@ def self.resolve_forge_module(mod, opts={})
...
+      # @option opts [Hash] :forge Additional options to control interaction with a Puppet Forge API implementation.
+      # @return [String] Returns the full path to the tarball on success, raises on failure.
+      # @raise [RuntimeError]
+      # @api private
+      def self.get_cached_forge_release(release_slug, cachedir, force = false, forge_opts = {})
+        module_slug = module_slug_from_release_slug(release_slug)
+        tarball_cachedir = cachedir_for_forge_module(module_slug, cachedir)
+
+        # TODO: Implement forge options, probably with supporting code in puppet_forge gem.
+        if !forge_opts.empty?
```
adreyer, line 220: Is there a ticket for this?

lib/r10k/api/modules.rb
andersonmills: Currently, it goes to something like
/tmp/forge/gozar-amodule-0.1.0-unpack/gozar-amodule-0.1.020160217-31443-1ctpusx
Where "/tmp" is the specified cachedir.

I think `Dir.mktmpdir('', unpack_tmdir)` would be just as clear. Then it would be
/tmp/forge/gozar-amodule-0.1.0-unpack/20160217-31443-vr94g9

lib/r10k/api/modules.rb
andersonmills, line 220: RK-218.

lib/r10k/api/util.rb
andersonmills, line 34: This is not the only place where we need to handle /'s. RK-225 is created to asses and fix the slash-dash issue.

lib/r10k/api/modules.rb
```diff
@@ -123,12 +125,22 @@ def write_module(module_name, env_map, path, opts={})
...
+
+          release_tarball = get_cached_forge_release(release_slug, cachedir, false, opts[:forge])
+
+          ## Unpack the files to a given directory
+          # @raise RuntimeError if it fails to extract the contents of the release tarball
+          begin
+            Dir.mktmpdir(release_slug, unpack_tmpdir) do |tmpdir|
+              PuppetForge::Unpacker.unpack(release_tarball, path, tmpdir)
+            end
+          rescue
```
adrienthebo: Why are we swallowing all exceptions here?

lib/r10k/api/modules.rb
```diff
@@ -195,6 +207,53 @@ def self.resolve_forge_module(mod, opts={})
...
+      # @return [String] Returns the full path to the tarball on success, raises on failure.
+      # @raise [RuntimeError]
+      # @api private
+      def self.get_cached_forge_release(release_slug, cachedir, force = false, forge_opts = {})
+        module_slug = module_slug_from_release_slug(release_slug)
+        tarball_cachedir = cachedir_for_forge_module(module_slug, cachedir)
+
+        # TODO: Implement forge options, probably with supporting code in puppet_forge gem.
+        if !forge_opts.empty?
+          raise NotImplementedError "Updating the forge cached madules does not support forge options: #{forge_opts}"
```
adrienthebo: "madules"

adrienthebo:
Travis-CI isn't reporting the status of this build on this page but appears to be running jobs; I'm going to do some poking to figure out what's going on.

lib/r10k/api/modules.rb
andersonmills: That was the leftover of actually catching the RuntimeError first. Thanks for catching that.

lib/r10k/api/modules.rb
andersonmills: Thonks!

lib/r10k/api/modules.rb
```diff
@@ -123,12 +125,19 @@ def write_module(module_name, env_map, path, opts={})
...
-          tarball_cachedir = cachedir_for_forge_module(mod[:source], opts[:cachedir])
+          release_slug = release_slug_from_module_slug_version(mod[:source], mod[:resolved_version])
+          cachedir = opts[:cachedir]
 
-          # TODO: cache tarball
+          unpack_tmpdir = cachedir_for_forge_module([release_slug, 'unpack'].join('-'), cachedir)
+
+          release_tarball = get_cached_forge_release(release_slug, cachedir, false, opts[:forge])
+
+          begin
```
adrienthebo: This `begin/end` block seems to be meaningless - there's no ensure or rescue statement so it doesn't seem to do anything. Am I missing something?

lib/r10k/api/modules.rb
```diff
@@ -123,12 +125,19 @@ def write_module(module_name, env_map, path, opts={})
...
+            Dir.mktmpdir(release_slug, unpack_tmpdir) do |tmpdir|
```
adrienthebo: The docs for `Dir.mktmpdir` indicate that the block form will remove the tmpdir before returning:

```
If a block is given, it is yielded with the path of the directory. The
directory and its contents are removed using FileUtils.remove_entry before
Dir.mktmpdir returns. The value of the block is returned.
```

Is this using a tmpdir to unpack and then copying the contents from the tmpdir to the final location?

andersonmills: Yep. That _is_ what this is doing.

lib/r10k/api/modules.rb
```diff
@@ -195,6 +202,53 @@ def self.resolve_forge_module(mod, opts={})
...
+          Dir.mktmpdir(release_slug, tarball_cachedir) do |tmpdir|
+            tmp_destination = File.join(tmpdir, release_slug + ".tar.gz")
+
+            begin
+              PuppetForge::Release.find(release_slug).tap do |release|
+                release.download(Pathname(tmp_destination))
+                release.verify(Pathname(tmp_destination))
+              end
+            rescue PuppetForge::ReleaseNotFound, PuppetForge::ReleaseForbidden => e
+              raise RuntimeError.new e.message
```
adrienthebo: This should include the backtrace from the original error; otherwise this will generate a new backtrace and may hide where the error actually came from. https://github.com/puppetlabs/r10k/blob/master/lib/r10k/errors.rb#L10-L22 implements exception wrapping which might come in handy here.

adrienthebo:
I have one more minor comment, but things seem generally good to me. :+1: for me, though @adreyer may have additional concerns.

lib/r10k/api/modules.rb
```diff
@@ -123,12 +125,17 @@ def write_module(module_name, env_map, path, opts={})
             git.clean(path, git_dir: git_dir, force: true)
           end
         when :forge
-          tarball_cachedir = cachedir_for_forge_module(mod[:source], opts[:cachedir])
+          release_slug = release_slug_from_module_slug_version(mod[:source], mod[:resolved_version])
+          cachedir = opts[:cachedir]
 
-          # TODO: cache tarball
+          unpack_tmpdir = cachedir_for_forge_module([release_slug, 'unpack'].join('-'), cachedir)
```
adreyer: perhaps this should just be `join(cachedir, unpack)` or `join(cachedir,unpack,release_slug)`. This shouldn't be considered API so I guess it doesn't block the merge.

lib/r10k/api/modules.rb
```diff
@@ -123,12 +125,19 @@ def write_module(module_name, env_map, path, opts={})
...
+          release_slug = release_slug_from_module_slug_version(mod[:source], mod[:resolved_version])
+          cachedir = opts[:cachedir] 
+          unpackdir = File.join(cachedir, 'unpack')
+          if !File.directory?(unpackdir)
+            FileUtils.mkdir_p(unpackdir)
+          end
 
-          # TODO: cache tarball
+          forge_opts = opts[:forge] || {}
+          release_tarball = get_cached_forge_release(release_slug, cachedir, false, forge_opts)
```
scotje, line 136: Do we need to set the force option from opts? I guess it depends what we are "forcing"

lib/r10k/api/modules.rb
```diff
@@ -162,13 +171,13 @@ def self.resolve_forge_module(mod, opts={})
           if !mod[:deployed_version]
             resolve_to = :latest
           else
-            resolve_to = mod[:deployed_version]
+            # this is the short circuit where the deployed version and the resolved version are the same and no forge search is required
+            #resolve_to = mod[:deployed_version]
```
scotje: can get rid of the commented line

lib/r10k/api/modules.rb
andersonmills, line 136: Through in-person discussion, we decided no.

PR #575, opened 2016-02-24
(RK-218) stateless forge opts to puppet forge

andersonmills:
This commit changes to using puppet_forge's new stateless requests.

This is currently based on #573 and in parallel with forge-ruby#26


scotje:
:+1: when Travis is green

PR #584, opened 2016-04-13
(RK-229) Better proxy configuration for PuppetForge.

scotje:


andersonmills:
:+1:  When travis finishes his oatmeal.

oats

PR #592, opened 2016-04-24
(RK-242) test for PE only modules with proxy

zreichert:


integration/tests/basic_functionality/proxy_with_pe_only_module.rb
```diff
@@ -0,0 +1,128 @@
...
+create_remote_file(master, "#{git_environments_path}/Puppetfile", 'mod "ztr-peonly"')
+
+step 'Push Changes'
+git_add_commit_push(master, 'production', 'add Puppetfile', git_environments_path)
+
+step 'Install and configure squid proxy'
+on(master, install_squid)
+
+step 'turn off the firewall'
+on(master, puppet("apply -e 'service {'iptables' : ensure => stopped}'"))
```
scotje, line 108: just out of curiosity, why do we turn off iptables?

scotje:
I had one question just for my own enlightenment, overall looks good though! 👍

PR #591, opened 2016-04-22
(RK-236) Add git proxy settings support to r10k.yaml.

scotje:

PR #593, opened 2016-05-02
(RK-237) Manage proxy env for Git clone/fetch operations.

scotje:
To be merged with #591 to make the new settings have an actual effect. May need a quick rebase after 591 is merged.


andersonmills:
:+1:   Tested this by hand and got it working (and not working when expected to break).

PR #596, opened 2016-05-16
(RK-243) Refuse to generate rugged credentials more than 50 times.

scotje:
In libgit2/rugged 0.24.0 the callback for fixed credentials continues to be called until either it raises an error or the server gives up. Since neither was happening, invalid credentials were leading to an
infinite retry loop.


scotje:
It's unclear to me if there are situations where the credentials callback instance's #call would be legitimately called more than once in normal r10k operations.

scotje:
Bumped the limit up to 50 to be roughly consistent with how many times SSH auth will retry.

scotje:
See https://github.com/libgit2/rugged/issues/592

andersonmills:
:+1: I'm happy with this. Tested by hand and watched the infinite loop and the 50 count.

PR #602, opened 2016-05-31
(RK-222) Add :install_path option to non-Forge Puppetfile entries

scotje:
I'm going to get some unit tests added to this before merging, but there wasn't any existing test coverage so the tests should still be passing.


andersonmills:
:+1: up to here. Waiting for tests for the whole thumb.

scotje:
I should probably update the README and stuff as well...

andersonmills:
:+1: I'm so happy! But even more happy to see the updated docs.

andersonmills:
👍 Happy with the docs, too.

PR #605, opened 2016-06-09
(CODEMGMT-731) Add maintainers to README

austb:


README.mkd
```diff
@@ -103,3 +103,14 @@ Contributors
 ------------
 
 Please see the CHANGELOG for a listing of the (very awesome) contributors.
+
+git@github.com:puppetlabs/r10k.git
```
scotje: Is there a reason for the git URI here?

austb: Absolutely none... I'm now confused how that ended up there.

PR #611, opened 2016-06-20
(MAINT) update regex used in proxy tests

zreichert:


scotje:
👍 

Did you figure out why this was different now? Just a difference between shellgit and rugged?

integration/tests/git_source/HTTPS_proxy_and_git_source.rb
```diff
@@ -64,7 +64,7 @@
 
 #test
 on(master, "#{r10k_fqp} deploy environment -p", :accept_all_exit_codes => true) do |r|
-  regex = /Couldn't resolve proxy 'ilovecatvideos\.com'/
+  regex = /(Couldn't|Could not) resolve proxy.*ilovecatvideos\.com/
```
Lemonne, line 67: The 'ferritsarebest' regex has ignore case, but this regex does not. I think you also meant to have ignore case here?

Lemonne, line 67: This looks super rad!

PR #612, opened 2016-06-22
Feature support custom puppetfile name

elconas:
We need a way to specify a custom puppetfile name (e.g. Puppetfile.r10k) as Puppetfiles for r10k and librarian-puppet are not compatible and we use both tools for Module management. Therefore we need the possibility to use a different Puppetfile in our repos.

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -1,5 +1,6 @@
 require 'pathname'
 require 'r10k/module'
+require 'r10k/settings'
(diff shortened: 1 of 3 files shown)


scotje:
This seems reasonable, but should it maybe be a per-source setting rather than global?

elconas:
Refactored in https://github.com/puppetlabs/r10k/pull/651, so once https://github.com/puppetlabs/r10k/pull/651 is accepts, this can be closed.

PR #613, opened 2016-06-22
(MAINT) Fixup test regex for unmanaged content removal logging.

scotje:
This was changed in RK-246 I think, the logging now happens after deletion instead of before and the message was updated to reflect that.


andersonmills:
:+1: All the way!

PR #616, opened 2016-06-27
Check version while parsing Puppetfile

patrobinson:
[Librarian Puppet supports constraints](https://github.com/voxpupuli/librarian-puppet/blob/master/features/install/forge.feature#L88) in the Puppetfile, but r10k doesn't and throws a rather obscure error:

```
ERROR    -> bad URI(is not URI?): /v3/files/puppetlabs-concat-< 2.0.0.tar.gz
```

This Pull Request seeks to fix that by ensuring the argument provided for a forge module is a valid Semantic Version number (using `semantic_puppet`) or `:latest` or it's absent.

Along the way I couldn't figure out what `args` would look like without a version, `module_spec` led me to believe it would be an empty array, but that is incorrect. So I've attempted to better define it with an example spec that has some Puppetfile fixtures for forge with and without args.

### DIFF
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -17,7 +17,7 @@ class R10K::Module::Forge < R10K::Module::Base
   end
 
   def self.valid_version?(expected_version)
-    expected_version == :latest || expected_version.empty? || SemanticPuppet::Version.valid?(expected_version)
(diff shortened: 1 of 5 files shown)


patrobinson:
ping @scotje

scotje:
@patrobinson I had to fixup one test against the latest master and I went ahead and squashed everything down into one commit (with your name still attached to it). I'm going to merge this in with #632.

Thanks for your contribution!

PR #620, opened 2016-07-15
(RK-258) Make symlinks in Forge modules warn instead of raising.

scotje:
This should also have the effect of letting r10k still exit 0 even if symlinks are encountered in a module.


austb:
+1

PR #623, opened 2016-07-22
Fix minor issues

scotje:
Fixes a couple minor issues that surfaced during acceptance tests.


austb:
👍  Looks good, I still wish we had something to warn us if we use # instead of %. It seems like that could create a bit of a headache fairly frequently.

PR #624, opened 2016-07-28
Maint/master/codemgmt 771 update error assertions

zreichert:


austb:
👍 Looks good. There are two commits in this PR, but I'm not sure if you want to squash the commits or leave them separate.

PR #630, opened 2016-08-05
(MAINT) fix flaky purging test

zreichert:
This block is redundant.  Later in the test we check that instrumented files are purged or not purged depending on location.


andersonmills:
:+1:

PR #635, opened 2016-08-11
(MAINT) Clarify docs around environment purging with Puppetfile errors.

scotje:


austb:
👍  New wording is much more clear.

PR #640, opened 2016-08-19
(RK-265) Make 'puppetfile install' skip content with local modifications.

scotje:


andersonmills:
:+1: I saw it work and it was beautiful!

Issue #647, opened 2016-09-02
r10k doesn't identify itself as the cause when exiting on exception


GeoffWilliams:
## Summary

When r10k exits on exception, it doesn't identify that the program generating the error was R10K
## Expected behaviour

r10k should identify itself when exiting on error, eg:

```
Error running r10k: detailed error message...
```
## Actual behaviour

Users receive errors that are completely generic, eg:

```
Error while running: #<NoMethodError: undefined method `keys' for false:FalseClass>
```
## Steps to reproduce

See #645 
## Why this is a problem

r10k tends to get embedded in other systems or be deeply nested inside other systems/puppet code.  The current style of error messages are easy look like could come from any Ruby code on the system and make debugging systems end-to-end very difficult

scotje:
Resolved by #694

PR #648, opened 2016-09-02
[puppetlabs/r10k] (GH-647) identify r10k on error

GeoffWilliams:
Incorporate the program name into the message when exiting on exception
eg "error while running r10k: ..." not just "error while running: ..."


scotje:
Is there a situation where people are renaming the r10k script or could we just hardcode 'r10k' in the error strings?

scotje:
Closed in favor of #694

PR #652, opened 2016-09-13
Update cri dependency to 2.7.0

mmoll:
tests are still :green_heart: for me


mmoll:
BTW, what was the reason to pin so stricly? would a PR with `~> 2.6` be acceptable?

mmoll:
closing in favour of #676

scotje:
Thanks for the contribution!

PR #655, opened 2016-09-14
(MAINT) Various performance optimizations.

scotje:
I did some basic profiling and identified some places where we were making extra comparisons or checking things in an inefficient order. Taken together, these optimizations can make a significant improvement to total deploy times.


andersonmills (approved):
Tested several configurations by hand that would cause this code to be exercised. Across several runs, saw improvements in run time. Yay!

PR #657, opened 2016-09-19
(MAINT) Add more specific logging when local modifications are found.

scotje:
This adds some logging at DEBUG and DEBUG1 when the #dirty? method of
the WorkingRepository classes (for both rugged and shellgit providers)
find local modifications. This allows users to determine what files r10k
thinks are locally modified, as well as how it thinks those files have
been modified (when logging at DEBUG1).


haus:
Is there any opportunity to add some spec tests around this new logging?

scotje:
Yeah I can add some tests.

andersonmills (approved):

PR #663, opened 2016-10-25
Changed r10k.yaml cache option to /var/cache/r10k.

ckabalan:
The /etc filesystem is for host-specific configuration information.
Cache data is not configuration information and is better suited for
/var/cache, per the Linux Foundation's Filesystem Hierarchy Standard.
See "FHS 3.0" section "5.5. /var/cache : Application cache data".

In application of the standard the new setting is: /var/cache/r10k

Link:
- http://refspecs.linuxfoundation.org/fhs.shtml

### DIFF
--- a/doc/dynamic-environments/quickstart.mkd
+++ b/doc/dynamic-environments/quickstart.mkd
@@ -95,7 +95,7 @@ Configure r10k by creating the following directory structure and file `/etc/pupp
 
 ```
 # The location to use for storing cached Git repos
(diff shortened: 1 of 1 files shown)


scotje:
Is this just to make the documentation consistent with the value in `r10k.yaml.example` ?

ckabalan:
I hadn't actually looked at `r10k.yaml.example`, but yes, this would make the quickstart guide consistent with the entry on line 12 (which has the proper location).

scotje:
Thanks for the contribution!

PR #667, opened 2016-11-04
Add support for handling untracked files in Git repositories

seanmil:
These commits add support to both shellgit and rugged to detect and (when force is enabled) remove untracked files from Git repositories.

Also included is a commit which fixes shellgit's checkout function to correctly preserve modified files when force is false, making the behavior consistent with rugged's behavior.

Spec tests are included for both sets of changes.


seanmil:
I discovered some issues that weren't caught by the existing spec tests. Going to close this PR until I can sort them out.

PR #668, opened 2016-11-08
Fix force=false behavior on shellgit checkout

seanmil:
Add spec tests for the git.checkout method covering both
force=true and force=false.

Fix the behavior with shellgit's checkout to behavior according
to the documentation and to be consistent with rugged's behavior.

### DIFF
--- a/lib/r10k/git/shellgit/working_repository.rb
+++ b/lib/r10k/git/shellgit/working_repository.rb
@@ -49,14 +49,14 @@ class R10K::Git::ShellGit::WorkingRepository < R10K::Git::ShellGit::BaseReposito
   # @param ref [String] The git reference to check out
   # @param opts [Hash] Optional hash of additional options.
   def checkout(ref, opts = {})
+    argv = ['checkout', ref]
+
     # :force defaults to true
     if !opts.has_key?(:force) || opts[:force]
-      force_opt = '--force'
-    else
-      force_opt = ''
+      argv << '--force'
     end
 
-    git ['checkout', ref, force_opt], :path => @path.to_s
+    git argv, :path => @path.to_s
   end
 
   def fetch(remote_name='origin')
--- a/spec/shared-examples/git/working_repository.rb
+++ b/spec/shared-examples/git/working_repository.rb
@@ -158,4 +158,25 @@ RSpec.shared_examples "a git working repository" do
       expect(subject.origin).to eq remote
     end
   end
+
+  describe "checking out ref" do
+    before(:each) do
+      subject.clone(remote)
+      File.open(File.join(subject.path, 'README.markdown'), 'a') { |f| f.write('local modifications!') }
+    end
+
+    context "with force = true" do
+      it "should revert changes in managed files" do
+        subject.checkout(subject.head, {:force => true})
+        expect(File.read(File.join(subject.path, 'README.markdown')).include?('local modifications!')).to eq false
+      end
+    end
+
+    context "with force = false" do
+      it "should not revert changes in managed files" do
+        subject.checkout(subject.head, {:force => false})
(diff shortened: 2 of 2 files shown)


scotje:
Thanks for the contribution! I'll try to get this reviewed next week.

lib/r10k/git/shellgit/working_repository.rb
```diff
@@ -49,14 +49,14 @@ def clone(remote, opts = {})
   # @param ref [String] The git reference to check out
   # @param opts [Hash] Optional hash of additional options.
   def checkout(ref, opts = {})
+    options = ['checkout', ref]
```
scotje: could you name this variable `argv` to be consistent with the `clone` method above it?

scotje:
I don't see any actual change in behavior in the diff now, is it possible something was lost in a rebase? Or perhaps I'm just missing it?

lib/r10k/git/shellgit/working_repository.rb
seanmil: Done. I have also rebased against master and pushed both up.

seanmil:
The change in behavior is subtle, but important. Note that I am testing against the `git-1.8.3.1-6.el7_2.1.x86_64` package from CentOS 7. It is possible that different versions of Git won't exhibit this problem. However, I just also tested it on a Fedora 24 system with Git 2.7.4 and it still shows the same behavior.

Currently in the code, when force is not enabled, R10K invokes a git checkout with a trailing empty argument ''. The `force_opt` parameter gets passed in the argv regardless of whether it is set to a non-empty value or not.

For reasons I don't understand, Git behaves quite differently between:
`git checkout <ref>` and `git checkout <ref> ''`. The former does what you would expect. The latter - as best as I can tell - checks out the requested ref but never updates .git/HEAD. This results in a working tree showing modifications (the files from version X, the HEAD ref being compared against from X+1).

Assuming you are on a similar version of Git you can easily demonstrate this behavior:
```$ git --version
git version 2.7.4
$ git clone https://github.com/puppetlabs/puppetlabs-stdlib
Cloning into 'puppetlabs-stdlib'...
remote: Counting objects: 9188, done.
remote: Compressing objects: 100% (64/64), done.
remote: Total 9188 (delta 19), reused 0 (delta 0), pack-reused 9121
Receiving objects: 100% (9188/9188), 1.86 MiB | 1.40 MiB/s, done.
Resolving deltas: 100% (4398/4398), done.
Checking connectivity... done.
$ cd puppetlabs-stdlib
$ git checkout -q 4.13.1
$ cat .git/HEAD
f2492ee916c1c8e0345514045432c4a049674029
$ git status -s
$ git checkout -q 4.13.0
$ cat .git/HEAD 
ef55080c1f32b14a3efec401d4ef8828f052d17b
$ git status -s
$ git checkout -q 4.13.1 ''
$ cat .git/HEAD 
ef55080c1f32b14a3efec401d4ef8828f052d17b
$ git status -s
M  .travis.yml
M  CHANGELOG.md
A  lib/puppet/functions/is_absolute_path.rb
A  lib/puppet/functions/is_array.rb
A  lib/puppet/functions/is_bool.rb
A  lib/puppet/functions/is_float.rb
A  lib/puppet/functions/is_ip_address.rb
...
```

Hope that helps.

seanmil:
Actually, after some further investigation, it sounds like the behavior when passed `''` is documented Git behavior. From the man page:

```
git checkout [-p|--patch] [<tree-ish>] [--] <pathspec>...
       When <paths> or --patch are given, git checkout does not switch branches.
       It updates the named paths in the working tree from the index file or from a named
       <tree-ish> (most often a commit).
```

It looks like Git is taking the empty string passed in by R10K as the first pathspec in that form.

scotje:
Ah, that makes sense!

Thanks for the fix and the good explanation!

PR #669, opened 2016-11-11
(CODEMGMT-702) Update rugged code for rugged behavior

andersonmills:
This commit changes the behavior of a checkout in the rugged provider to
deal with the execute bit behavior (not updating) of rugged/libgit2
around reset-ting to a SHA that the working directory is already at.


andersonmills:
Basically, the issue is that rugged/libgit2 won't update (at least) the execute
bit on a checkout.  And it won't update it on a reset --hard to the same
SHA, apparently.

sync is called from deploy::environment with force = true, and reset
--hard has the same effect on the working directory in the end, so the
checkout isn't necessary.

scotje:
Looks reasonable to me, is there any value it adding a test around this? Maybe an assertion that checkout isn't called on the repo when force is true?

scotje (approved):

PR #680, opened 2016-12-01
(RK-78) Use new :prune option for #fetch in Rugged::BareRepository

scotje:
Prior to rugged 0.24.0, the fetch operation did not support pruning
stale branch references from the local repo. We had implemented a
workaround but now that rugged 0.24.0 is out, we should use the built in
:prune option when fetching.


andersonmills (approved):
+1 once the tests pass.

PR #690, opened 2017-01-04
(RK-284) Replace reachable server with unresolvable server

haus:
Tests are currently failing because ilovecatvideos.com is resolvable but
no available on port 3128, so connections time out. In order to provoke
the desired error state, this commit changes the target server to be
unresolvable.


scotje (approved):

scotje:
Ping @zreichert

zreichert:
👍

PR #691, opened 2017-01-05
Add a new local_directory module

aptituz:
This commit introduces a new module type, which allows to include
modules by copying them from a specified directory.
This can be used to include modules which are updated by different
means, but should be part of this environment.

### DIFF
--- a/doc/puppetfile.mkd
+++ b/doc/puppetfile.mkd
@@ -290,6 +290,43 @@ module.
 For more information see the [FAQ entry](faq.mkd#how-do-i-prevent-r10k-from-removing-modules-in-the-modules-directory-of-my-git-repository)
 on managing internal and external modules in the same directory.
 
+### Local directory
+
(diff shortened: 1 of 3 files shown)


binford2k:
Hm. I'm not convinced that this added configuration overload is worth it, given the existing options. You can already use local modules by either putting them in a `/site-modules` directory, or putting them in `/modules` and protecting them by declaring a `:local` mod. And if you need them in a different directory (as in the "installed by a package manager" case) then you could just symlink to that directory.

I'm going to go ahead and close this for now. If you feel strongly about it, please feel free to reopen and make the case for it. Thanks so much for the contribution!

(also, if you do, we'd need the CLA signed)

aptituz:
@binford2k I don't feel strongly about it. Not after 4 years. ;) Back then, I had a use case and a reason to believe that this is the best way to solve the issue. Thanks for consideration, though.

Issue #659, opened 2016-09-30
Cannot use r10k puppetfile check when a git module is tracking the control repo branch


chriscowley:
When a module entry in my `Puppetfile` contains something like:

```
mod 'profile',
    :git => 'ssh://git@<my-gitlab>/puppet/puppet-profile.git',
    :branch => :control_branch,
    :default_branch => 'master'
```

running `r10k puppetfile check` returns an error with the message:

```
Failed to evaluate /<path-to-controlrepo>/Puppetfile
Original exception:
Cannot track control repo branch from Puppetfile in this context: environment is nil or did not provide a valid ref
```

scotje:
Indeed, I'll see if I can add a condition to `puppetfile check` for this.

scotje:
Created https://tickets.puppetlabs.com/browse/RK-269 to track this.

andersonmills:
#698 has been merged. We should be publishing 2.5.2 soon with the fix.

tuxmea:
this is solved logn ago.
can we close this issue?

PR #698, opened 2017-02-06
(RK-269) Update 'puppetfile' subcommands to handle control repo branch tracking.

scotje:
`default_branch` value will be used if available during `puppetfile install` operation, otherwise command will exit with an error.

`puppetfile check` and `puppetfile purge` should run without issue.


scotje:
Will resolve #659

andersonmills:
Tested this, and it works. :+1:

andersonmills (approved):

PR #708, opened 2017-03-27
fix link for control repo in workflow guide

raarts:
Jeff Malnick moved his blog to technoblogic.io. Link fixed.


andersonmills (approved):
Checked up on this docs change, and this is valid.

PR #719, opened 2017-04-19
(RK-239) Fail fast on Puppetfiles with duplicate module names

ajroetker:
This commit fails fast when loading environments where the Puppetfile
contains duplicate module names.


andersonmills:
Nice!
```
roquefort:~/puppet/src/r10k (PR719)
$ be r10k puppetfile check
Puppetfiles cannot contain duplicate module names. Please remove the remove the duplicates of the following modules:
happiness
```

andersonmills:
I think this code does exactly what is intended. I think it's worth answering the questions in RK-239 before we merge it, though.

It at least needs a rebase.

andersonmills (approved):
Tested and approved.

PR #723, opened 2017-04-25
(maint) Update master branch to 3.0.0.dev.

andersonmills:


lib/r10k/version.rb
```diff
@@ -1,3 +1,3 @@
 module R10K
-  VERSION = '2.6.0.dev'
+  VERSION = '3.0.0.dev'
```
ajroetker, line 2: Does this need to be updated anywhere else in CI/PE?

andersonmills: Not in the previous setup. No idea what you've done recently.

andersonmills: I take that back.

pkg_version in rubygem-r10k.rb in pe-r10k-vanagon.

binford2k:
@justinstoller is this PR even still relevant?

justinstoller:
no

PR #727, opened 2017-04-26
(maint) Fixup duplicate module name error string

ajroetker:
This commit adjusts the error string for duplicate module names by
separating the sentences (for i18n) and space separates the modules with
duplicates instead of newlines.


andersonmills (approved):

andersonmills:
Rockit!

PR #728, opened 2017-05-01
(CODEMGMT-1064) Test pe-r10k project before promoting it.

libbymo:
Prior to this commit the acceptance task would only target a previously packaged version of pe-r10k.  This addition allows for the installation of a dev version of pe-r10k over the install of pe. The acceptance task has been modified to allow for this additional component focused test, vs. a pre-suite test targeting a PEZ build. The ci:test:component task will need to be called by a component configuration to be added to ci-job-configs.

This is related to https://github.com/puppetlabs/ci-job-configs/pull/2840


integration/Rakefile
```diff
@@ -1,13 +1,38 @@
 require 'rototiller'
 
+namespace :ci do
+  namespace :test do
+
+    desc 'Tests at the component level for the pe-r10k project'
+    task :component => [:check_client_tool_env_vars] do
```
libbymo: I only brought in the :component task, and not :package let me know if that's ok. :)

integration/Rakefile
```diff
@@ -1,13 +1,38 @@
...
+pre-suite/30_test_utils.rb,
+EOS
+  common_setup.gsub!("\n", '')
 
   flags = [
       {:name => '--hosts', :default => 'configs/generated', :override_env => 'BEAKER_HOST'},
       {:name => '--keyfile', :default => "#{ENV['HOME']}/.ssh/id_rsa-acceptance", :override_env => 'BEAKER_KEYFILE'},
       {:name => '--load-path', :default => 'lib'},
-      {:name => '--pre-suite', :default => 'pre-suite'},
+      {:name => '--pre-suite', :default => @acceptance_pre_suite || common_setup, :override_env => 'BEAKER_PRE_SUITE'},
```
libbymo, line 35: `BEAKER_PRE_SUITE` is total copy pasta.  And I'm unclear what it is and if it's what we want here.

integration/component/05_install_dev_r10k.rb
```diff
@@ -0,0 +1,37 @@
+test_name "Install PE r10k" do
+
+  step "Install PE r10k" do
+    product = 'pe-r10k'
```
ajroetker: I don't think this gets used?

integration/component/05_install_dev_r10k.rb
```diff
@@ -0,0 +1,37 @@
...
+    client = find_only_one('client')
+
+    install_pe_client_tools_on(client, opts)
+
+  end
+
+end
+
+# taken from beaker-pe/lib/beaker-pe/pe-client-tools/install_helper.rb
+def install_pe_client_tools_on(hosts, opts = {})
```
ajroetker: Could we use `install_pe_r10k_on`? Or even, `install_pe_product_on` and pass in product as an arg?

integration/component/05_install_dev_r10k.rb
```diff
@@ -0,0 +1,37 @@
...
+    product = 'pe-r10k'
+    required_keys = [:puppet_collection, :pe_client_tools_sha, :pe_client_tools_version]
```
ajroetker: Similarly can we rename these vars `:product_sha/:product_version` or `:pe_r10k_sha/:pe_r10k_version` depending on the outcome of my above comment?

integration/Rakefile
ajroetker: I think you need the package task too, let me verify though

integration/component/05_install_dev_r10k.rb
libbymo: You think it's ok to pass product as part of opts?

ajroetker: I'd make it it's own arg just because it's so important for the function.

integration/Rakefile
libbymo: Any verdict on that? And also, not sure where :check_client_tool_env_vars comes from? Will this be passed in to the test call in CI? And should I rename this to be r10k specific?

integration/configs/README.mkd
```diff
@@ -1,7 +0,0 @@
-Config
```
libbymo, line 1: @zreichert could you verify this `integration/configs` was the folder we discussed should be removed?  Also should these be removed as well: https://github.com/puppetlabs/r10k/tree/master/integration/scripts?  Thanks!

PR #729, opened 2017-05-11
(RK-291) Cache Forge module releases.

andersonmills:
This commit adds caching to the forge module releases. This is external
to environments, so will significantly increase the speed of environment
deploys across environments using the same forge module releases.

### DIFF
--- a/lib/r10k/forge/module_release.rb
+++ b/lib/r10k/forge/module_release.rb
@@ -13,6 +13,7 @@ module R10K
 
       def_setting_attr :proxy
       def_setting_attr :baseurl
+      def_setting_attr :cache_root, File.expand_path(ENV['HOME'] ? '~/.r10k/cache': '/root/.r10k/cache')
 
       include R10K::Logging
 
@@ -26,6 +27,14 @@ module R10K
       # @!attribute [rw] download_path
       #   @return [Pathname] Where the module tarball will be downloaded to.
       attr_accessor :download_path
+
+      # @!attribute [rw] tarball_cache_path
+      #   @return [Pathname] Where the module tarball will be cached to.
+      attr_accessor :tarball_cache_path
+
+      # @!attribute [rw] tarball_cache_root
+      #   @return [Pathname] Directory where the module tarball will be cached to.
+      attr_accessor :tarball_cache_root
 
       # @!attribute [rw] unpack_path
       #   @return [Pathname] Where the module will be unpacked to.
@@ -41,9 +50,14 @@ module R10K
         # objects are created in the class instances and thus are not shared with
         # subclasses.
         PuppetForge::V3::Release.conn = PuppetForge::V3::Base.conn
+
         @forge_release = PuppetForge::V3::Release.new({ :name => @full_name, :version => @version, :slug => "#{@full_name}-#{@version}" })
 
-        @download_path = Pathname.new(Dir.mktmpdir) + (@forge_release.slug + '.tar.gz')
+        tarball_name = @forge_release.slug + '.tar.gz'
+        @download_path = Pathname.new(Dir.mktmpdir) + (tarball_name)
+        @tarball_cache_root = Pathname.new(settings[:cache_root]) + (@forge_release.slug + "/tarball/")
+        @tarball_cache_path = @tarball_cache_root + tarball_name
+
         @unpack_path   = Pathname.new(Dir.mktmpdir) + @forge_release.slug
       end
 
@@ -65,34 +79,47 @@ module R10K
         cleanup
       end
 
-      # Download the module release to {#download_path}
+      # Download the module release to {#download_path} and cache to {#tarball_cache_path}
       #
       # @return [void]
       def download
-        logger.debug1 "Downloading #{@forge_release.slug} from #{PuppetForge::Release.conn.url_prefix} to #{@download_path}"
-        @forge_release.download(download_path)
+        if @tarball_cache_path.exist?
+          logger.debug1 "Using cached copy of #{@forge_release.slug} tarball"
+        else
+          logger.debug1 "Downloading #{@forge_release.slug} from #{PuppetForge::Release.conn.url_prefix} to #{@download_path}"
+          @forge_release.download(download_path)
+          FileUtils::mkdir_p(@tarball_cache_root)
+          FileUtils::mv(@download_path, @tarball_cache_path)
+        end
       end
 
-      # Verify the module release downloaded to {#download_path} against the
-      # module release checksum given by the Puppet Forge
+      # Verify the module release cached in {#tarball_cache_path} against the
+      # module release checksum given by the Puppet Forge. On mismatch, remove
+      # the cached copy.
       #
       # @raise [PuppetForge::V3::Release::ChecksumMismatch] The
-      #   downloaded module release checksum doesn't match the expected Forge
+      #   cached module release checksum doesn't match the expected Forge
       #   module release checksum.
       # @return [void]
       def verify
-        logger.debug1 "Verifying that #{download_path} matches checksum #{@forge_release.file_md5}"
-        @forge_release.verify(download_path)
+        logger.debug1 "Verifying that #{@tarball_cache_path} matches checksum #{@forge_release.file_md5}"
+        begin
+          @forge_release.verify(@tarball_cache_path)
+        rescue PuppetForge::V3::Release::ChecksumMismatch => e
+          logger.error "Checksum Mismatch. Not installing #{@forge_release.slug}. Cleaning up #{@tarball_cache_path}."
+          cleanup_cached_tarball_path
+          raise e
+        end
       end
 
-      # Unpack the module release at {#download_path} into the given target_dir
+      # Unpack the module release at {#tarball_cache_path}  into the given target_dir
       #
       # @param target_dir [Pathname] The final path where the module release
       #   should be unpacked/installed into.
       # @return [void]
       def unpack(target_dir)
-        logger.debug1 _("Unpacking %{download_path} to %{target_dir} (with tmpdir %{tmp_path})") % {download_path: download_path, target_dir: target_dir, tmp_path: unpack_path}
-        file_lists = PuppetForge::Unpacker.unpack(download_path.to_s, target_dir.to_s, unpack_path.to_s)
+        logger.debug1 _("Unpacking %{tarball_cache_path} to %{target_dir} (with tmpdir %{tmp_path})") % {tarball_cache_path: tarball_cache_path, target_dir: target_dir, tmp_path: unpack_path}
+        file_lists = PuppetForge::Unpacker.unpack(tarball_cache_path.to_s, target_dir.to_s, unpack_path.to_s)
         logger.debug2 _("Valid files unpacked: %{valid_files}") % {valid_files: file_lists[:valid]}
         if !file_lists[:invalid].empty?
           logger.debug1 _("These files existed in the module's tar file, but are invalid filetypes and were not unpacked: %{invalid_files}") % {invalid_files: file_lists[:invalid]}
@@ -119,6 +146,13 @@ module R10K
       def cleanup_download_path
         if download_path.exist?
           download_path.delete
+        end
+      end
+
+      # Remove the cached module release.
+      def cleanup_cached_tarball_path
+        if tarball_cache_path.exist?
+          tarball_cache_path.delete
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/r10k/forge/module_release.rb
```diff
@@ -13,6 +13,7 @@ class ModuleRelease
 
       def_setting_attr :proxy
       def_setting_attr :baseurl
+      def_setting_attr :cache_root, File.expand_path(ENV['HOME'] ? '~/.r10k/cache': '/root/.r10k/cache')
```
haus, line 16: If home isn't set, why do we choose `/root` here and do we think it will be writable? If running as a service user or via a script that scrubs env vars, it seems like $HOME probably won't be set and the user won't be able to write to `/root`.

haus: But it seems like there are other settings like that, so maybe it's fine.

spec/unit/forge/module_release_spec.rb
```diff
@@ -9,33 +9,57 @@
...
-  describe '#download' do
-    it "downloads the module from the forge into `download_path`" do
-      expect(subject.forge_release).to receive(:download).with(download_path)
-      subject.download
+  context "no cached tarball" do
+    describe '#download' do
+      it "downloads the module from the forge into `download_path`" do
+        expect(tarball_cache_path).to receive(:exist?).and_return(false)
+        expect(subject.forge_release).to receive(:download).with(download_path)
+        allow(FileUtils).to receive(:mkdir_p).with(anything())
```
haus: Do we want to be more specific about which directories are going to be created here?

lib/r10k/git/cache.rb
```diff
@@ -16,7 +16,7 @@ class R10K::Git::Cache
 
   include R10K::Settings::Mixin
 
-  def_setting_attr :cache_root, File.expand_path(ENV['HOME'] ? '~/.r10k/git': '/root/.r10k/git')
+  def_setting_attr :cache_root, File.expand_path(ENV['HOME'] ? '~/.r10k/cache': '/root/.r10k/cache')
```
haus: Why change the default location for the git cache?

haus (commented):
Looks pretty good to me. I had a few comments/questions, mainly from unfamiliarity with r10k.

lib/r10k/forge/module_release.rb
andersonmills: Here, I was following the original git cache settings. Do you want to move this to /tmp if no $HOME?

lib/r10k/git/cache.rb
andersonmills: I changed this because originally, there was only the git cache. Now, we're adding the forge module cache to it, and in my opinion, they should be using the same root.

lib/r10k/forge/module_release.rb
haus: i don't have context on why the other cache was in root to begin with, so it's probably fine, just struck me as odd.

lib/r10k/git/cache.rb
haus: does the git cache just go immediately under the directory? i'm curious about how tarballs and git caches will be laid out together here.

do you have tree output or something from a cache dir with a mix of the two?

spec/unit/forge/module_release_spec.rb
andersonmills: Will use instance_variable_set to set that, otherwise it's controlled by the presence/absence of r10k.yaml.

lib/r10k/git/cache.rb
andersonmills: Ah. forge tarball caching happens in a directory named for the module, ex branan-eight_hundred-8.0.0/, and git module and control caching happens in a directory named for the repo, ex. git@github.com-andersonmills-r10k-testing.git/.

lib/r10k/forge/module_release.rb
andersonmills: Agreed, that does seem weird. 

It is documented, though. :)

> The cachedir setting defaults to ~/.r10k. If the HOME environment variable is unset r10k will assume that r10k is being run with the Puppet prerun_command setting and will set the cachedir default to /root/.r10k.

lib/r10k/git/cache.rb
haus: okay, cool. that makes sense i just wanted to make sure they wouldn't cross paths.

lib/r10k/forge/module_release.rb
```diff
@@ -41,9 +46,14 @@ def initialize(full_name, version)
         # objects are created in the class instances and thus are not shared with
         # subclasses.
         PuppetForge::V3::Release.conn = PuppetForge::V3::Base.conn
+
         @forge_release = PuppetForge::V3::Release.new({ :name => @full_name, :version => @version, :slug => "#{@full_name}-#{@version}" })
 
-        @download_path = Pathname.new(Dir.mktmpdir) + (@forge_release.slug + '.tar.gz')
+        tarball_name = @forge_release.slug + '.tar.gz'
+        @download_path = Pathname.new(Dir.mktmpdir) + (tarball_name)
+        @tarball_cache_root = Pathname.new(settings[:cache_root]) + (@forge_release.slug + "/tarball/")
```
ajroetker, line 62: Does this variable need an `attr_accessor` above too? (as with https://github.com/puppetlabs/r10k/pull/729/files#diff-41ce558d2e9e83cf146b0fad6c9811d6R33)

lib/r10k/forge/module_release.rb
```diff
@@ -65,34 +75,47 @@ def install(target_dir)
...
 
-      # Download the module release to {#download_path}
+      # Download the module release to {#download_path} and cache to {#tarball_cache_root}
       #
       # @return [void]
       def download
-        logger.debug1 "Downloading #{@forge_release.slug} from #{PuppetForge::Release.conn.url_prefix} to #{@download_path}"
-        @forge_release.download(download_path)
+        if @tarball_cache_path.exist?
+          logger.debug1 "Cached copy of #{@forge_release.slug} tarball found"
```
ajroetker: Nit pick, could you change this to be an action like `Using cached copy of tarball for ..`

lib/r10k/forge/module_release.rb
```diff
@@ -65,34 +75,47 @@ def install(target_dir)
...
+        else
+          logger.debug1 "Downloading #{@forge_release.slug} from #{PuppetForge::Release.conn.url_prefix} to #{@download_path}"
```
ajroetker: Why don't we just down directly to the cache directory?

ajroetker: Actually, why do we need download_path at all anymore? Can they turn off cacheing?

lib/r10k/forge/module_release.rb
```diff
@@ -105,7 +128,6 @@ def unpack(target_dir)
       # Remove all files created while downloading and unpacking the module.
       def cleanup
         cleanup_unpack_path
-        cleanup_download_path
```
ajroetker: Why aren't we calling this function anymore?

lib/r10k/git/cache.rb
ajroetker: Should we document that they need to cleanup the old cache? Or emit a warning for them that their old cache will need a delete?

ajroetker (changes requested):
Should we have an option that bypasses the cache?
Should we have a command to cleanup the cache for them? (i.e. remove outdated modules that aren't in their enviroments?)

lib/r10k/forge/module_release.rb
andersonmills: This still exists so there's a location that it's downloaded to, which takes time, and then it's mv'd to the cache directory, which is effectively atomic. It minimizes the chance of getting partial downloads stuck somewhere.

lib/r10k/forge/module_release.rb
andersonmills: Because putting this in the cache mv's the tarball now instead of unpacking from that location and leaving it there.

lib/r10k/forge/module_release.rb
andersonmills: Need? No. It would make testing in the spec test a bit more clean, though.

lib/r10k/forge/module_release.rb
andersonmills: Good point.

lib/r10k/forge/module_release.rb
andersonmills: This did make me do a search for that variable, and made me realize I missed changing some comments/documentation.

lib/r10k/forge/module_release.rb
andersonmills: They are currently not able to turn off caching.

lib/r10k/forge/module_release.rb
andersonmills: This could be added back for the case of things getting stopped halfway through.

lib/r10k/git/cache.rb
andersonmills: Yes and no. In r10k documentation, we don't talk about the cache in a way which suggests the user might do anything with it. It might be worth noting this in the CHANGELOG, for those interested in that level of detail? I don't have another location which seems appropriate for a mid-version cache clearing.

As for code-manager/PE, we always specify the cache location.

ajroetker: The cache won't get cleaned up for them then @andersonmills, how big is that cache?

ajroetker:
@andersonmills @haus what do you all think about https://github.com/puppetlabs/r10k/pull/729#pullrequestreview-38230422 ?

lib/r10k/git/cache.rb
andersonmills: Nowadays, that cache is as big as the data they want to import for the module. We have one customer who had 10G in a cache.

andersonmills: So, documenting where the cache is AND was seems like a good idea. Letting them know that they can blow the cache away, and the result will simply be that it is rebuilt also seems important.

andersonmills:
Sorry, didn't see the "conversation" comment, just looked at the ones in the code.

An option to bypass the cache. That has not been a part of r10k so far. Is that a good idea? Doesn't feel like it for the git portion of the cache.

Or are you speaking specifically about the forge module cache? In that we shouldn't implement forge module caching until we add an option for bypassing it?

Should we make an option/action to clean up the cache? Does this mean blow away the entire cache? Probably. I do this so often, it'd be great to have it just for me. Documenting the cache and how it can be removed seems like a good move in that direction.

andersonmills:
I just took a look and adding the forge module cache is only 5M for ~70 modules.

We could leave the default location for the git cache where it is, and make the forge module cache location next to it.

We could write up tickets for bypassing cache, cleaning the cache, and unifying the default cache locations.

lib/r10k/forge/module_release.rb
```diff
@@ -65,34 +79,47 @@ def install(target_dir)
...
       #
       # @raise [PuppetForge::V3::Release::ChecksumMismatch] The
-      #   downloaded module release checksum doesn't match the expected Forge
+      #   cached module release checksum doesn't match the expected Forge
       #   module release checksum.
       # @return [void]
       def verify
-        logger.debug1 "Verifying that #{download_path} matches checksum #{@forge_release.file_md5}"
-        @forge_release.verify(download_path)
+        logger.debug1 "Verifying that #{@tarball_cache_path} matches checksum #{@forge_release.file_md5}"
```
andersonmills: This should download the md5 and store it next to the tarball, and verify using that.

spec/unit/forge/module_release_spec.rb
```diff
@@ -7,35 +7,110 @@
...
+      expect(File).to receive(:write).with(md5_file_path, good_md5)
+      subject.verify_from_forge(md5_of_tarball)
+    end
+
+    it "raises an error and cleans up when the checksums do not match" do
+      expect(subject.forge_release).to receive(:file_md5).and_return(bad_md5)
+      expect(subject).to receive(:cleanup_cached_tarball_path)
+      expect { subject.verify_from_forge(md5_of_tarball) }.to raise_error(PuppetForge::V3::Release::ChecksumMismatch)
+    end
+    #it "verifies the module checksum based on the Forge file checksum" do
```
ajroetker: Should this be commented out still?

andersonmills: Ah. Oops. That should be completely removed. That test is superseded by the "write the md5 to file when the checksums match" test. Removing.

ajroetker (approved):

haus:
spacecat_03

haus (approved):

PR #730, opened 2017-05-16
(CODEMGMT-1064) Don't execute hostgenerator task body if BEAKER_HOST set

ajroetker:
This commit prevents the beaker_hostgenerator task from being run if the
BEAKER_HOST env variable is set which will override the output of the
task anyways.


haus:
:man_shrugging:

haus (approved):
alf-watching-toilet

PR #731, opened 2017-05-17
(CODEMGMT-1064) Move component pre-suite to "standard" location

ajroetker:
This commit moves the component pre-suite to the expected location in
the repo (under component/pre-suite instead of just component).


libbymo:
LGTM!

PR #732, opened 2017-05-18
(CODEMGMT-1064) Install dev r10k on master not client

ajroetker:
This client role doesn't exist in r10k testing, this commit changes the
installation of the dev r10k repos to use the master role instead.


libbymo:
LGTM!

PR #733, opened 2017-05-18
(CODEMGMT-1064) Inline the installation of dev repos function

ajroetker:
This commit changes the installation of the dev repos of r10k in testing
to be inline rather than def'd as a function which wasn't loading in
ruby during testing.


libbymo:
LGTM.

PR #734, opened 2017-05-18
(CODEMGMT-1064) Install pe-r10k on master not undefined host var

ajroetker:


haus (approved):
ron-swanson-is-excited

PR #735, opened 2017-05-18
(MAINT) Fixup Forge module deprecation check performance

scotje:
Before it was checking for module deprecation on every pass, but this
change makes it check only when it has decided it needs to
install/reinstall/upgrade anyway so there should be little to no
performance impact.

### DIFF
--- a/spec/unit/module/forge_spec.rb
+++ b/spec/unit/module/forge_spec.rb
@@ -63,15 +63,15 @@ describe R10K::Module::Forge do
   context "when a module is deprecated" do
     subject { described_class.new('puppetlabs/corosync', fixture_modulepath, :latest) }
 
-    it "warns on sync" do
-      allow(subject).to receive(:install)
+    it "warns on install" do
+      allow(R10K::Forge::ModuleRelease).to receive(:new).and_return(double('mod_release', install: true))
 
       logger_dbl = double(Log4r::Logger)
       allow_any_instance_of(described_class).to receive(:logger).and_return(logger_dbl)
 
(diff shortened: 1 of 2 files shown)


haus:
@scotje should we target this at 2.5.x instead?

scotje:
Should be easy to cherry pick either way but we can target 2.5.x initially if you like.

scotje:
Ooof, let me rebase.

haus:
oh nice

scotje:
OK, that looks better

spec/unit/module/forge_spec.rb
```diff
@@ -63,15 +63,15 @@
...
+    it "warns on install" do
+      allow(R10K::Forge::ModuleRelease).to receive(:new).and_return(double('mod_release', install: true))
 
       logger_dbl = double(Log4r::Logger)
       allow_any_instance_of(described_class).to receive(:logger).and_return(logger_dbl)
 
       expect(logger_dbl).to receive(:warn).with(/puppet forge module.*puppetlabs-corosync.*has been deprecated/i)
 
-      subject.sync
+      subject.install
```
haus: is it worth adding an expectation that there won't be a warn on a sync?

scotje: Probably...

scotje:
@haus how do you feel about the revised test cases?

haus:
i-love-you-so-much

haus (approved):

andersonmills (approved):

PR #736, opened 2017-05-22
(CODEMGMT-1064) Remove expected_error for RK-101

ajroetker:
This commit removes the expected_error related to duplicate module names
now that we properly error out on duplicate module names in r10k.

### DIFF
--- a/integration/tests/user_scenario/basic_workflow/negative/neg_duplicate_module_names.rb
+++ b/integration/tests/user_scenario/basic_workflow/negative/neg_duplicate_module_names.rb
@@ -40,12 +40,8 @@ git_add_commit_push(master, 'production', 'Add modules.', git_environments_path)
 #Tests
 step 'Attempt to Deploy via r10k'
 on(master, "#{r10k_fqp} deploy environment -v -p", :acceptable_exit_codes => [0, 1]) do |result|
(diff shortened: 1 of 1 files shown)


integration/tests/user_scenario/basic_workflow/negative/neg_duplicate_module_names.rb
```diff
@@ -40,12 +40,8 @@
...
 on(master, "#{r10k_fqp} deploy environment -v -p", :acceptable_exit_codes => [0, 1]) do |result|
-  expect_failure('Expected exit code to be wrong due to RK-101') do
-    assert_equal(1, result.exit_code, "Expected command to indicate error with exit code")
-  end
-
-  expect_failure('Expected module to be deployed twice due to RK-101') do
-    matches = result.stderr.scan(deploy_str)
-    assert_equal(1, matches.size, "Expected motd module to be deployed once, but deployed #{matches.size}) times")
-  end
+  assert_equal(1, result.exit_code, "Expected command to indicate error with exit code")
```
haus, line 43: So this change will need to be rolled back when we bring the work back to 2.5.x right?

ajroetker: I don't think this one will need to be since this change to r10k was only added in master @haus

PR #739, opened 2017-05-23
(CODEMGMT-1064) Move r10k 2.5.x CI to new pipelines

ajroetker:
This commit moves r10k 2.5.x CI to the new testing pipelines which align
r10k CI with the client-tools vanagon CI feeding r10k pipelines into
pe-r10k pipelines.


haus (approved):
dont-panic

PR #751, opened 2017-06-08
(RK-162) Remove PUPPETFILE and PUPPETFILE_DIR

andersonmills:
This commit fully removes effect of the PUPPETFILE and
PUPPETFILE_DIR environment variables.

### DIFF
--- a/lib/r10k/cli/puppetfile.rb
+++ b/lib/r10k/cli/puppetfile.rb
@@ -31,8 +31,8 @@ Puppetfile (http://bombasticmonkey.com/librarian-puppet/).
           usage   'install'
           summary 'Install all modules from a Puppetfile'
 
-          required nil, :moduledir, 'Path to install modules to (can also be set with PUPPETFILE_DIR environment variable)'
-          required nil, :puppetfile, 'Path to Puppetfile (can also be set with PUPPETFILE environment variable)'
+          required nil, :moduledir, 'Path to install modules to'
+          required nil, :puppetfile, 'Path to Puppetfile'
           # @todo add --no-purge option
           runner R10K::Action::Puppetfile::CriRunner.wrap(R10K::Action::Puppetfile::Install)
         end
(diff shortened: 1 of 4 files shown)


lib/r10k/cli/puppetfile.rb
```diff
@@ -46,7 +46,7 @@ def self.command
           usage 'check'
           summary 'Try and load the Puppetfile to verify the syntax is correct.'
 
-          required nil, :puppetfile, 'Path to Puppetfile (can also be set with PUPPETFILE environment variable)'
```
ajroetker, line 49: Should we actually remove these in the deprecation PR, that way we're not pointing users at them?

andersonmills: ~~Gah. I had some back and forth with original branches on the deprecation. In the original commit for RK-161, I put a big "DEPRECATED" inside the parens on the help page. Which do you think is a better choice?~~

D'oh. You fooled me just like the 2.5.x branch did. Those helpful parenthetical phrases already don't exist in the 2.5.x branch.

andersonmills: D'oh. You fooled me just like the 2.5.x branch did. Those helpful parenthetical phrases already don't exist in the 2.5.x branch.

andersonmills:
```
$ be r10k help puppetfile install
NAME
    install - Install all modules from a Puppetfile

USAGE
    r10k puppetfile install

OPTIONS
       --moduledir     Path to install modules to
       --puppetfile    Path to puppetfile

OPTIONS FOR PUPPETFILE
    -c --config        Specify a global configuration file
       --color         Enable colored log messages
    -h --help          Show help for this command
    -t --trace         Display stack traces on application crash
    -v --verbose       Set log verbosity. Valid values: fatal, error, warn, notice, info, debug, debug1, debug2

$ be r10k version
r10k 2.5.5
```

PR #753, opened 2017-07-10
(RK-295) Remove locales/r10k.pot from gitignore

andersonmills:


haus:
Looks good to me. Do we want to update https://github.com/puppetlabs/r10k/blob/master/r10k.gemspec#L34 at all?

andersonmills:
">= 0.10", "< 1" ??

I see that 0.25 is out, and that's what I'm using on my laptop. Can go all the way to there, but I saw puppet itself specifies the above.

haus:
That seems fine. Folks in the i18n room seemed to think it should be fine to go all the way to 0.25

`~> 0.25` should be the same as that btw.

haus (approved):
ronswansondancing

Issue #752, opened 2017-07-03
puppet_forge 2.2.7 installs semantic_puppet 1.0.0, which breaks r10k


stblassitude:
/usr/local/lib/ruby/site_ruby/2.3/rubygems/specification.rb:1436:in `block in activate_dependencies': can't satisfy 'semantic_puppet (~> 0.1.0)', already activated 'semantic_puppet-1.0.0' (Gem::LoadError)
	from /usr/local/lib/ruby/site_ruby/2.3/rubygems/specification.rb:1428:in `each'
	from /usr/local/lib/ruby/site_ruby/2.3/rubygems/specification.rb:1428:in `activate_dependencies'
	from /usr/local/lib/ruby/site_ruby/2.3/rubygems/specification.rb:1410:in `activate'
	from /usr/local/lib/ruby/site_ruby/2.3/rubygems.rb:300:in `block in activate_bin_path'
	from /usr/local/lib/ruby/site_ruby/2.3/rubygems.rb:300:in `synchronize'
	from /usr/local/lib/ruby/site_ruby/2.3/rubygems.rb:300:in `activate_bin_path'
	from /usr/local/bin/r10k:23:in `<main>'

I believe r10k should either bump semantic_puppet as well, or require puppet_forge 2.2.6.

scotje:
Looks like there is only one place that `r10k` invokes code from `semantic_puppet` so it might be better to add a helper to `puppet_forge` to handle that case so that `r10k` doesn't need to depend directly on `semantic_puppet` at all.

scotje:
https://github.com/puppetlabs/r10k/blob/master/lib/r10k/module/forge.rb#L20 is the only usage I see.

stblassitude:
I believe bumping the dependency is necessary and useful for cases where r10k is installed through an OS package. I'm running a Puppetserver on FreeBSD, where all components are installed from OS packages. Since the FreeBSD packages mirror the dependencies of the gems, the package manager would normally refuse to install both rubygem-puppet_forge 2.2.7 (requiring semantic_puppet 1.0.0) and rubygem-r10k (semantic_puppet 0.1.0). The maintainers have chosen to create a separate package rubygem-semantic_puppet0, which works around the issue for the package manager, but only shifts the problem to runtime, because r10k loads both puppet_forge and semantic_puppet, creating a conflict with two versions of semantic_puppet being in the path.

[...]

PR #755, opened 2017-07-11
Fix issue #752 by bumping the dependency

stblassitude:
Fix issue #752 by bumping the dependency.

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'multi_json', '~> 1.10'
 
   s.add_dependency 'puppet_forge', '~> 2.2'
-  s.add_dependency 'semantic_puppet', '~> 0.1.0'
+  s.add_dependency 'semantic_puppet', '~> 1.0.0'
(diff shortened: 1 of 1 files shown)


haus:
@stblassitude Thanks for this (and sorry for the delay in reviewing it).

r10k.gemspec
```diff
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'multi_json', '~> 1.10'
 
   s.add_dependency 'puppet_forge', '~> 2.2'
-  s.add_dependency 'semantic_puppet', '~> 0.1.0'
+  s.add_dependency 'semantic_puppet', '~> 1.0.0'
```
haus, line 32: I think this change will stop folks on puppet-agent 1.x from installing r10k, which would be too bad.
What if you changed the dependency to be `s.add_dependency 'semantic_puppet', '>= 0.1.0', '< 2.0.0'`?

andersonmills:
This PR is superceded by https://github.com/puppetlabs/r10k/pull/775

PR #763, opened 2017-09-10
WIP: Fixes #710 - Create serialized format API

### LINKED ISSUE #710: Create serialized format API for importing / exporting the Puppetfile
In reference to #564.

Many tools, and external processes often need to interface with the Puppetfile in a programmatic form. The puppetfile allows us to interact by evaling the Puppetfile but we also want the ability to interact with r10k via different Puppetfile formats.  Since these formats and serialization methods can differ we need to create an API that will allows users to create their own format or use any 3rd party format easily. 

The API should should meet the following expectations.

1. allow the user to choose which serialization to use (json, yaml, hocon, toml, ...)
2. allow the user to choose which data format to use (the one in #564, direct (@modules.to_json), some_other_format)
3. The API should follow the adapter pattern and provide a contract with only a few methods the user needs to override to create a new format.  This should be simple to create with only a type, serialize, and deserialize method. 
4. The r10k interface should add a new action called export which would allow the user to export the puppetfile to a different format.
5. The r10k interface should add a new action called import which would convert a format to a Puppetfile
6. The r10k interface should also automatically try to import a format given a serialized formatted file.  
7.  A serialized format should include the format type in the data it contains so that r10k can decipher which format it should use to deserialize the data. 
8. R10k should respect environments when exporting or importing files.
9. R10k should automatically detect the type of serialization to use given the extension name of the file. 
10. R10k should import the user's specified format via  settings
11. R10k should import the Puppetfile format via some predefined order if none is set in settings

Possible workflow:

1. r10k --export battletech.json --environment=dev
2. r10k deploy dev   (assumes battletech.json is in r10k-control)
3. r10k --import battletech.json deploy dev  (overrides any setting of format to search for)

Possible API.  Never tested.  Take with a grain of salt.

```
module R10k
  module SerializedFormat
    class BattleTech
         include SerializedFormatAPI

        def serialize(modues)
          modules.to_json
        end

        def deserialize(formatted_object)
           # reconstruct puppetfile ruby object
           to_puppetfile_obj(formatted_object)
        end

        def to_puppetfile_obj(data)
           
        end

        def type
           :battletech
        end

    end
  end
end



```

logicminds:
* allows for various implementations of a Puppetfile format

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ end
 group :development do
   gem 'simplecov', '~> 0.9.1'
   gem 'ruby-prof', :platforms => :ruby
+  gem 'pry'
 end
 
 if File.exists? "#{__FILE__}.local"
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -105,7 +105,7 @@ module R10K
         end
 
         def visit_puppetfile(puppetfile)
-          puppetfile.load
+          puppetfile.desired_contents
 
           yield
 
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -43,7 +43,7 @@ module R10K
         end
 
         def visit_puppetfile(puppetfile)
-          puppetfile.load
+          puppetfile.desired_contents
           yield
         end
 
--- a/lib/r10k/action/puppetfile/check.rb
+++ b/lib/r10k/action/puppetfile/check.rb
@@ -10,7 +10,7 @@ module R10K
         def call
           pf = R10K::Puppetfile.new(@root, @moduledir, @puppetfile)
           begin
-            pf.load!
+            pf.desired_contents  # loads the contents
             $stderr.puts _("Syntax OK")
             true
           rescue => e
--- a/lib/r10k/action/puppetfile/install.rb
+++ b/lib/r10k/action/puppetfile/install.rb
@@ -20,7 +20,7 @@ module R10K
         include R10K::Action::Visitor
 
         def visit_puppetfile(pf)
-          pf.load!
+          pf.desired_contents
           yield
           pf.purge!
         end
--- a/lib/r10k/action/puppetfile/purge.rb
+++ b/lib/r10k/action/puppetfile/purge.rb
@@ -9,7 +9,7 @@ module R10K
 
         def call
           pf = R10K::Puppetfile.new(@root, @moduledir, @puppetfile)
-          pf.load!
+          pf.desired_contents  # loads the contents
           pf.purge!
           true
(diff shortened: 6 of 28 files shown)


logicminds:
I spent a few hours on a prototype to implement what I was alluding to with #710.  This will need some more work but should illustrate how one would implement a new formatter.   Additionally, we can also load from external gems should someone not want to put their formatter in r10k core.

The API workflow follows this pattern:

1. Set puppetfile path, no matter what kind of format that may be  
2. Find all library paths and gems that contain a r10k formatter
3. loop through each formatter and find the first formatter that has successfully validated the given puppetfile.  
4. Return this formatter to the puppetfile object
5. Load the content of the puppetfile file and populate the puppetfile object with modules and such.


Things I don't like:

1. The puppetfile object is a bad name, it should be renamed to librarian. 
2. The puppetfile object currently requires the formatter class to populate the modules object.  This seems backwards.  Instead each formatter should return a list of modules and the puppetfile object would set directly based on the formatter.  This can be fixed, but in the meantime I just passed the librarian object to the formatter.

logicminds:
I can finish this up if this is something that would have a chance of being merged.

logicminds:
@gregswift  any feedback on this?  It has been over two weeks and nothing but crickets.  I also created another branch that implements a new provider based on this adapter pattern which add support for #564.

At the moment the auto detection mechanism would need to be enhanced unless we start adding a type attribute to the puppetfile.

logicminds:
https://github.com/nwops/r10k/blob/yaml_serializer/lib/r10k/formatter/yaml_formatter.rb 

This is the yaml adapter based off of #564.  It is not quite finished but represents how one would go about creating a new adapter.

rackergs:
i had actually come and given a thumbs up on the original comment you added to this ticket, basically giving my agreement. I didn't have any specific comments to add at this or that time.  I'm more interested in improved support happening than the specific implementation.

It seems like a reasonable path.

logicminds:
Great, I will polish this a little more and possibly discuss at the contributor summit if you are attending. There is probably another good 8 hours involved and some testing.  Travis seems to be broken on this branch for select rubies, not sure why.

rackergs:
I wont be at the summit.. unfortunately but that would be an awesome outcome from the conf! woot

logicminds:
I made some more progress and only have a few tests that are failing.  I also implemented the yaml format for demo purposes.  Just need to finalize a few things and add some docs.  Have a look at the yaml formatter to get an idea of how one could implement a new formatter.  

https://github.com/puppetlabs/r10k/pull/763/files#diff-f2cadeb2175d51e6f90420d2ebd25b0e

logicminds:
closing this since I just don't have any motivation to finish this.

PR #772, opened 2017-10-27
(RK-264) Add --update_force flag to puppetfile install

andersonmills:
This PR adds a --update_force flag to the puppetfile install action and subcommand. This will force the overwrite of modules which have local changes.


cbrock:
👍

PR #774, opened 2017-11-08
(RK-305) Add --no-force to deploy action.

andersonmills:
This commit adds a --no-force flag to the deploy action. The default of the r10k deploy action is to deploy all environments regardless of local modifications in the modules. The --no-force flag, the user can choose to have the deploy skip environments where local modifications have occurred to the modules.

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -13,7 +13,7 @@ module R10K
 
         include R10K::Action::Deploy::DeployHelpers
 
-        attr_reader :no_force
+        attr_reader :force
 
         def initialize(opts, argv, settings = nil)
           settings ||= {}
@@ -22,6 +22,8 @@ module R10K
 
           super
 
+          # @force here is used to make it easier to reason about
+          @force = !@no_force
           @argv = @argv.map { |arg| arg.gsub(/\W/,'_') }
         end
 
@@ -119,8 +121,7 @@ module R10K
 
         def visit_module(mod)
           logger.info _("Deploying Puppetfile content %{mod_path}") % {mod_path: mod.path}
-          # no_force here is the opposite of the expected value of force.
-          mod.sync(force: !@no_force)
+          mod.sync(force: @force)
         end
 
         def write_environment_info!(environment, started_at, success)
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -146,7 +146,7 @@ module R10K
         end
 
         def allowed_initialize_opts
-          super.merge(puppetfile: :self, cachedir: :self, no_force: :self)
+          super.merge(puppetfile: :self, cachedir: :self, 'no-force': :self)
         end
       end
     end
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -146,7 +146,7 @@ module R10K
         end
 
         def allowed_initialize_opts
-          super.merge(puppetfile: :self, cachedir: :self, 'no-force': :self)
+          super.merge(puppetfile: :self, cachedir: :self, :'no-force' => :self)
         end
       end
     end
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -13,6 +13,8 @@ module R10K
 
         include R10K::Action::Deploy::DeployHelpers
 
+        attr_reader :no_force
+
         def initialize(opts, argv, settings = nil)
           settings ||= {}
           @purge_levels = settings.fetch(:deploy, {}).fetch(:purge_levels, [])
@@ -117,7 +119,8 @@ module R10K
 
         def visit_module(mod)
           logger.info _("Deploying Puppetfile content %{mod_path}") % {mod_path: mod.path}
-          mod.sync
+          # no_force here is the opposite of the expected value of force.
+          mod.sync(force: !@no_force)
         end
 
         def write_environment_info!(environment, started_at, success)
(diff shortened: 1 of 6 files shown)


lib/r10k/action/deploy/environment.rb
```diff
@@ -117,7 +119,8 @@ def visit_puppetfile(puppetfile)
 
         def visit_module(mod)
           logger.info _("Deploying Puppetfile content %{mod_path}") % {mod_path: mod.path}
-          mod.sync
+          # preseve_force here is the opposite of the expected value of force.
```
haus: typo in comment

haus:
I tried to figure out what the flag was doing based on its name and the commit message and I still have no idea. What are we preserving and what are we forcing?

reidmv:
Reading the original ticket, the UX suggestions were `--force` and the opposing `--no-force`. Those start to imply something and are relatively intuitive.

I'm completely confused by `--preserve_force`. Regardless of what it does, I don't think it's good UX.

andersonmills:
I agree, --preserve_force isn't what we want.

In this case, we would --no-force (or --no_force) be clear in that it is not overwriting local modifications when overwriting local modification would be the default. Note that for ```r10k deploy environment```, there is no --force option. That would only appear in ```r10k puppetfile install```.

Another thought I had was --overwrite_local, which defaulted to true or you could also specify ```--overwrite_local true``` and ```--overwrite_local false```. Thoughts @reidmv @haus ? Should I put this in front of the command_line UX group?

gdubicki:
2 cents from r10k user here: `--overwrite_local` to overwrite local modifications, and not adding this to preserve them seems to be simplest and most intuitive to me.

andersonmills:
@gdubicki That would, unfortunately, mean that we would reverse the default once again for ```r10k deploy environment```.

reidmv:
Puppet's convention for Boolean arguments is to support `--[no-]arg`. That is, for any boolean flag we add for "arg", both `--arg` (arg=true) and `--no-arg` (arg=false) are supported.

Noting this since regardless of what we call the flag, there should be both a true/false version of them and thus we don't need to worry much about what the default is in the context of what the flag is.

andersonmills:
@reidmv @haus I've changed this flag to no_overwrite_force. I'll write up another ticket to change RK-264's "--update-force" to "--overwrite_local" (EDIT: NOPE! Changing it to "--force").

scotje:
`<bikeshedding>`

I don't know if it's defined explicitly anywhere (and it looks like r10k currently doesn't have any existing multi-word options or flags) but it seems like using hyphens as a word separator in command line options and flags is more prevalent than underscores.

`</bikeshedding>`

haus:
@scotje puppet-nogui seems to agree with you based on https://github.com/puppetlabs/puppet-nogui/blob/4c2467e88387daf79236f6eb9760e32e1b9caf57/patterns/command_taxonomy.md#default-options

> An external subcommand SHOULD format options that contain multiple words using hyphens, as in --help-summary. Short words may be grouped together for commonly used options. It MUST NOT use underscores to separate words.

reidmv:
@scotje not bike shedding, totally legitimate UX observation.

scotje:
@haus ha, I even poked around in that repo and didn't find that. :)

reidmv:
FYI: I argued the case to @andersonmills offline that for this use case, we really just want `--force` / `--no-force`.

Consider a single environment, _bar_.

Let a user perform a deployment against _bar_, and let the result of that deployment be no action because r10k detected local file modifications and therefore declined to act.

I posit that the result of this operation is "Error: failed to deploy". That is, the deployment did not succeed.

Specifically, the deployment did not succeed because r10k detected that a destructive action would be required to complete it, and so aborted the deployment operation.

From this perspective, r10k needs some directive to tell it to proceed with a deployment even if it needs to perform a destructive action in order to complete it. The right option to pass here is just `--force`.

The opposing boolean option then is `--no-force`.

We don't need to get more complicated about it or try to come up with three-words-or-less ways to describe the implementation. We just need to provide a way for the user to indicate one of:
* JUST DO IT, even if it's potentially destructive
* No, no, no; don't "just do it" if it would be in any way destructive; stand down

So, `--force`. Same basic rational as what `-f` means to `rm`.

andersonmills:
After running around town with this, we're back to --no-force. Whee!

andersonmills:
@haus @scotje Yep. I agree with y'all's comments. The usage we were making of the cri library meant that we couldn't have dashes in our options flags, because they were getting turned into instance variables. I traced the issue down and made a change in the options settings for this. It wasn't found in the first cut.

I ALSO poked around in the nogui repo and didn't find it.

andersonmills:
I've created #780 to parallel the final decision about the flag names.

haus:
This looks pretty reasonable to me. I have two bits of feedback. First, I think even if the flag for users is `--no-force`, it should probably set a `@force` instance variable internally that is referenced. I think this will make it a little easier to reason about.

Second, I think having a test with a dirty module that asserts that the deploy doesn't happen if `--no-force` is set would be valuable?

andersonmills:
I updated according to your first suggestion. Because the option and an instance variable are created from the option name, it feels a bit like a mishmash, BUT it does look better in the Module#sync calls.

haus:
Awesome.

andersonmills:
@haus So, the test you're suggesting sounds like an integration test. Mind if I put that into a different ticket?

haus:
@andersonmills sure that seems fine. I'm not familiar enough with r10k's tests, so i sort of figured it would be in rspec. not sure it's worth it for a beaker test.

andersonmills:
Does that mean this gets an approval from you, @haus ?

haus:
Sure. Hit it.

PR #775, opened 2017-11-09
(RK-306) Remove dependency on semantic_puppet.

andersonmills:
This commit uses a helper method in puppet_forge for version validation.
This change removes the direct dependency of r10k on semantic_puppet.

### DIFF
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -6,7 +6,7 @@ require 'r10k/forge/module_release'
 
 require 'pathname'
 require 'fileutils'
-require 'puppet_forge'
+require 'puppet_forge/util'
 
 class R10K::Module::Forge < R10K::Module::Base
 
@@ -17,7 +17,7 @@ class R10K::Module::Forge < R10K::Module::Base
   end
 
   def self.valid_version?(expected_version)
-    expected_version == :latest || expected_version.nil? || SemanticPuppet::Version.valid?(expected_version)
+    expected_version == :latest || expected_version.nil? || PuppetForge::Util.version_valid?(expected_version)
   end
 
   # @!attribute [r] metadata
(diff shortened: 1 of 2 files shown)


haus:
It seems like if one of our libraries isn't doing what we need we should be fixing semantic_puppet itself and not working around it?

andersonmills:
However, fixing this up in this way allows puppet_forge to implement a version_valid? check in whatever way it wants. Then we get the added benefit of not requiring semantic_puppet with r10k.

haus:
Sure, but both Puppet (which uses semantic_puppet) and r10k and the forge use this to decide what is valid. Seems like they might want to agree on what is valid (unless puppet isn't using this on modules (or puppet disagrees on what semver means), in which case I don't know why we would ever use the same shared libraries).

haus:
@andersonmills also I'm looking through the ticket and not seeing any actual issues with semantic_puppet, just version conflicts. Does relaxing the version constraints not address the problem and if not, how does it not?

scotje:
> Sure, but both Puppet (which uses semantic_puppet) and r10k and the forge use this to decide what is valid. Seems like they might want to agree on what is valid (unless puppet isn't using this on modules (or puppet disagrees on what semver means), in which case I don't know why we would ever use the same shared libraries).

r10k itself doesn't care about version numbers except in the context of interacting with the Forge. So my feeling is that moving that responsibility into the puppet_forge gem makes logical sense and also reduces the overall complexity of the dependency graph.

haus:
I think that's fine. I don't have the history as to why we were using semantic_puppet to begin with. Mostly I worry that if semantic_puppet has an issue with valid module versions then puppet will experience a similar problem someday too.

andersonmills:
@haus Ah, I get more where you're coming from now. Yeah, the only place that r10k uses semantic_puppet is for forge modules, so it seems right to me that the version validation is moved to the puppet_forge library. Any semantic_puppet error will be surfaced through the puppet_forge library.

haus:
Do we want this change here or in 2.6.x?

andersonmills:
I don't think it'll matter. Everything we add to 2.5.x will need to get merged up.

I also don't anticipate tagging another 2.5.x version.

haus (approved):
:woman_shrugging:

PR #757, opened 2017-07-16
Filter branches that match regex

nmaludy:
Implementation of feature request #574 

This PR implements branch filtering based on a regex provided in the r10k.yaml configuration file for a given source. There is a new setting option for a source called `branch_filter:`. If specified, then only branches that match the regex will be deployed into environments. If `branch_filter:` is not specified, then all branches will be deployed. This makes it a backwards compatible change.

Please see the updated documentation for configuration in this PR for example configurations and regex patterns.

Comments and reviews appreciated.

Issue #574, opened 2016-02-18
[feature]: ignore branches that matches regexp


iakovgan:
Some times we need a temporary feature branch in environment repository, and we do not need to deploy r10k environment. It would be  nice to have a :ignore_branch_regexp: . If branch matches regexp, it would not be deployed by r10k.

hybby:
This would be great.  Or the inverse, a setting where only branch names that match the regexp will be deployed.

fatmcgav:
Another +1 from me aswell... I've got some 'local' dev env setups that I dont want to deploy on a production PM for example...

martinseener:
What's the status of this? I don't think it's too hard but iam Ops not a dev, so it's hard for me to do the PR myself but i like this feature, since we constantly have 2 to 5 other branches we develop on, so the more branches r10k uses, the longer the run takes :(

nmaludy:
+1 here

We use "feature" and "bugfix" branches all the time and it would be nice to have r10k ignore then instead of cluttering our environment directory.

gregoirefra:
+1 We create a bunch of branch for sometime no reason because some of my team member are not used to git yet and some of those branch if not all are not really required to be deploy so being able to whitelist the one we want would be a really nice to have 😃

nmaludy:
I've opened up a PR that fully implements this feature. If anyone is interested in helping test, the code can be found here: https://github.com/EncoreTechnologies/r10k/tree/feature/branch-filtering

pietervogelaar:
+1

I really want this. We use feature branches to make changes to the Puppetfile in environment branches. But r10k must not try to deploy these feature branches.

nmaludy:
@pietervogelaar good news, i have a PR available: https://github.com/puppetlabs/r10k/pull/757

Just needs to be reviewed and merged.

adrienthebo:
PR #777 looks like it implemented this feature. I'm going to double check that this feature was implemented and if so I'll close out this issue.

PR #777, opened 2017-11-17
(RK-307) Ignore branches for deployment with specified prefixes

andersonmills:
This is a partial implementation of feature request #574

This PR implements ignoring branches based on a prefix list provided in the r10k.yaml configuration file for a given source. There is a new setting option for a source called ignore_branch_prefixes:. If specified, then only branches that do not start with a string in the list will be deployed into environments. If ignore_branch_prefixes: is not specified, then all branches will be deployed. This makes it a backwards compatible change.

Please see the updated documentation for configuration in this PR for examples.

Comments and reviews appreciated.

The original work for this came from @nmaludy and is adapted from #757

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -501,7 +501,7 @@ sources:
   app1_data:
     remote: 'git://git-server.site/my-org/app1-hieradata'
     basedir: '/etc/puppet/hieradata'
-    branch_filter: '^(prod-)\\w+$'
+    branch_filter: '^(prod-)\w+$'
 ```
 
 #### Filter branches - Match branches exactly
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -356,27 +356,30 @@ sources:
 * if `false` (default) environment folder will not be prefixed
 * if `String` environment folder will be prefixed with the `prefix` value.
 
-### branch_filter
+### ignore_branch_prefixes
 
-The 'branch_filter' setting restricts the environment that are deployed to only
-those that match the given filter. The filter is a regex string. Each branch in
-the 'git' repo will have its named tested against this regex and, if the regex
-matches the branch name, then an environment will be deployed for this branch.
-If no 'branch_filter' is specified, then all branches in the 'git' repo will
+The 'ignore_branch_prefixes' setting causes environments to be ignored which match in part or whole
+to any of the prefixes listed in the setting.
+The setting is a list of strings. Each branch in
+the 'git' repo will have its name tested against all prefixes and, if the prefix
+is found, then an environment will not be deployed for this branch.
+If no 'ignore_branch_prefixes' is specified, then all branches in the 'git' repo will
 be deployed (default behavior).
 
-#### branch_filter behaviour
+#### ignore_branch_prefixes behaviour
 * if empty, deploy environments for all branches
 * for each branch in git repo
-** if `branch.name` matches `branch_filter` regex, then deploy an environment for branch
+** if `branch.name` has a prefix found in `ignore_branch_prefixes`, then do not deploy an environment for branch
 
-Example: only deploy branches named 'prod' or 'qa'
+Example: do not deploy branches with names starting with (or completely named) 'test' or 'dev'.
 ```yaml
 ---
 sources:
   mysource:
     basedir: '/etc/puppet/environments'
-    branch_filter: '^(prod)|(qa)$'
+    ignore_branch_prefixes:
+      - 'test'
+      - 'dev'
 ```
 
 Examples
@@ -477,46 +480,3 @@ This will create the following directory structure:
 |-- app1_production  # app1 data repository, production branch
 |-- app1_develop     # app1 data repository, develop branch
 ```
-
-### Filtering branches
-
-#### Filtering branches - Exclude branches with a prefix
-
-If you would like to filter out all branches starting with and including "dev-".
-
-```yaml
----
-sources:
-  app1_data:
-    remote: 'git://git-server.site/my-org/app1-hieradata'
-    basedir: '/etc/puppet/hieradata'
-    branch_filter: '^(?!dev-).*$'
-```
-
-#### Filter branches - Include branches with a prefix
-
-If you would like to only deploy branches starting with a prefix "prod-" followed by word characters. Note: this does NOT match the "prod-" branch exactly due to the \w+.
-
-```yaml
----
-sources:
-  app1_data:
-    remote: 'git://git-server.site/my-org/app1-hieradata'
-    basedir: '/etc/puppet/hieradata'
-    branch_filter: '^(prod-)\w+$'
-```
-
-#### Filter branches - Match branches exactly
-
-If you would like to only deploy branches whose name matches exactly to 'master', or 'production'.
-
-```yaml
----
-sources:
-  app1_data:
-    remote: 'git://git-server.site/my-org/app1-hieradata'
-    basedir: '/etc/puppet/hieradata'
-    branch_filter: '^(master)|(production)$'
-```
-
-
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -478,6 +478,8 @@ This will create the following directory structure:
 |-- app1_develop     # app1 data repository, develop branch
 ```
 
+### Filtering branches
+
 #### Filtering branches - Exclude branches with a prefix
 
 If you would like to filter out all branches starting with and including "dev-".
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -356,6 +356,29 @@ sources:
 * if `false` (default) environment folder will not be prefixed
 * if `String` environment folder will be prefixed with the `prefix` value.
 
+### branch_filter
+
+The 'branch_filter' setting restricts the environment that are deployed to only
+those that match the given filter. The filter is a regex string. Each branch in
+the 'git' repo will have its named tested against this regex and, if the regex
+matches the branch name, then an environment will be deployed for this branch.
+If no 'branch_filter' is specified, then all branches in the 'git' repo will
+be deployed (default behavior).
+
+#### branch_filter behaviour
+* if empty, deploy environments for all branches
+* for each branch in git repo
+** if `branch.name` matches `branch_filter` regex, then deploy an environment for branch
+
+Example: only deploy branches named 'prod' or 'qa'
+```yaml
+---
+sources:
+  mysource:
+    basedir: '/etc/puppet/environments'
+    branch_filter: '^(prod)|(qa)$'
+```
+
 Examples
 --------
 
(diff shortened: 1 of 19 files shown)


nmaludy (approved):

doc/dynamic-environments/configuration.mkd
```diff
@@ -356,6 +356,29 @@ sources:
...
+* for each branch in git repo
+** if `branch.name` matches `branch_filter` regex, then deploy an environment for branch
+
+Example: only deploy branches named 'prod' or 'qa'
+```yaml
+---
+sources:
+  mysource:
+    basedir: '/etc/puppet/environments'
+    branch_filter: '^(prod)|(qa)$'
```
haus: You shouldn't need 2 groups here. `'^(prod|qa)$'` should also work?

doc/dynamic-environments/configuration.mkd
```diff
@@ -454,3 +477,46 @@ This will create the following directory structure:
...
+
+If you would like to only deploy branches starting with a prefix "prod-" followed by word characters. Note: this does NOT match the "prod-" branch exactly due to the \w+.
+
+```yaml
+---
+sources:
+  app1_data:
+    remote: 'git://git-server.site/my-org/app1-hieradata'
+    basedir: '/etc/puppet/hieradata'
+    branch_filter: '^(prod-)\w+$'
```
haus: You shouldn't need a grouping here at all?

doc/dynamic-environments/configuration.mkd
```diff
@@ -454,3 +477,46 @@ This will create the following directory structure:
...
+
+If you would like to only deploy branches whose name matches exactly to 'master', or 'production'.
+
+```yaml
+---
+sources:
+  app1_data:
+    remote: 'git://git-server.site/my-org/app1-hieradata'
+    basedir: '/etc/puppet/hieradata'
+    branch_filter: '^(master)|(production)$'
```
haus: This can be `'^(master|production)$'`.

doc/dynamic-environments/configuration.mkd
```diff
@@ -454,3 +477,46 @@ This will create the following directory structure:
 |-- app1_production  # app1 data repository, production branch
 |-- app1_develop     # app1 data repository, develop branch
 ```
+
+### Filtering branches
+
+#### Filtering branches - Exclude branches with a prefix
+
+If you would like to filter out all branches starting with and including "dev-".
```
haus: Excluding branches this way feels a bit awkward. I wonder if having two filters, one for include and one for exclude would be better? That way you could write positive regexes in both cases, which is usually easier. I think in that case you would add `'^dev-.*'` to an exclude regex and any matches would be excluded. Not sure if that is better or not.

lib/r10k/source/git.rb
```diff
@@ -113,7 +119,17 @@ def desired_contents
 
   def branch_names
     opts = {:prefix => @prefix, :invalid => @invalid_branches, :source => @name}
-    @cache.branches.map do |branch|
+    branches = @cache.branches
+    if !@branch_filter.nil? && !@branch_filter.empty?
+      branches = branches.select do |branch|
+        result = branch[/#{@branch_filter}/]
```
haus: This regex match syntax is fairly weird for ruby. I would expect something more like:

```
filter = Regexp.new(@branch_filter)
...
result = filter.match(branch)
...
```

lib/r10k/source/git.rb
```diff
@@ -113,7 +119,17 @@ def desired_contents
 
   def branch_names
     opts = {:prefix => @prefix, :invalid => @invalid_branches, :source => @name}
-    @cache.branches.map do |branch|
+    branches = @cache.branches
+    if !@branch_filter.nil? && !@branch_filter.empty?
```
haus: The double negation here is funky. I would recommend a partially positive test instead such as...

`if @branch_filter && !@branch_filter.empty?`
or
`if @branch_filter && @branch_filter.size > 0`

lib/r10k/source/git.rb
```diff
@@ -113,7 +119,17 @@ def desired_contents
...
+      branches = branches.select do |branch|
```
haus: The select call could be lifted into its own method and then unit tested on its own. `filter_branches` or something of the sort.

nmaludy:
@andersonmills not sure how you want me to proceed with fixing the changes that @haus requested?

doc/dynamic-environments/configuration.mkd
nmaludy: What happens if they specify both filters, then which filter applies?

I would be more inclined to implement an additional option branch_filter_type that can either be include or exclude. Then the regex will be interpreted as either inclusive or exclusive.

doc/dynamic-environments/configuration.mkd
nmaludy: Yes, but it also doesn't matter. We're not actually using the group information so having 1 or 2 groups makes no difference.

doc/dynamic-environments/configuration.mkd
nmaludy: Yes, but it also doesn't matter. We're not actually using the group information so having 1 or 2 groups makes no difference.

doc/dynamic-environments/configuration.mkd
nmaludy: No, it doesn't matter in terms of functionality, but does help to logically separate out the prefix from the "extra stuff at the end".

andersonmills:
@nmaludy It was my intention to make the changes. :) Please feel free to comment on @haus 's comments, though.

doc/dynamic-environments/configuration.mkd
andersonmills: I'm a big fan of the simplicity of having one filter and using the regex negation. I expect that users getting to this level of usage will be able to follow the examples or be able to look it up quickly.

lib/r10k/source/git.rb
andersonmills: Making the change.

lib/r10k/source/git.rb
andersonmills: Opting for the former suggestion.

lib/r10k/source/git.rb
andersonmills: Workin' on that.

doc/dynamic-environments/configuration.mkd
haus: I feel like adding a setting that is a regex is probably a sign that we're taking the wrong approach to solving the problem. Do users actually need full regex matching, or would a simple allow/reject on a prefix or lists of allowed/denied environments be adequate?
Regexes tend to be simple in implementation, but not in actual usage, and debugging them can be challenging.

andersonmills: I can see users wanting to both be able to use a list of denied/allowed environments, and a list of denied/allowed prefixes. As I start thinking about how to document how this works for users, it gets ...uh... complicated. Do you have an configuration spec in mind?

haus: I would probably start with something like `ignore_branches` or `ignore_branch_prefixes` as a yaml array. I suspect that having a way to reject environments with a given prefix would be sufficient for most users.

haus: `ignore_branch_prefixes` could also be a list of environments to ignore.

andersonmills: I like it.

lib/r10k/source/git.rb
andersonmills: Well, I lifted it out, but it's a private method, as is the branch_names method. Should we be unit testing the private methods? I can, but I thought that was considered bad testing methodology.

lib/r10k/source/git.rb
```diff
@@ -111,9 +117,25 @@ def desired_contents
 
   private
 
+  def filter_branches(branches, ignore_prefixes)
+    filter = Regexp.new("^(#{ignore_prefixes.join('|')}).?")
+    branches = branches.select do |branch|
```
haus: i think if you use `reject` (https://ruby-doc.org/core-2.4.2/Enumerable.html#method-i-reject) instead of `select` here you can avoid the negation of `result` below.

lib/r10k/source/git.rb
```diff
@@ -111,9 +117,25 @@ def desired_contents
 
   private
 
+  def filter_branches(branches, ignore_prefixes)
+    filter = Regexp.new("^(#{ignore_prefixes.join('|')}).?")
```
haus: What is the `.?` doing in the regex? I assume it is an optional `any character`, but I'm not sure why it is needed.

lib/r10k/source/svn.rb
```diff
@@ -106,6 +112,13 @@ def names_and_paths
 
     branches << [R10K::Environment::Name.new('production', opts), "#{@remote}/trunk"]
     @svn_remote.branches.each do |branch|
+      if !@branch_filter.nil? && !@branch_filter.empty?
```
haus: I think the same change as was made to git could be applied here. (Both the conditional change and the ignore_branch_prefix change).

haus:
@andersonmills i don't have a great answer about testing private methods. I think testing the method seems nice because it is small and contained, but I think in that case it probably shouldn't be a private method? No strong feelings.

lib/r10k/source/git.rb
andersonmills: Changing.

lib/r10k/source/git.rb
andersonmills: It is not. Removing.

lib/r10k/source/svn.rb
andersonmills: D'oh! It's been so long since I paid attention to svn in r10k. Adding.

andersonmills:
@haus I made those methods public and wrote tests. Those methods are now tested individually, but with the same fail cases as the previous larger scale tests (which aren't really unit tests, anyway).

lib/r10k/source/git.rb
```diff
@@ -109,11 +115,27 @@ def desired_contents
     environments.map {|env| env.dirname }
   end
 
+  def filter_branches(branches, ignore_prefixes)
+    filter = Regexp.new("^(#{ignore_prefixes.join('|')})")
```
haus: This looks pretty good. I think `Regexp.new("^#{Regexp.union(ignore_prefixes)}")` is equivalent, but I'm not sure it's any cleaner.

spec/unit/source/git_spec.rb
```diff
@@ -62,6 +62,38 @@
       expect(master_env.dirname).to eq 'master'
     end
   end
+
+  describe "generating environments without the ignore_branch_prefixes" do
```
haus: I think this test description is a little funky. When i first read it I presumes it would be a test without using the ignore_branch_prefixes setting. Maybe `describe "generate_environments respects ignore_branch_prefixes setting"`?

spec/unit/source/git_spec.rb
```diff
@@ -62,6 +62,38 @@
...
+    end
+
+    it "uses the branch name as the directory by default" do
+      expect(environments[0].dirname).to eq 'master'
+      expect(environments[1].dirname).to eq 'production'
+    end
+  end
+
+  describe "filtering branches with ignore prefixes" do
+    let(:branches) { ['master', 'development', 'production', 'dev-test', 'dev', 'test_2'] }
```
haus: I think a test branch with dev or test at the end or the middle of a valid branch name would be good.

spec/unit/source/svn_spec.rb
```diff
@@ -61,6 +61,33 @@
       expect(environments[3].dirname).to eq 'robobutler'
     end
   end
+
+  describe "generating environments with branch filter exclude prefix" do
```
haus: This describe should probably be the same as whatever you end up with for git.

spec/unit/source/svn_spec.rb
```diff
@@ -61,6 +61,33 @@
...
+    end
+
+    it "uses the branch name as the directory by default" do
+      expect(environments[0].name).to eq 'production'
+      expect(environments[1].name).to eq 'master'
+    end
+  end
+
+  describe "filtering branches with ignore prefixes" do
+    let(:branches) { ['master', 'development', 'production', 'dev-test', 'dev', 'test_2'] }
```
haus: Same comment as in the git about having a test case with dev/test in the middle or end of a valid branch name.

haus:
This looks great. I had some feedback for the test cases, but I think this is good to merge after that?

spec/unit/source/git_spec.rb
andersonmills: Yeah, this has been through a few iterations to get here.

andersonmills: Yeah this went through too many editing iterations.

lib/r10k/source/git.rb
andersonmills: It's maybe a little clearer (maybe).

spec/unit/source/git_spec.rb
andersonmills: Good idea.

andersonmills: Adding this caused me to notice that "dev-test" is not a valid environment name, anyway.

andersonmills:
@haus I think this hits the test changes you wanted. :)

haus (approved):
Sweet. Looks great.

andersonmills:
Rebased after a merge up. But it seems like I shouldn't see all that, then. Hm. Noting that all the merged up commits have checks beside them already.

Aha! It looks like there is a problem caused by rebasing with commits shared between two users. I though it's going to be messy, I don't think there's a way around it.

Merging.

PR #779, opened 2017-11-21
(RK-308) Enable forge in Puppetfile to actually override one used.

cdenneen:
Thanks to @jearls for this work... now the `forge` line in a Puppetfile will actually attempt to pull down from that. 

This has been tested with and without a forge line.

### DIFF
--- a/lib/r10k/action/puppetfile/install.rb
+++ b/lib/r10k/action/puppetfile/install.rb
@@ -21,8 +21,23 @@ module R10K
 
         def visit_puppetfile(pf)
           pf.load!
+
+          original_forge = PuppetForge.host
+
+          if !pf.forge.nil?
+            if @settings[:forge][:allow_override]
+              logger.notice _("Using 'forge' value of '%{newforge}' declared in '%{puppetfile}'") % {puppetfile: pf.puppetfile_path, newforge: pf.forge}
+              PuppetForge.host = pf.forge
+            else
+              logger.notice _("Ignoring 'forge' value declared in '%{puppetfile}'") % {puppetfile: pf.puppetfile_path}
+            end
+          end
+
           yield
+
           pf.purge!
+        ensure
+          PuppetForge.host = original_forge
         end
 
         def visit_module(mod)
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -50,7 +50,6 @@ class Puppetfile
 
     @modules = []
     @managed_content = {}
-    @forge   = 'forgeapi.puppetlabs.com'
 
     @loaded = false
   end
@@ -74,6 +73,9 @@ class Puppetfile
 
   # @param [String] forge
   def set_forge(forge)
+    unless @forge.nil?
+      raise R10K::Error.new(_('You cannot declare multiple \'forge\' settings in the same Puppetfile'))
+    end
     @forge = forge
   end
 
--- a/lib/r10k/settings.rb
+++ b/lib/r10k/settings.rb
@@ -70,6 +70,12 @@ module R10K
         URIDefinition.new(:baseurl, {
           :desc => "The URL to the Puppet Forge to use for downloading modules."
         }),
+
+        Definition.new(:allow_override, {
+          :desc => "Whether or not to allow the Forge URL to be overridden by a 'forge' directive inside an environment Puppetfile.",
+          :default => lambda { false },
+          :normalize => lambda { |v| v.to_s.strip == 'true' },
+        }),
       ])
     end
 
--- a/r10k.yaml.example
+++ b/r10k.yaml.example
@@ -104,3 +104,8 @@ forge:
   # The 'baseurl' setting indicates where Forge modules should be installed
   # from. This defaults to 'https://forgeapi.puppetlabs.com'
   #baseurl: 'https://forgemirror.example.com'
+
+  # The 'allow_override' setting controls whether or not to allow the Forge
+  # 'baseurl' to be overridden by a 'forge' directive inside a Puppetfile.
+  # This defaults to 'false'.
+  #allow_override: true
--- a/spec/fixtures/unit/puppetfile/alternate-forge/Puppetfile
+++ b/spec/fixtures/unit/puppetfile/alternate-forge/Puppetfile
@@ -1,0 +1,3 @@
+forge 'http://forge.example.com/'
+
+mod 'puppetlabs/stdlib'
--- a/spec/unit/action/puppetfile/install_spec.rb
+++ b/spec/unit/action/puppetfile/install_spec.rb
@@ -9,6 +9,7 @@ describe R10K::Action::Puppetfile::Install do
 
   before(:each) do
     allow(puppetfile).to receive(:load!).and_return(nil)
+    allow(puppetfile).to receive(:forge).and_return(nil)
     allow(R10K::Puppetfile).to receive(:new).with("/some/nonexistent/path", nil, nil, nil, nil).and_return(puppetfile)
   end
 
--- a/spec/unit/puppetfile_spec.rb
+++ b/spec/unit/puppetfile_spec.rb
@@ -181,6 +181,14 @@ describe R10K::Puppetfile do
       subject = described_class.new(path)
       expect { subject.load! }.not_to raise_error
     end
+
+    it "accepts alternate forge" do
+      path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'alternate-forge')
+      pf_path = File.join(path, 'Puppetfile')
+      subject = described_class.new(path)
+      expect { subject.load! }.not_to raise_error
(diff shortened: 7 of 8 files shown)


rnelson0:
FWIW there’s an existing check https://github.com/puppetlabs/r10k/blob/0f68baf7c68da3072a2bb400945003481526f8a7/spec/unit/initializers_spec.rb that may be helpful to understand the history of this (as discussed on slack). I don’t know if the base url should be used, but either way there should probably be an especially test and this looks like a good one to copy.

cdenneen:
@rnelson0 This is a test for reading the r10k configuration and getting a configured `forge.baseurl` problem I'm having here is this is a parameter passed to the forge module which already has a test for the configuration being passed to it.
Since this is just reading that legacy line and now using it vs. noop, not sure how to test it. 
If you can find the noop test I can add the new condition but I haven't found one nor do I know how to properly test this.

rnelson0:
I'll see if I can find something soon, if not we can look at this next week during puppethack.

cdenneen:
Tests added

cdenneen:
@scotje @andersonmills can you please review?

adreyer:
If I remember correctly a lot of puppetfiles have an invalid 'forge' directive in them. Honoring it will cause them to break.

adreyer:
Does this forge take precedence over one configured in the config file? I think that would be a mistake in case people have site specific proxies they need to configure.

andersonmills:
Yep. I'm sorry, but I agree with @adreyer on this.

We have a forge directive in r10k.yaml, and it's not obvious which takes precedence in the deploy case. Even for the ```r10k puppetfile install``` case, how would proxies be handled?

This would have to be a 3.0.0 feature, even with that inclusion. There are currently no resources allocated to r10k v3.0.0 at this point. :(

cdenneen:
@andersonmills the `forge.baseurl` directive is global... this allows `bundle exec r10k puppetfile install` to use the environment/project based forge. So it is an environment option.

As far as proxies there is nothing stopping that to be configured in the `r10k.yaml`... this was just a way to take the existing `forge:` line in the `Puppetfile` and make use of it.

andersonmills:
I understand @cdenneen . But it also affects the ```r10k deploy environment``` experience as well. Does this take precedence? If so, we're probably going to break a lot of deploys, because of an invalid setting in current Puppetfiles. If not, we now have a Puppetfile that behaves differently in two different situations.

@scotje Suggested an "opt-in" flag or setting in for using the Puppetfile's forge setting in the case of ```r10k deploy environment```

andersonmills:
There are also some concerns in @scotje 's comments on https://tickets.puppetlabs.com/browse/RK-282

cdenneen:
@andersonmills @adreyer @scotje would it make more sense to use a different configuration, brand new, something like `forgeurl` so if `forgeurl` is found in Puppetfile then force `PuppetForge.host = forgeurl`? This would allow for all the old bad forge values to remain which would avoid breaking existing deployments. @andersonmills is there a way you could test the deploy environment piece once I've updated this PR?

scotje:
To me the simplest option would be to add a new config setting, something like `forge.allow_override` which defaults to `false`, but if set to `true` will then respect the `forge` directive in a `Puppetfile`.

A couple other things that I think would be nice would be adding information to the docs about the precedence of the Forge URL settings and logging when an alternate Forge URL is activated.

Lastly, I think it would encourage good practices to raise an error if there are multiple `forge` declarations evaluated in the same `Puppetfile`. (We are trying to reduce the statefulness of `Puppetfile` evaluation.)

Edit: And I don't think any of the above would require a 3.x

scotje:
(I can probably help with some of those additional changes as well.)

cdenneen:
@scotje that would be helpful... I was actually going down the route of allowing a `forge_url` attribute to be the one to override but if you think that's not worth it... and the `forge.allow_override` would be better then that's fine too...

scotje:
@cdenneen opened https://github.com/cdenneen/r10k/pull/2 with a pretty good start on those changes, let me know if you have any questions and I'll be happy to help!

scotje:
> @scotje the test needs to load the current allow_override to determine if it's allowed or not... currently it's passing the test even when it shouldn't be... since allow_override as you wrote it defaults to false then the current test shouldn't have PuppetForge.host set to the overrided value... it should return the default.

The way I implemented things, the `Puppetfile` class will always store the value of a `forge` directive and the decision about whether or not to honor the value from the Puppetfile is made inside the relevant actions. I feel like the act of parsing/evaluating the Puppetfile shouldn't have side effects on the configured `PuppetForge.host` value.

So I think in terms of additional things needed:

- Currently only `puppetfile install` respects the `allow_override` option and `forge` declaration, but at least `deploy environment` probably should as well.
- We should add tests cases to the tests for `puppetfile install` (and any other modified actions) to verify the `forge` declaration is honored or ignored in the correct circumstances.

> Also the default you removed. I didn't put that there... you think it's okay to remove that?

I did some grepping around and it didn't seem like that value was being used anywhere, and having it default to nil made it easier to detect multiple instance of the directive in a Puppetfile.

scotje:
Also, @andersonmills pointed out that we can rebase this on 2.6.x and then it can get released much sooner since master is 3.x at this point.

cdenneen:
@scotje I created a `enable_forge_head` branch to track original changes... and rebased this off `2.6.x` with the changes provided.
I've gone through a few attempts at the following:
- [ ] Update Puppetfile Action Install spec to include a test for `allow_override`:
https://github.com/cdenneen/r10k/blob/61aab3dda11d5fc0d4b1346a636d940da56ea12c/lib/r10k/action/puppetfile/install.rb#L22-L41
- [ ] Update Puppetfile Unit spec to simulate `allow_override = false` and `allow_override = true` which would return the `forge` value from fixture Puppetfile or default `https://forgeapi.puppet.com`
- [ ] I think for default to work when `allow_override = false` there would need to be some sort of else logic here: https://github.com/cdenneen/r10k/blob/61aab3dda11d5fc0d4b1346a636d940da56ea12c/lib/r10k/puppetfile.rb#L85-L92

Something like:
```
  # @param [String] forge
  def set_forge(forge)
    unless @forge.nil?
      raise R10K::Error.new(_('You cannot declare multiple \'forge\' settings in the same Puppetfile'))
    end
    unless R10K::Settings.forge_settings(:allow_override)
      @forge = 'https://forgeapi.puppet.com' # (or somehow load the default back from PuppetForge gem)
    else
      @forge = forge
  end
```

scotje:
I'll see if I can take a look at this again this week.

TJM:
We could really use this. I found out Artifactory can act as a PuppetForge, and configured it as such. The command line tests were promising, but when I pointed the "forge" option to the internal server in Puppetfile, imagine my surprise when it had no effect.

andersonmills:
Good news. This is currently in prioritization by our Products Team. For the short-term workaround, redirect using DNS, if that's an option.

@cdenneen - I see you're watching this. I'm hoping to be able to move on this VerySoonNow™.

cdenneen:
@scotje have you had anytime to look into this further? 

@TJM same setup here

andersonmills:
It looks like we're not 100% of the way there, but close. I agree that we're going to need that "else logic" you referred to.

TJM:
# Workaround

If you are using code manager, you can add the following to your hieradata (common.yaml or whatever) to configure r10k to use a custom baseurl, instead of using the "forge" line in the Puppetfile.

```
puppet_enterprise::master::code_manager::forge_settings:
  baseurl: https://artifactory.domain.com/artifactory/api/puppet/puppet
```

Or, if you are just using r10k, add the requisite config to your r10k.yaml however your might be managing that file:
```
forge:
  baseurl: https://artifactory.domain.com/artifactory/api/puppet/puppet
```

Reference: https://github.com/puppetlabs/r10k/blob/2.6.0/doc/dynamic-environments/configuration.mkd

NOTE: 2.6.0 appears to be the version currently in Puppet Enterprise 2017.3.5

andersonmills:
@TJM That workaround works for r10k's `deploy` command, but not the `puppetfile` command.

TJM:
@andersonmills Interesting. I would have figured it used the same code to retrieve modules from the forge. Sounds like a different bug, though. :)

andersonmills:
@TJM It actually does. What the `puppetfile` command doesn't do is read r10k.yaml.

TJM:
oops, that would be useful :)

cdenneen:
@mwaggett any reason for this to be closed?

mwaggett:
Hi @cdenneen, the 2.6.x stream of r10k is no longer supported, so I deleted the associated branch, which automatically closed this PR. 

Since this PR is extremely stale, if you are still interested in this feature, please either open a new PR against the default branch or update the [associated JIRA ticket](https://tickets.puppetlabs.com/browse/RK-308) to describe the feature you want to see and we will evaluate it.

Thanks for your contribution!

mwaggett:
@npwalker context for https://github.com/puppetlabs/r10k/pull/1134

PR #780, opened 2017-11-27
(RK-264) Change "update_force" to "force"

andersonmills:
This commit changes the --update_force flag to just a --force flag.
This improves the UX and puts it in line with the final implementation
of RK-305.

### DIFF
--- a/lib/r10k/action/puppetfile/install.rb
+++ b/lib/r10k/action/puppetfile/install.rb
@@ -10,7 +10,7 @@ module R10K
 
         def call
           @visit_ok = true
-          pf = R10K::Puppetfile.new(@root, @moduledir, @puppetfile, nil , @update_force)
+          pf = R10K::Puppetfile.new(@root, @moduledir, @puppetfile, nil , @force)
           pf.accept(self)
           @visit_ok
         end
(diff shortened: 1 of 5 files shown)


lib/r10k/action/puppetfile/install.rb
```diff
@@ -26,18 +26,18 @@ def visit_puppetfile(pf)
         end
 
         def visit_module(mod)
-          @update_force    = @update_force || false
+          @force    = @force || false
```
haus: I think this can be just `@force ||= false`? When false, this will be re-evaluated every time, but that's probably okay.

haus: The usage is similar to the `initialize` method in puppetfile.rb, but in this case the left and right hand sides reference the same variable, while in puppetfile.rb the right hand side refers to a parameter that might be nil while the left hand side refers to an instance variable.

haus (changes requested):
Looks good to me. I had one minor change but I'm good after that. Since it looks like this has never been released I have no problem changing the flag name.

lib/r10k/action/puppetfile/install.rb
andersonmills: Changed.

haus (approved):
Sweet! I'm in.

PR #783, opened 2017-12-01
Add boolean flag support to setopts

reidmv:
This commit adds support to setopts for :boolean flags. Specifying a
boolean flag means the flag when passed directly, --flag, is boolean
true. When the flag is denoted boolean, it will also be possible to pass
--no-flag, which will set the flag to false.

In the R10K::CLI module, if you want both true and false to be valid
flags you must reference both of them. (If you only want one value to be
passable, provide only the true or false version of the flag, as you
like). Example:

    module R10K::CLI
      module Example
        def self.command
          @cmd ||= Cri::Command.define do
            name    'example'

            flag nil, :'flag',    'Lets the user set flag to true'
            flag nil, :'no-flag', 'Lets the user set flag to false'

            run { |opts, args, cmd| exit 0 }
          end
        end
      end
    end

In allowed_initialize_opts, it is only necessary to define the flag
name once. E.g. for a flag "flag":

    def allowed_initialize_opts
      super.merge(:flag => :boolean)
    end

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -22,8 +22,9 @@ module R10K
 
           super
 
-          # @force here is used to make it easier to reason about
-          @force = !@no_force
+          # If neither --force nor --no-force was specified, set a default
+          @force = true if @force.nil?
+
           @argv = @argv.map { |arg| arg.gsub(/\W/,'_') }
         end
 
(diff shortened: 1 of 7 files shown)


andersonmills:
I think if we're going to do this for deploy/environment and deploy/module, then we need to do it for puppetfile/install also.

PR #794, opened 2018-01-04
(RK-311) Update yard dependency.

andersonmills:
r10k's dependency on yard needed to be updated due to a security
vulnerability seen in yard < 0.9.11. This updates the dependency to
~> 0.9.11.


cbrock:
👍

PR #797, opened 2018-01-08
Add mbaynton/SimplePuppetProvisioner to list of webhook listeners

mbaynton:
A modest documentation update ;)

I think it's worth a mention because it's simpler to deploy than alternatives that may require web servers and application frameworks, redis etc to be set up, and because it has other functionality like client certificate signing via http API that may be interesting to some users.


binford2k:
Invoking the trivial patch exemption for docs update.

PR #800, opened 2018-01-12
(CODEMGMT-1152) Implement basemod Puppetfile feature

reidmv:
A basemod is a git repository checked out into the environment's basedir, layered alongside any content defined directly in the control-repo. The intended use case is to reduce the content in a control-repo to ONLY the Puppetfile. More information in CODEMGMT-1152.

### DIFF
--- a/doc/puppetfile.mkd
+++ b/doc/puppetfile.mkd
@@ -305,3 +305,36 @@ mod 'apache',
 The given 'install\_path' can be an absolute path or a path relative to the base of
 the environment. Note that r10k will exit with an error if you attempt to set the
 'path' option to a directory outside of the environment.
+
+### Basemod
+
+The special directive `basemod` can be used to deploy the contents of a Git
+repository to the directory containing the Puppetfile, rather than to a
+`modules/` (or `modulepath`) subdirectory. Using basemod allows the contents of
+the control repo to be reduced to _only_ the Puppetfile, and for all other
+content (environment.conf, hiera.yaml, site-modules) to be versioned and
+developed in a separate repo as if they too were just another module.
+
+#### Example
+
+```
+##
+# SITE-SPECIFIC CONTENT
+#
+# Main environment content, like hiera.yaml and environment.conf.
+# Should be assigned a specific version; tag or commit (except when developing)
+#
+basemod 'site',
+  :git => 'https://github.com/reidmv/puppet-site-code.git',
+  :ref => '1.0'
+
+##
+# MODULES
+#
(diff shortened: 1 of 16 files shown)


reidmv:
Closing in favor of alternative proposed in [CODEMGMT-1177](https://tickets.puppetlabs.com/browse/CODEMGMT-1177).

In the event this ever gets re-investigated, note that 4c2d0ad is a problem commit. Need to revert that and find a more reliable way of dealing with warnings.

c33s:
https://tickets.puppetlabs.com/browse/CODEMGMT-1177 -> `Permission violation`

reidmv:
@c33s thanks for pointing that out. Unfortunately the CODEMGMT project in Jira seems to be internal-only. The description of the issue as currently written is as follows.

The biggest improvement workflow-wise this idea has over something like basemod is that it isn't necessary to separate the Puppetfile from the rest of a control-repo. It is still possible to define deployment fully independent of development.

The end-goal here is not a feature, per se, but a workflow improvement. CODEMGMT-1177 is a better and more flexible workflow than CODEMGMT-1152, thus obviating this PR. 1177 came about following a group discussion of significant problems that 1152 doesn't solve.

---
# CODEMGMT-1177
## "link" control-repo branch to any git ref via Puppetfile

Implement a feature wherein a branch of a control-repo may optionally contain a single Puppetfile as content, with a directive pointing to a different version of the control-repo to deploy as the named environment.

For example:

```ruby
#=============================================================================
# ENVIRONMENT: production
#=============================================================================

##
# ENVIRONMENT CODE VERSION
#
# Should be assigned a specific version (tag or commit). R10K will populate
# the production environment directory with code from the control-repo at the
# given ref.
#
environment 'production',
    :ref => '1.1'
```

This feature would allow customers to use normal Git development workflows for Puppet code, while simultaneously providing a means to deploy versions or tags of that code to named environments.

If the feature is extended to accept normal module source parameters, such as `:git`, it would become possible to have a puppet-deployment repository for production environments entirely separate from the repository used to develop the code.

```ruby
environment 'production',
    :git => 'https://github.com/reidmv/puppet-site-code.git',
    :ref => '1.1'
```

In the event an environment is deployed containing a Puppetfile with an `environment` stanza:

* R10K should require that the Puppetfile contain no other DSL directives, such as `mod` or `forge`.
* R10K should require that the name given the environment stanza matches the name of the Git branch in the control-repo.
* R10K should substitute content checkout from the active control-repo branch for content checkout from the version (and optionally source) the `environment` directive specifies.

PR #802, opened 2018-02-19
(CODEMGMT-1177) Add environment definition feature

reidmv:
Opening PR for visibility and collaboration. Code works. Needs Engineering review.

README is rough but gives an idea of what this does.
https://github.com/reidmv/r10k/blob/CODEMGMT-1177/doc/puppetfile.mkd#environment

Ticket:
https://tickets.puppetlabs.com/browse/CODEMGMT-1177

### DIFF
--- a/doc/puppetfile.mkd
+++ b/doc/puppetfile.mkd
@@ -315,8 +315,7 @@ Example: the "production" branch of a control repository should have version 3.2
 To link the production environment branch to a code version:
 
 1. The production branch should have one, and only one, file committed. The Puppetfile.
-2. The Puppetfile should have only one content item in it: the environment declaration.
-3. The environment declaration in the Puppetfile must match the git branch name
+2. The environment declaration in the Puppetfile must match the git branch name
 
 For example:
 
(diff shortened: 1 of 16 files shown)


danzilio:
I love this

doc/puppetfile.mkd
```diff
@@ -305,3 +305,142 @@ mod 'apache',
...
+
+## Environment
+
+When using a Git control repo, branches, and a Puppetfile, it is possible to "link" the branch to an environment version, rather than comitting code to the branch. This advanced feature may be used to aid in deployment workflows.
+
+Example: the "production" branch of a control repository should have version 3.2.0 of the puppet environment code deployed. Development of puppet code is happening constantly in the master branch, but when releases are ready they are tagged so that the code may be deployed to upper-tier environments. Ensuring that the version of code checked out to the production branch matches a given tag or commit is a process that requires a high level of Git knowledge. A simpler approach to controlling that deployment may be to make the production branch link to a code version instead.
+
+To link the production environment branch to a code version:
+
+1. The production branch should have one, and only one, file committed. The Puppetfile.
```
binford2k: Can I suggest a little less draconian rule? It should whitelist files like `README.md`, for example. This would allow users to document their workflow in the place where it's used.

reidmv: @binford2k What would you think about documenting the workflow in the Puppetfile? [Example](https://github.com/reidmv/puppet-deployment/blob/integration/Puppetfile)

(note: this example was created for an old workflow iteration, ignore the actual instructions in it)

reidmv: The concern with allowing more than the Puppetfile is potential conflicts and "flickering" during deployments with files of the same name in the code the Puppetfile "links" to. This already happens for the Puppetfile itself, of course.

It's also just an implementation limitation. In theory it could be engineered around.

binford2k: I'm not fully opposed to the idea, but certain files already have known meaning. For example, the nice pretty rendered view that GitHub gives you of `README.md`.

reidmv:
Closing in favor of #983. 

#983 implements a cleaner generalization of the pattern/capability, which can then be easily built on.

PR #805, opened 2018-03-05
Quickstart update: Add hieradata path to hiera.yaml

WaaZaa666:
Current manual is not working with a fresh install, the hiera path needs to be set in hiera.yaml


binford2k:
This is the default Hiera path. It shouldn't need to be specified.

PR #808, opened 2018-03-29
fix link to puppet.com/blog

xorpaul:


binford2k:
The `workflow` source is defined on line #L11. This change isn't needed. Thanks though!

PR #812, opened 2018-05-15
(RK-317) Reorder git source selection

ananace:
This changes the order in which the source reference is chosen - prioritizing them in the descending order of accuracy instead.


binford2k:
Hey @ananace! Thanks for the PR. We now warn when specifying conflicting options (See #1130), so the search order isn't as relevant.

PR #814, opened 2018-05-25
(maint) Update spec so r10k vanagon promotion may succeed

amysahli:


CHANGELOG.mkd
```diff
@@ -1,5 +1,10 @@
 CHANGELOG
 =========
+2.6.3
+----
+## Changes
+
+Update specs to allow r10k-vanagon promotion.
```
Magisus: I would instead say "Update specs with new error string", probably. Since this repo doesn't really care that it winds up being packaged in vanagon.

amysahli: Done.

Magisus (approved):

PR #820, opened 2018-06-25
(RK-319) Do not expose module tmpdirs as separate from the full path

Magisus:
This commit removes the addition of accessors for the unpack and
download path containing directories, to maintain the fixed relationship
between these paths and their root dirs.

### DIFF
--- a/lib/r10k/forge/module_release.rb
+++ b/lib/r10k/forge/module_release.rb
@@ -28,10 +28,6 @@ module R10K
       #   @return [Pathname] Where the module tarball will be downloaded to.
       attr_accessor :download_path
 
-      # @!attribute [rw] download_root
-      #   @return [Pathname] Directory where the module tarball will be downloaded to.
-      attr_accessor :download_root
-
       # @!attribute [rw] tarball_cache_path
       #   @return [Pathname] Where the module tarball will be cached to.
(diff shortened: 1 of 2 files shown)


amysahli (approved):

lib/r10k/forge/module_release.rb
```diff
@@ -48,10 +44,6 @@ class ModuleRelease
       #   @return [Pathname] Where the module will be unpacked to.
       attr_accessor :unpack_path
 
-      # @!attribute [rw] unpack_root
-      #   @return [Pathname] Directory where the module will be unpacked to.
-      attr_accessor :unpack_root
-
       # @param full_name [String] The hyphen separated name of the module
       # @param version [String] The version of the module
       def initialize(full_name, version)
```
MikaelSmith, line 49: No longer need to set `download_root` and `unpack_root` in the initializer.

MikaelSmith (changes requested):

lib/r10k/forge/module_release.rb
Magisus: Thanks, updated.

MikaelSmith (approved):

MikaelSmith:
If CI passes, go ahead and merge.

PR #823, opened 2018-07-07
add filter_command (e.g. for gpg signature validation) (git only)

mhumpula:
please consider for merging.

## motivation
In my current puppet setup, the git repository is hosted on the shared git manager with wide user base. Though the repository has correct permissions configured I would like to be able to add some extra checks. Specifically stop the deployment of the code, that is not signed with proper gpg key by using built-in git gpg funkcionality.

The check script is easy, but I need to intercept the r10k deployment between fetching the latest repository state to *cache* and the actual deployment to puppet environments. So I came up with the branch filtering hooks, which IMHO can be used by larger user base for filtering branches and environments based on their requirements.

## implementation

I've tried to model this in the same fashion as `ignore_branch_prefixes`. I've only done git implementation, because I know very little about the svn.


anarcat:
any reason why this has not been merged yet? interest?

does this work with forge modules?

mhumpula:
@anarcat I'm using this for two years now with simple script check https://gist.github.com/mhumpula/10b7a333d2a9a470f5e0e44beb5c91f0 and it works ok for my requirements. The checkout of puppet modules is done in different part of the code I believe, so it's not affected by this change.

dhollinger (changes requested):
Just one little change in the documentation for the change. Otherwise looks good.

doc/dynamic-environments/configuration.mkd
```diff
@@ -381,6 +381,24 @@ sources:
       - 'dev'
 ```
 
+### filter_command
+
+You can filter out any branch based on the result of the command specified as 'filter_command'. Non zero return
+status of the command results in a branch beeing removed. The command is passed additional environment variables
+* GIT_DIR – path to the cached git repository
+* R10K_BRANCH – brach which is beeing filtered
```
dhollinger: Typo here. I think you meant `branch` and not `brach`

binford2k:
```suggestion
* R10K_BRANCH – branch which is being filtered
```

doc/dynamic-environments/configuration.mkd
```diff
@@ -381,6 +381,24 @@ sources:
...
+* R10K_NAME – source name from r10k configuration
+
+This can be used for example for filtering out the branches with invalid gpg signature of their latests commit
```
binford2k:
```suggestion
This can be used for example for filtering out the branches with invalid gpg signature of their latest commit
```

binford2k (changes requested):
I like it. Small typo fixes.

mwaggett:
hey @mhumpula would you mind adding a changelog entry to the 'Unreleased' section describing this change?

also @dhollinger @binford2k would love if y'all could take another look at this.

binford2k:
@mhumpula Can you also add some docs to clarify that this only works when using the git source? We'll file a ticket for the missing svn counterpart when it's merged. Otherwise, 👍

mhumpula:
@binford2k note about git added, @mwaggett added entry to changelog, but I'm not sure if it is in format you like.

mwaggett:
hey @mhumpula! Would you mind rebasing against the master branch? There have been many updates to the changelog since this PR was first opened. 😅  And then, usually, we add the PR as a link at the end of the note if there wasn't a ticket the work was based on, for example: https://github.com/puppetlabs/r10k/blob/master/CHANGELOG.mkd#350

Let us know if you need help with any of that! And thanks again for your contribution!

binford2k:
I went ahead and resolved the conflicts for you. One final review from @dhollinger and we'll get this merged! Thanks so much for the contribution @mhumpula 🎉

dhollinger (approved):

binford2k (approved):

PR #829, opened 2018-07-25
(RK-321) Various test fixes

Magisus:


amysahli (approved):

Magisus:
I tested these locally on both ubuntu1804 and centos6 (one of which is python 3, the other 2).

integration/tests/Puppetfile/HTTP_PROXY_effects_git_source.rb
```diff
@@ -19,7 +19,7 @@
 
 puppetfile =<<-EOS
 mod 'motd',
-  :git    => 'https://github.com/puppetlabs/puppetlabs-motd'
+  :git    => 'git://github.com/puppetlabs/puppetlabs-motd'
```
justinstoller: Does this still test the same thing? iow, Does it have to be https to test the proxy?

justinstoller: it looks like it's expecting failure below but with a failure that it could use the proxy correctly. I wonder why it's even falling back to github if the proxy doesn't work?

justinstoller: Ahh, I don't think this was failing:
https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/job/experimental_auto_r10k_intn-van-sys_irving/GIT_PROVIDER=shellgit,SLAVE_LABEL=beaker,TEST_TARGET=sles11-64.mdca/2/#showFailuresLink

not a big deal, and doesn't hurt to change it, was just curious

integration/tests/git_source/HTTP_proxy_and_git_source.rb
```diff
@@ -16,7 +16,7 @@
 
 puppetfile =<<-EOS
 mod 'motd',
-  :git    => 'https://github.com/puppetlabs/puppetlabs-motd'
+  :git    => 'git://github.com/puppetlabs/puppetlabs-motd'
```
justinstoller, line 19: same as above wrt proxies

integration/tests/Puppetfile/HTTP_PROXY_effects_git_source.rb
Magisus: Oh, whoops. I'll put this one and the other proxy one back. Thanks for the catch!

justinstoller (approved):

integration/component/pre-suite/05_install_dev_r10k.rb
```diff
@@ -1,10 +1,12 @@
 test_name "Install PE r10k" do
 
   step "Install PE r10k" do
+    variant, version, arch, codename = master['platform'].to_array
+    if variant == 'ubuntu' && version.split('.').first.to_i >= 18
+      on master, "echo 'Acquire::AllowInsecureRepositories \"true\";' > /etc/apt/apt.conf.d/90insecure"
```
justinstoller, line 6: Oh, nit: we're not installing the signing key with this commit, we're instructing apt to ignore the fact we don't have it.

integration/tests/Puppetfile/HTTP_PROXY_effects_git_source.rb
Magisus: Going to run both versions first, see if it made any difference.

Magisus:
Ran these through on SLES 11 now also, successfully.

Issue #697, opened 2017-01-25
Expose environment in postrun


jkroepke:
When I deploy a specific branch like `r10k deploy environment prod -pv`, it would be awsome to expose the environment prod to to postrun process, for example as environment variable, eg.

shaheed121:
This option is much needed. For eg: To run puppet generate types for the environment we are deploying.

elconas:
This is important as it is needed for a post provisioning "puppet generate type --environment ENV" command.

mwaggett:
fixed by https://github.com/puppetlabs/r10k/pull/836

PR #836, opened 2018-08-16
Support $environment variable in postrun

raphink:
This implements https://github.com/puppetlabs/r10k/issues/697, allowing to pass `$environment` as a `postrun` hook to run e.g. `puppet generate types --environment $environment`.

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -64,6 +64,11 @@ module R10K
           end
         ensure
           if (postcmd = @settings[:postrun])
+            if postcmd.grep('$modifiedenvs').any?
+              envs = deployment.environments.map { |e| e.name }
+              envs.reject! { |e| !@argv.include?(e) } if @argv.any?
+              postcmd = postcmd.map { |e| e.gsub('$modifiedenvs', envs.join(' ')) }
+            end
             subproc = R10K::Util::Subprocess.new(postcmd)
             subproc.logger = logger
             subproc.execute
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -183,6 +183,9 @@ postrun: ['/usr/bin/curl', '-F', 'deploy=done', 'http://my-app.site/endpoint']
 
 The postrun setting can only be set once.
 
+Occurrences of the string `$modifiedenvs` in the postrun command will be
+replaced with the current environment(s) being deployed.
+
 ### sources
 
 The `sources` setting specifies what repositories should be used for creating
--- a/spec/unit/action/deploy/environment_spec.rb
+++ b/spec/unit/action/deploy/environment_spec.rb
@@ -92,6 +92,61 @@ describe R10K::Action::Deploy::Environment do
           subject.call
         end
       end
+
+      context "supports environments" do
+        context "when one environment" do
+          let(:settings) { { postrun: ["/generate/types/wrapper", "$modifiedenvs"] } }
+          let(:deployment) { R10K::Deployment.new(mock_config.merge(settings)) }
+
+          before do
+            expect(R10K::Deployment).to receive(:new).and_return(deployment)
+          end
+
+          subject do
+            described_class.new( {config: "/some/nonexistent/path" },
+                                 %w[first],
+                                 settings                             )
+          end
+
+          it "properly substitutes the environment" do
+            mock_subprocess = double
+            allow(mock_subprocess).to receive(:logger=)
+            expect(mock_subprocess).to receive(:execute)
+
+            expect(R10K::Util::Subprocess).to receive(:new).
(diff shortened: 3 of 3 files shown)


lib/r10k/action/deploy/environment.rb
```diff
@@ -64,6 +64,7 @@ def visit_deployment(deployment)
           end
         ensure
           if (postcmd = @settings[:postrun])
+            postcmd.map! { |e| e.gsub('$environment', deployment.environments) }
```
alexjfisher: @raphink 
For me, this code fails with `no implicit conversion of Array into String`

raphink: Hmmm. Should the `postrun` command be run for every environment if `$environment` is found?

raphink:
Not fully working yet, but it'll be fixed soonish :smile:

raphink:
There, this is working now /cc @alexjfisher

raphink:
As discussed in https://github.com/xorpaul/g10k/pull/111#issuecomment-415755935, another approach to dealing with multiple environments would be to pass them all to the script in one go and let the script loop over them. What would be the preferred method?

bastelfreak (approved):

raphink:
Does this require more changes?

raphink:
Any news of this? This PR would help a lot…

bastelfreak:
@binford2k can you have a look at this and inform whoever is responsible here?

raphink:
FYI, I've started successfully using this in production (via the `camptocamp/r10k:3.0.3_postrun-3` container image)

bastelfreak:
@scotje @binford2k could you have a look at this please?

justinstoller:
Hey, thanks for opening this and pinging binford. He got a hold of me and I'll be taking a look at this over the next day!

justinstoller:
I applied this patch to a PE install (my goto testbed) and configured it to let me use r10k directly with this config:
```
[root@qkf8arr4p0kbxh0 ~]# cat /etc/puppetlabs/r10k/r10k.yaml
---
cachedir: /opt/puppetlabs/server/data/puppetserver/r10k
sources:
  puppet:
    basedir: /etc/puppetlabs/code/environments
    remote: https://github.com/puppetlabs/puppetlabs-puppetserver_perf_control
postrun: ["/tmp/run-generate-types.sh", "$environment"]
```

```
[root@qkf8arr4p0kbxh0 ~]# cat /tmp/run-generate-types.sh
for environment in $1; do
  echo "$environment"
  /opt/puppetlabs/bin/puppet generate types --environment $environment
done
```
I wasn't able to get this actual patch to work for me however. I had to update grep to use a regex rather than a string, ie
```
[root@qkf8arr4p0kbxh0 ~]# grep grep /opt/puppetlabs/puppet/lib/ruby/gems/2.5.0/gems/r10k-3.0.3/lib/r10k/action/deploy/environment.rb
            if postcmd.grep(/\$environment/).any?
```

Having done that it works great!

So I have three thoughts on this:
1. It seems like the grep call needs to be updated before we merge
2. The set up seems a little more involved than the impression I got reading "a postrun hook to run e.g. puppet generate types --environment $environment." and would like to see some docs around that. Specifically: the "$environment" variable 
 should be its own element in the postcmd array and "$environment" may be a space separated list, consequently the generate types command should be wrapped to iterate over those environemnts. I don't know if documenting all that in the in repo docs is necessary for this PR but something a little more than what exists now would be nice.
3. Related to 1, this should have tests. However, it doesn't look like there are any unit tests for any of the postcmd feature. And I don't expect you to backfill those tests as a prerequisite for this PR. But I will write those backfill tests today or tomorrow and either ask you to then add tests for this feature to them or also write tests for this feature while I'm backfilling postcmd tests and push them to your branch.

justinstoller:
I opened a PR against @raphink 's branch with some tests for this. I think with that and a little more info in the docs (maybe an example of configuring it for generate types since that seems like something most folks will want to use it for?) this will be ready for merge.

justinstoller:
Actually, looking at all t's to cross and i's to dot. Can I get you to squash a few of those commits and preface them with `(maint)` or `(GH-697)` too?

(let me know if you want help with that)

raphink:
Thanks @justinstoller , I merged your PR.

raphink:
Squashed and rebased. Now you might also want to consider how this was done in g10k in case keeping it compatible would make sense https://github.com/xorpaul/g10k/pull/113

justinstoller:
Thanks for pointing that out @raphink. It looks like for the extent of what is being done in the PR changing the substituted string from "$environment" to "$modifiedenvs" would keep it compatible with g10k? I'd totally be for that.

raphink:
Ok, let's do that then.

raphink:
There, fixed.

justinstoller:
Right on, sorry for the delay, got sick end of last week.

PR #840, opened 2018-08-23
(maint) Cherry-pick docker-related commits from 2.5.x

underscorgan:


docker/r10k/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM ubuntu:16.04
+
+ARG vcs_ref
+ARG build_date
+ARG version="2.5.5"
```
shrug: Does this just set a default if one isn't passed in?

mwaggett (approved):
if this is just cherry-picks, seems good to me

docker/r10k/Dockerfile
underscorgan: yeah, but I should update this to at least something in the 2.6 series probably, will update that

shrug (approved):

PR #845, opened 2018-09-11
(maint) Update spec tests to not use docker-api gem

underscorgan:


underscorgan:
depends on https://github.com/puppetlabs/puppet_docker_tools/pull/26

Iristyle:
#847 has now passed, which means this should be good to go!

PR #850, opened 2018-09-21
(maint) let the container run for longer

underscorgan:


underscorgan:
depends on https://github.com/puppetlabs/puppet_docker_tools/pull/28

docker/Gemfile
```diff
@@ -10,5 +10,5 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem 'puppet_docker_tools', *location_for(ENV['PUPPET_DOCKER_LOCATION'] || '~> 0.2.0')
+gem 'puppet_docker_tools', *location_for(ENV['PUPPET_DOCKER_LOCATION'] || '~> 0.2.1')
```
Iristyle: Rather than bumping these in small increments every time we release updates to the gem, want to just float it up to 1.0 with `~> 0.2`?

underscorgan: I think I ran into issues on this PR in particular due to the workspace existing before 0.2.1 was released, but ~> 0.2 should be fine, will update

PR #861, opened 2018-10-03
Add option to install modules concurrently

oesgalha:
This pull request add the option to install/update modules concurrently.

It adds a `pool_size` setting with a default value equals to `1`:

```yaml
# /etc/puppetlabs/r10k/r10k.yaml
pool_size: 5
```

When the `Puppefile#accept` method is called, if the `pool_size` value is 1, the regular sequential install/update process starts. If the `pool_size` is greater than 1 a thread pool with `pool_size` threads is created to fetch the modules. A `pool_size` with value less or equal to zero is not allowed.

Since the default setting value is 1 this feature is _opt-in_.

The install/update process is heavily I/O based so there are some visible performance gains in the total (wall) time to install modules declared in a Puppetfile.

We've been using this patch for almost 5 months in production with no reported problems.

## Benchmarks

Puppetfile with 9 modules

|Flavor|Time|Comparison|
|--------|-------|----------------|
|Vanilla|69.846 s|base|
|pool_size=5|19.985  s|-71.39%|

Puppetfile with 608 modules

|Flavor|Time|Comparison|
|--------|-------|----------------|
|Vanilla|8.705 m|base|
|pool_size=5|2.056  m|-76.38%|
|pool_size=10|1.901  m|-78.16%|

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -194,7 +194,8 @@ class Puppetfile
     Thread.new do
       begin
         while mod = mods_queue.pop(true) do mod.accept(visitor) end
-      rescue ThreadError
+      rescue ThreadError => e
+        logger.error _("Thread error during concurrent module deploy: %{message}") % {message: e.message}
         Thread.exit
       end
     end
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -1,3 +1,4 @@
+require 'thread'
 require 'pathname'
 require 'r10k/module'
 require 'r10k/util/purgeable'
@@ -6,6 +7,10 @@ require 'r10k/errors'
 module R10K
 class Puppetfile
   # Defines the data members of a Puppetfile
+
+  include R10K::Settings::Mixin
+
+  def_setting_attr :pool_size, 1
 
   include R10K::Logging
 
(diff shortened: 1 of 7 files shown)


smortex:
WOW! We want this :smile: !

Quick numbers from my setup with no `modules` directory (cold start, 57 modules to download):

| Flavor      | Time    | Comparison |
|-------------|---------|------------|
| Vanilla     | 1:25,38 | base       |
| pool_size=5 |  24,848 | - 70.90 %  |

With all modules already here (hot start):

| Flavor      | Time   | Comparison |
|-------------|--------|------------|
| Vanilla     | 10,668 | base       |
| pool_size=5 | 3,630  | - 65.97 %  |

You may want to consider rebasing this on top of master and perhaps allow setting `pool_size` from the command line (e.g. `-j5`).

Awesome :hugs: !

lib/r10k/puppetfile.rb
```diff
@@ -146,14 +151,48 @@ def purge_exclusions
...
+        modules.each { |mod| queue << mod }
+      end
+    end
+  end
+
+  def visitor_thread(visitor, mods_queue)
+    Thread.new do
+      begin
+        while mod = mods_queue.pop(true) do mod.accept(visitor) end
+      rescue ThreadError
```
Sharpie: What sorts of errors could trigger this rescue? Should they be logged?

reidmv:
Rebased this against the current master

Sharpie (approved):
Looks good 👍

PR #866, opened 2018-10-11
(RK-324) Improve reliability on Ubuntu Bionic

justinstoller:
Previously on Bionic we would see issues where the Ruby interpreter
would die with an "ASYNC BUG EPIPE" message and no meaningful
stacktrace.

When attempting to strace, when registering listeners for various
syscalls with sysdig, or when attaching to a debug build of Ruby with
gdb this issue no longer presents itself.

This seems to happen when forking during the while loop of
R10K::Util::Subprocess::Pump#pump though smaller reproducers that only
run the while loop and fork seem to rarely ever trigger this issue,
there maybe be some requirement on calling IO#reopen as we do on STDOUT
and STDERR.

This commit moves starting the pipe output pumping until after the
subprocess fork which seems to avoid the problematic code.

### DIFF
--- a/lib/r10k/util/subprocess/runner/posix.rb
+++ b/lib/r10k/util/subprocess/runner/posix.rb
@@ -28,15 +28,15 @@ class R10K::Util::Subprocess::Runner::POSIX < R10K::Util::Subprocess::Runner
 
     @stdout_pump = R10K::Util::Subprocess::Runner::Pump.new(@stdout_r)
     @stderr_pump = R10K::Util::Subprocess::Runner::Pump.new(@stderr_r)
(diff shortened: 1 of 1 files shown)


Magisus:
Thanks for tracking this down!

branan:
Generally speaking any threading/async stuff and `fork` don't play well together - and Ruby does a bunch of its own magic around forking. I'm not surprised that moving things around fixes it :/

Magisus:
We should merge this to 2.6.x

justinstoller:
updated

Magisus (approved):

PR #871, opened 2018-11-12
(maint) Move to makefile for build

underscorgan:


docker/Makefile
```diff
@@ -0,0 +1,51 @@
...
+lint:
+ifeq ($(hadolint_available),0)
+	@$(hadolint_command) r10k/$(dockerfile)
+else
+	@docker pull $(hadolint_container)
+	@docker run --rm -v $(PWD)/r10k/$(dockerfile):/Dockerfile -i $(hadolint_container) $(hadolint_command) Dockerfile
+endif
+
+build: prep
+	@docker pull $(base_image)
```
nwolfe: Do we really need this?

underscorgan: yeah, otherwise we won't get updates when new ubuntu containers are published since we have shared build hardware

nwolfe: Is there another way we can solve that that doesn't require us assuming the base image here?

Maybe we can instead use some flag like `--no-cache` on our docker invocations? Or purge the images?

underscorgan: To follow up with some context from chat: we're going to leave this as is for now, and in the future move the `docker pull` to a separate step that we manually invoke in ci. We don't want to run with `--no-cache` because we do want to be able to take advantage of layer caching, especially with our multi-staged/compiled builds, and methods to programatically figure out the base image are complicated by multi-staged builds.

docker/Makefile
```diff
@@ -0,0 +1,47 @@
...
+ifeq ($(IS_LATEST),true)
+	@docker tag puppet/r10k:$(version) puppet/r10k:latest
+endif
+
+test: prep
+	@bundle install --path .bundle/gems
+	@PUPPET_TEST_DOCKER_IMAGE=puppet/r10k:$(version) bundle exec rspec r10k/spec
+
+publish: prep
+	@docker push puppet/r10k:$(version)
```
Iristyle, line 42: I made a suggestion in the `puppetserver` PR to do something like this in the `IS_NIGHTLY` check on line 8:

```make
container_version = latest
```

And then use a conditional fallback here to get rid of the branching like:

```make
container_version ?= $(version)
@docker push puppet/r10k:$(container_version)
```

But we can always make that update later

Issue #693, opened 2017-01-08
Support running `generate types` after deploying an environment


alexjfisher:
Over in this [thread](https://github.com/voxpupuli/puppet-r10k/issues/293), we were wondering if R10K itself could be enhanced to support running `generate types` after deploying an environment.

This [documentation](https://docs.puppet.com/puppet/latest/environment_isolation.html#enable-environment-isolation-with-r10k) suggests running `generate types` using the [`postrun`](https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#postrun) option, but AFAICT, that is only run when deploying *all* environments; it is not suitable when just deploying a single environment.

Perhaps there are already plans to add this functionality?

Thanks.

raphink:
`puppet generate types` can be ran with `--environment <env>` to generate for individual environments. The problem is that [`postrun` doesn't allow to pass the environment name](https://github.com/puppetlabs/r10k/issues/697)

PR #874, opened 2018-11-16
Add support for running `puppet generate types`

nabertrand:
This adds support for running `puppet generate types` after a successful environment or module deployment. See https://puppet.com/docs/puppet/latest/environment_isolation.html
for more information on generated types.

Should solve #693.

### DIFF
--- a/lib/r10k/action/runner.rb
+++ b/lib/r10k/action/runner.rb
@@ -42,10 +42,23 @@ module R10K
       def setup_settings
         config_settings = settings_from_config(@opts[:config])
 
-        overrides = {:cachedir => @opts[:cachedir]}
-        overrides.delete_if { |_, val| val.nil? }
+        overrides = {
+          cachedir: @opts[:cachedir],
+          deploy: {
+            puppet_path: @opts[:'puppet-path'],
+            generate_types: @opts[:'generate-types']
+          }
+        }
 
         with_overrides = config_settings.merge(overrides) do |key, oldval, newval|
+          break oldval if newval.nil?
+          if oldval.is_a? Hash
+            newval.merge!(oldval) do |_, subold, subnew|
+              subnew.nil? ?
+                subold :
+                subnew
+            end
+          end
           logger.debug2 _("Overriding config file setting '%{key}': '%{old_val}' -> '%{new_val}'") % {key: key, old_val: oldval, new_val: newval}
           newval
         end
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -297,6 +297,26 @@ deploy:
   write_lock: "Deploying code is disallowed until the next maintenance window (2038-01-19)"
 ```
 
+#### generate\_types
+
+The `generate_types` setting controls whether r10k should update generated types
+after a successful environment update. See [Environment isolation](https://puppet.com/docs/puppet/latest/environment\_isolation.html)
+for more information on generated types. Defaults to false.
+
+```yaml
+deploy:
+  generate_types: true
+```
+
+#### puppet\_path
+
+The path to the puppet executable used for generating types. Defaults to `/opt/puppetlabs/bin/puppet`.
+
+```yaml
+deploy:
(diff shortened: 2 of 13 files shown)


dhollinger:
Why not utilize `Puppet::Generate::Type.generate` instead of shelling out to a command?

nabertrand:
As far as I can tell using `Puppet::Generate::Type.generate` would require loading the puppet gem, configuring the current environment, and searching for inputs with `Puppet::Generate::Type.find_inputs`. Is it worth adding the puppet gem and its dependencies to r10k's dependencies? I also wonder if the `puppet generate types` command interface is more likely to remain stable than the arguments to `Puppet::Generate::Type.generate` and `Puppet::Generate::Type.find_inputs`. I also see the merit in a pure ruby solution.

dhollinger:
I'd argue that the best options would be to either use the puppet gem to execute generate types from within r10k or utilize the `postrun` configuration option to execute `puppet generate types` directly after the r10k run completes.

binford2k:
My preference would be to invoke `Puppet::Generate::Type.generate` directly as well, since it would avoid the ruby startup cost of shelling out. It's only a couple seconds or so, but that could easily add 30s or more to a deploy with more than a few environments changing.

```
$ time puppet generate types --environment production --environmentpath /etc/puppetlabs/code/environments/
Notice: Generating Puppet resource types.
[...]

real	0m2.013s
user	0m1.631s
sys	0m0.346s
```

nabertrand:
`Puppet::Generate::Type.generate` loads each custom type from the environment, which when run against multiple environments in the same ruby process is the exact situation generated types were created to prevent. I would argue that a separate ruby process is required to properly generate types without risk of ruby code bleeding between environments.

I also dislike the idea of requiring users to utilize `postrun` to generate types as this requires users to create a program/script that takes care of running `puppet generate types` against each of the modified environments. Even if r10k came bundled with such a script, `postrun` can only be specified once so if users wish to run their own program and use the bundled generate types script they must either create a wrapper script that calls both or modify their program to call the bundled script. In my opinion a baked-in type generation solution provides a better UX.

binford2k:
You're totally right.

justinstoller:
Pretty sure this PR broke r10k, is anyone using this code in their environments yet?

justinstoller:
I was under the impression that the supplying of environments to the post deploy hook was the way we were going to solve this problem?

lib/r10k/action/runner.rb
```diff
@@ -42,10 +42,23 @@ def setup_logging
...
+        overrides = {
+          cachedir: @opts[:cachedir],
+          deploy: {
+            puppet_path: @opts[:'puppet-path'],
+            generate_types: @opts[:'generate-types']
+          }
+        }
 
         with_overrides = config_settings.merge(overrides) do |key, oldval, newval|
+          break oldval if newval.nil?
```
justinstoller, line 54: I'm currently seeing all non-trivial r10k actions failing with `NoMethodError: undefined method 'keys' for "/var/cache/r10k":String`, I think this line is the culprit.

AFAICT, `break` will exit the merge loop here, causing `with_overrides` to be whatever the value of `oldval` is. eg. the "/var/cache/r10k" instead of the config_settings hash.

nabertrand: I've been using this code in my environments for months, but admittedly in the 2.6.x branch. I think you're right that this should have been a `next` instead of `break`. Does r10k work correctly for you if you make that change?

binford2k: @nabertrand tests were passing before I merged it. Think you could update tests to catch this problem too?

nabertrand: I confirmed that `break` is causing the problem @justinstoller is seeing and will add some tests to catch this in the future.

nabertrand: @binford2k #917 should fix this. Interestingly, this PR would have failed the spec tests if they had run more recently. It would be nice if the Travis CI/azure checks automatically ran again if the target branch changed.

PR #879, opened 2018-11-29
(maint) Manage locales and add an initial r10k.pot

justinstoller:


Magisus (approved):
Weird that we had this file in the gitignore. This seems in line with our other branches.

PR #886, opened 2019-01-11
(MAINT) Remove puppet-docker-tools

nwolfe:
In favor of direct Docker CLI calls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ notifications:
 sudo: false
 jdk:
   - oraclejdk8
-before_install: gem update bundler
+before_install: gem install bundler -v '< 2' --no-document
 matrix:
   include:
     - stage: r10k tests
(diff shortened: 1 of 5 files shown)


nwolfe:
No idea why Travis is failing on 2.1.0 but working on versions below and above that 🤷‍♂️ 

Those tests/failures have nothing to do with my changes, so....bitrot?

.travis.yml
```diff
@@ -9,7 +9,7 @@ notifications:
 sudo: false
 jdk:
   - oraclejdk8
-before_install: gem update bundler
+before_install: gem install bundler -v '< 2' --no-document
```
nwolfe, line 12: I tried doing what Josh did here: https://github.com/puppetlabs/puppet/blob/5.5.x/.travis.yml#L4-L9

But TravisCI kept having YAML syntax problems on it?  I even put it all on one line and still had trouble.

This seems to work.

justinstoller (approved):

PR #891, opened 2019-02-05
(RK-336) Update tests to point to artifactory instead of int-resources

Magisus:
Our internal test fixture servers changed, so this commit updates the
tests to use the new download location.


Magisus:
I ran these tests and they pass again now.

shrug (approved):

tvpartytonight (approved):

PR #893, opened 2019-02-07
(maint) Add container analytics and a few maintenance updates

underscorgan:


docker/r10k/Dockerfile
```diff
@@ -2,7 +2,13 @@ FROM ubuntu:16.04
 
 ARG vcs_ref
 ARG build_date
-ARG version="2.6.4"
+ARG version="3.1.0"
+# Used by entrypoint to submit metrics to Google Analytics.
+# Published images should use "production" for this build_arg.
+ARG pupperware_analytics_stream="dev"
+ENV PUPPERWARE_ANALYTICS_STREAM="$pupperware_analytics_stream"
+ENV PUPPERWARE_ANALYTICS_TRACKING_ID="UA-132486246-5"
```
nwolfe: I was putting as much as possible in the .sh script itself so it's all in one spot.

We couldn't do that with puppetserver since it has 2 images, so I generalized the .sh script to take an ENV variable that could be configured in both images.

It looks like that's the pattern you copied here, by putting the Tracking ID here as an ENV variable, even though that's not necessary with this image.

Was that intentional, and having the Tracking ID as an ENV variable is actually a better pattern?

Or should we go with the puppetdb pattern?

Either way tiny thing, just curious if you chose the puppetserver pattern over the puppetdb pattern for a reason.

This is awesome either way, glad you were able to follow the wiki page! 👍 ❤️

nwolfe (approved):

docker/r10k/Dockerfile
underscorgan: Oh, no, I think having the tracking ID in the script is better, will update that. I was going back and forth between puppetdb and puppetserver and I think I missed that one :)

Iristyle:
Because `master` didn't have the right `.travis.yml` changes at the time the PR went up, we didn't initially run the container build there.. but we do get a clean run in Azure.

Since we're green in Azure, we should be good on this. Merging

PR #903, opened 2019-02-26
(CPR-665) build r10k alpine image

underscorgan:
Build r10k alpine image including rugged gem. Does not include puppet or
puppet agent.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,4 +1,5 @@
 PUPPERWARE_ANALYTICS_STREAM ?= dev
+NAMESPACE ?= puppet
 git_describe = $(shell git describe)
 vcs_ref := $(shell git rev-parse HEAD)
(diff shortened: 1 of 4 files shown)


underscorgan:
When looking at a ticket that came in regarding installing the rugged gem for use with r10k, I started looking at converting r10k to use alpine. I don't think this container actually needs puppet/puppet-agent for anything, I think it was just using that as a means of not using system ruby? What do y'alls think?

nwolfe:
Awesome!

Seems like a good change.

Does this invalidate the need for #902 ?

underscorgan:
@nwolfe yep, this would invalidate #902

underscorgan:
I don't think there's more work here, mostly had this marked as a draft so we could make sure this seemed like a reasonable direction

PR #904, opened 2019-03-01
(MAINT) Push container README during publish

nwolfe:
Add a container README.md file and a step in the `make publish` target
to push the README up to DockerHub after pushing the image.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -48,10 +48,22 @@ test: prep
 	@bundle install --path .bundle/gems
 	@PUPPET_TEST_DOCKER_IMAGE=$(NAMESPACE)/r10k:$(version) bundle exec rspec r10k/spec
 
-publish: prep
+push-image: prep
 	@docker push $(NAMESPACE)/r10k:$(version)
 ifeq ($(IS_LATEST),true)
 	@docker push $(NAMESPACE)/r10k:latest
 endif
 
-.PHONY: lint build test publish
+push-readme:
+	@docker pull sheogorath/readme-to-dockerhub
+	@docker run --rm \
+		-v $(PWD)/README.md:/data/README.md \
+		-e DOCKERHUB_USERNAME=$(DISTELLI_DOCKER_USERNAME) \
+		-e DOCKERHUB_PASSWORD=$(DISTELLI_DOCKER_PW) \
+		-e DOCKERHUB_REPO_PREFIX=puppet \
+		-e DOCKERHUB_REPO_NAME=r10k \
+		sheogorath/readme-to-dockerhub
+
+publish: push-image push-readme
+
+.PHONY: prep lint build test publish push-image push-readme
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,7 @@
+# Configuration
+
+The following environment variables are supported:
+
+- `PUPPERWARE_DISABLE_ANALYTICS`
+
+  Set to `true` to disable Google Analytics metrics.


docker/Makefile
```diff
@@ -43,14 +43,22 @@ ifeq ($(IS_LATEST),true)
 	@docker tag puppet/r10k:$(version) puppet/r10k:latest
 endif
 
-test: prep
+test:
 	@bundle install --path .bundle/gems
 	@PUPPET_TEST_DOCKER_IMAGE=puppet/r10k:$(version) bundle exec rspec r10k/spec
 
-publish: prep
+publish:
```
nwolfe: I think the only thing that needs to depend on `prep` is the `build` target.

docker/Makefile
```diff
@@ -43,14 +43,22 @@ ifeq ($(IS_LATEST),true)
...
 	@docker push puppet/r10k:$(version)
 ifeq ($(IS_LATEST),true)
 	@docker push puppet/r10k:latest
 endif
+	@docker pull sheogorath/readme-to-dockerhub
+	@docker run --rm \
+		-v $(PWD)/README.md:/data/README.md \
+		-e DOCKERHUB_USERNAME=$(DISTELLI_DOCKER_USERNAME) \
+		-e DOCKERHUB_PASSWORD=$(DISTELLI_DOCKER_PW) \
```
nwolfe, line 62: I verified this image works on my own personal account + credentials, but haven't tested it yet with these DISTELLI_* yet.

docker/Makefile
```diff
@@ -43,14 +43,22 @@ ifeq ($(IS_LATEST),true)
...
+		-e DOCKERHUB_REPO_PREFIX=puppet \
+		-e DOCKERHUB_REPO_NAME=r10k \
+		sheogorath/readme-to-dockerhub
 
-.PHONY: lint build test publish
+.PHONY: prep lint build test publish
```
nwolfe: `prep` doesn't have a file associated with it so it should also be in `.PHONY`

docker/README.md
```diff
@@ -0,0 +1,7 @@
+# Configuration
+
+The following environment variables are supported:
+
+- `PUPPERWARE_DISABLE_ANALYTICS`
+
+  Set to `true` to disable Google Analytics metrics.
```
nwolfe, line 7: I'm not looking to properly document these containers right now. We can add a Trello card for that and circle back.

Just wanted to get the automation/plumbing in place, and start the README with the ENV configuration values.

Although, we already have some documentation up on DockerHub for [puppetdb](https://hub.docker.com/r/puppet/puppetdb) and [puppetserver](https://hub.docker.com/r/puppet/puppetserver), so I might include that text in the new README.md's for those repos.

docker/Makefile
nwolfe: Also I originally had this separated into two targets -- `push-image` and `push-readme` -- and then just had:
```makefile
push-image:
    ...

push-readme:
    ...

publish: push-image push-readme
```

I kinda like that separation but could go either way.

docker/Makefile
underscorgan: no, test and publish also depend on the version, so if you're running those on their own those depend on prep too.

docker/Makefile
underscorgan: Hm, would probably prefer to not have these tied to the `DISTELLI_*` environment variables, can we have other vars that we pass in in our distelli-manifest.yml?

I think I like having them separated into two targets

docker/Makefile
nwolfe: Oh yeah `publish` does need it for the version, but `test` doesn't since that's running the built image that's provided (and doesn't reference version anywhere).

I'll put them back I guess.

docker/Makefile
nwolfe: What's the concern with reusing these variables?

underscorgan: 🤷‍♀️ mostly just feels tied to the infrastructure we're using, though i guess other people won't be publishing

nwolfe: Makes sense.

Ok...let's consider this as "CI-only" for now, since `publish` already is.

If we're updating the README.md more frequently than pushing an image, and we find that we want to use this make target to update DockerHub instead of copy/pasting, then we can revisit credentials here and make this target easier for local execution.

PR #905, opened 2019-03-01
(MAINT) Disable metrics by default

nwolfe:
Make them opt-in instead of opt-out.

Also rename the variable so all the analytics-related variables follow
the naming pattern PUPPERWARE_ANALYTICS_*, since we also have
PUPPERWARE_ANALYTICS_STREAM (not user-facing though).

### DIFF
--- a/docker/r10k/docker-entrypoint.d/10-analytics.sh
+++ b/docker/r10k/docker-entrypoint.d/10-analytics.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-if [ "${PUPPERWARE_DISABLE_ANALYTICS}" = "true" ]; then
-    echo "($0) Pupperware analytics disabled; skipping metric submission"
+if [ "${PUPPERWARE_ANALYTICS_ENABLED}" != "true" ]; then
+    echo "($0) Pupperware analytics not enabled; skipping metric submission"
     exit 0
 fi
 
--- a/docker/README.md
+++ b/docker/README.md
@@ -2,6 +2,6 @@
 
 The following environment variables are supported:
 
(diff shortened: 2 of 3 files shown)


docker/r10k/docker-entrypoint.d/10-analytics.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-if [ "${PUPPERWARE_DISABLE_ANALYTICS}" = "true" ]; then
-    echo "($0) Pupperware analytics disabled; skipping metric submission"
+if [ "${PUPPERWARE_ANALYTICS_ENABLED}" != "true" ]; then
+    echo "($0) Pupperware analytics not enabled; skipping metric submission"
```
nwolfe, line 4: We _could_ add another message here saying how to enable metrics 🤷‍♂️ 

    echo "($0) Set PUPPERWARE_ANALYTICS_ENABLED=true to submit metrics"

stahnma:
Is there a README or any docs to update? (Either to show the option to opt in or the change in behavior?)

nwolfe:
> Is there a README or any docs to update? (Either to show the option to opt in or the change in behavior?)

Not yet, docs are still up in the PR here: https://github.com/puppetlabs/r10k/pull/904

That will need to be updated to reflect these changes, or vice-versa, whichever goes in first. 
Just getting the changes up for now.

docker/r10k/docker-entrypoint.d/10-analytics.sh
nwolfe: Nope, actually I'm going to remove all printouts from the container since this is a CLI, not a service.

These container messages being printed out on every run of the CLI is a poor experience.

PR #906, opened 2019-03-02
(maint) Refactor Docker specs for internally hosted Azure

Iristyle:
* Emit rspec version when running specs
* Output the gem env
* Rewrite Docker spec invocation so that stdout / stderr are emitted (for the sake of debugging). This also fixes a problem with running `%x` in environments other than Azure.
* Only emit Azure info when running in an Azure VM

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,10 +20,15 @@ steps:
     #
     # Azure
     #
-    Write-Host "`n`n$line`nAzure`n$line`n"
-    Invoke-RestMethod -Headers @{'Metadata'='true'} -URI http://169.254.169.254/metadata/instance?api-version=2017-12-01 -Method Get |
-      ConvertTo-Json -Depth 10 |
-      Write-Host
+    $assetTag = Get-WmiObject -class Win32_SystemEnclosure -namespace root\CIMV2 |
+      Select -ExpandProperty SMBIOSAssetTag
+    if ($assetTag -eq '7783-7084-3265-9085-8269-3286-77')
+    {
+      Write-Host "`n`n$line`nAzure`n$line`n"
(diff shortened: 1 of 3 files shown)


underscorgan (approved):
I trust you for the powershell, I still have a hard time reading that, but the rspec looks good

azure-pipelines.yml
```diff
@@ -20,10 +20,15 @@ steps:
     #
     # Azure
     #
-    Write-Host "`n`n$line`nAzure`n$line`n"
-    Invoke-RestMethod -Headers @{'Metadata'='true'} -URI http://169.254.169.254/metadata/instance?api-version=2017-12-01 -Method Get |
-      ConvertTo-Json -Depth 10 |
-      Write-Host
+    $assetTag = Get-WmiObject -class Win32_SystemEnclosure -namespace root\CIMV2 |
+      Select -ExpandProperty SMBIOSAssetTag
+    if ($assetTag -eq '7783-7084-3265-9085-8269-3286-77')
```
Iristyle, line 25: This is a magic BIOS value that indicates Azure... go figure!

nwolfe: Dear god.

Why not look for the existence of that `TF_BUILD` environment variable like we did in the `build.ps1`?

nwolfe: Yeah, can we not use this same pattern: https://github.com/puppetlabs/pe-console-services/blob/2019.1.x/docker/ci/build.ps1#L11-L14

Iristyle: Because that won't work. `TF_BUILD` is always set when you're running the VSTS agent for a build - it doesn't matter where the VM lives.

We need to know whether we're actually on an Azure VM - and this is the only documented way of verifying that (alternatively the metadata get is wrapped in a `try` / `catch`, but that could result in lengthy network timeouts as we've seen in Facter).

PR #910, opened 2019-03-09
(MAINT) Don't overwrite entrypoint script in tests

nwolfe:
Refactor the tests so they don't require overwriting the entrypoint
script.


Iristyle:
Much better!

I compared the spec runtime against the prior runs and there was no difference between keeping the container running all the time (like we did previously) and this improved mechanism. We also don't get the container id mixed in with the test output like we did in the past.

PR #917, opened 2019-04-09
Fix error merging config file settings

nabertrand:
This commit fixes an error handling config file settings introduced in #874. It also adds several tests to ensure settings are loaded and overridden as expected.

### DIFF
--- a/lib/r10k/action/runner.rb
+++ b/lib/r10k/action/runner.rb
@@ -42,23 +42,17 @@ module R10K
       def setup_settings
         config_settings = settings_from_config(@opts[:config])
 
-        overrides = {
-          cachedir: @opts[:cachedir],
-          deploy: {
-            puppet_path: @opts[:'puppet-path'],
-            generate_types: @opts[:'generate-types']
-          }
-        }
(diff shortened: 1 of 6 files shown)


lib/r10k/action/runner.rb
```diff
@@ -42,23 +42,17 @@ def setup_logging
...
         config_settings = settings_from_config(@opts[:config])
 
-        overrides = {
-          cachedir: @opts[:cachedir],
-          deploy: {
-            puppet_path: @opts[:'puppet-path'],
-            generate_types: @opts[:'generate-types']
-          }
-        }
+        overrides = if @opts[:'puppet-path'].nil? && @opts[:'generate-types'].nil?
```
justinstoller: Stylistically, we don't usually assign a value from a multiline if else.

Maybe either setting `overrides` in each branch, or something like:
```rb
overrides = {}
overrides[:deploy] = {} if @opts['puppet-path'] || @opts['generate-types']
```

nabertrand: Fixed

justinstoller (approved):

PR #918, opened 2019-04-16
(maint) Prep for 3.2.0 release

justinstoller:


justinstoller:
@binford2k we need to do an r10k release today, do you have any objections?

binford2k (approved):
Looks great to me! I was hoping to get #861 in the next release, but I don't trust it yet without some more reviews and testing.

PR #924, opened 2019-04-29
(RK-341) Pin back cri gem to maintain ruby 2.3 support

donoghuc:
This commit limits the cri dependency versions to `2.15.{1,2,3}` as `2.15.4` is incompatible with `ruby 2.3` which is the common system ruby interpreter version for test infrastructure (for example `osx10.{13,14}`)

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -23,7 +23,8 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored',   '1.2'
-  s.add_dependency 'cri',       '~> 2.15.1'
+  # cri 2.15.4 added dependency on ruby >= 2.4. Unpin when r10k drops ruby 2.3 support.
+  s.add_dependency 'cri',       '>= 2.15.1', '< 2.15.4'
 
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -1,5 +1,11 @@
 CHANGELOG
 =========
+3.0.4
+----
+
+## Changes
+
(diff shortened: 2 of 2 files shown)


donoghuc:
I am assuming this should be targeted against 3.0.x and then merged up to master? 
https://tickets.puppetlabs.com/browse/RK-341

r10k.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored',   '1.2'
-  s.add_dependency 'cri',       '~> 2.15.1'
+  s.add_dependency 'cri',       '>= 2.15.1', '< 2.15.4'
```
Magisus, line 27: Would you mind adding a comment here about why this was pinned? Ideally we should remove the pin eventually, once we also want to drop Ruby 2.3 support.

Magisus:
Yes, the 3.0.x target is correct. We're already pinning this gem to an even older version in 2.6.x, which is the oldest maintained branch.

donoghuc:
Was able to un-block bolt CI with https://github.com/puppetlabs/bolt/pull/971/commits/0fdc6ddd87071c68f17263de76844eee9c4e8c99 
is there a release planned for the 3.0.x series coming up soon?

binford2k (approved):

Magisus:
Can you also put a note in the Changelog about this, so we can cut the release directly after this is merged? Version 3.0.4.

justinstoller:
It looks like CRI released a new version 2.25.5 that fixes this and moves the required Ruby back to 2.3??  https://rubygems.org/gems/cri/versions/2.15.5  https://github.com/ddfreyne/cri/blob/master/NEWS.md#2155

donoghuc:
OH yeah, Sweet. https://github.com/ddfreyne/cri/issues/91

PR #927, opened 2019-05-10
Users should be able to include external Puppetfiles in their own.

mdwheele:
Related: #885 

Hello!

:tipping_hand_woman: This PR adds a new `incl` syntax to the `Puppetfile` spec for including an external Puppetfile as if it were part of the environments dependencies.

Sample Puppetfile:
```Puppetfile
forge 'http://forge.puppetlabs.com'

# Include shared modules.
incl '/usr/local/share/r10k/Puppetfile'

#
# Tenant-specific dependencies.
#
mod 'puppetlabs/stdlib'
mod 'puppetlabs-apache',
  :git => 'git@github.com:puppetlabs/puppetlabs-apache.git',
  :branch => 'master'
```

Sample External Puppetfile at `/usr/local/share/r10k/Puppetfile`:
```Puppetfile
#
# Shared dependencies.
#
mod 'puppetlabs/apt'
```

In our environment, we have multiple tenants separated into separate control repositories. Some of these tenants depend on shared modules (roles/profiles) we maintain as a community. This means that when we decide to update underlying component modules, we have to communicate those changes to individual tenants and it creates a release bottleneck.

One "hack" that we've been using in the last year was to have our r10k node write the "shared upstream `Puppetfile`" to disk and then in tenant `Puppetfile`, they do the following:

```
# "Include" shared Puppetfile
self.instance_eval(File.read('/usr/local/share/r10k/Puppetfile'))

# 
# Tenant-specific dependencies
#
mod 'camptocamp-kmod', '2.3.0'
mod 'crayfishx-firewalld', '3.4.0'
```

The fact that `Puppetfile` are `eval()-d` as Ruby currently allows us to solve the problem. I

In #885, an AST-based approach is proposed for hosts running Ruby 2.6+. If this were the standard behaviour, then our eval-based approach to sharing modules would break. My proposal is based on the implementation @binford2k provides in #885 and adds new syntax to officially support inclusion of an external `Puppetfile`.

**Implementation Details / Credits**

- I plucked the AST-based parser implementation from #885 and credited @binford2k as the author in e177d55. Very awesome work.
- Following, I implemented support for `incl`. The directive receives a single argument that can be a relative or absolute path. If a relative path is used, it is relative to the *calling* `Puppetfile`.
- I did not impose any limit on the depth of includes that can be used. Arguably, as a matter of best practice, one might say "limit to a single depth of includes". I'm amenable to enforcing this, but left it out for now.
- There is no cycle detection, but that seems like something someone would have to go out of their way to do here. This feature is intended to be a very simple include.
- Includes are expected to function as if the contents of the file were in-lined into the calling `Puppetfile`.

Please let me know if you have any questions. I will be adding additional unit tests to cover this code, but wanted to get this PR submitted. I appreciate your time and consideration!

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -298,7 +298,7 @@ class Puppetfile
 
             parse(File.read(path))
           else
-            # Should we log unexpected Ruby code?
+            raise LoadError, _("unrecognized declaration '%{method}'") % {method: args.shift}
           end
         end
         
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -274,23 +274,38 @@ class Puppetfile
               item.children.first
             end
           end.compact
-
+          
           case name
           when :mod
-            @librarian.add_module(args.shift, *args)
+            mod = args.shift
+            if args.empty? 
+              @librarian.add_module(mod, nil)
+            else
+              @librarian.add_module(mod, *args)
+            end            
           when :forge
             @librarian.set_forge(args.shift)
           when :moduledir
             @librarian.set_moduledir(args.shift)
+          when :incl
+            arg = args.shift
+            
+            if Pathname.new(arg).absolute?
+              path = arg
+            else
+              path = File.join(File.dirname(@librarian.puppetfile_path), arg)
+            end
+
+            parse(File.read(path))
           else
             # Should we log unexpected Ruby code?
           end
+        end
+        
+        node.children.each do |n|
+          next unless n.is_a? RubyVM::AbstractSyntaxTree::Node
 
-          node.children.each do |n|
-            next unless n.is_a? RubyVM::AbstractSyntaxTree::Node
-
-            traverse(n)
-          end
+          traverse(n)
         end
       rescue => e
         puts e.message
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -72,8 +72,19 @@ class Puppetfile
   end
 
   def load!
-    dsl = R10K::Puppetfile::DSL.new(self)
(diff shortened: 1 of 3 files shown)


binford2k:
I'm a little hesitant of this workflow. I think that if we wanted to use multiple Puppetfiles, I'd prefer a `Puppetfile.d/` directory for partials.

lib/r10k/puppetfile.rb
```diff
@@ -249,6 +344,16 @@ def moduledir(location)
...
     end
 
+    def incl(librarian)
+      if Pathname.new(librarian).absolute?
+        path = arg
+      else
+        path = File.join(File.dirname(@librarian.puppetfile_path), librarian)
+      end
+
+      instance_eval(File.read(path), @librarian.puppetfile_path)
```
binford2k, line 354: If we're trying to avoid eval'ing code directly then this line breaks that goal.

binford2k:
Hey, thanks for the idea @mdwheele! I'm going to close this now though, since we have a better way for you to include external `Puppetfiles` via your own policy. You'd want to write a short script that defined how to combine all the partials and then output a single YAML formatted list of modules to use in your environment. (You could even write your Puppetfiles in YAML, if you'd like)

See the [`exec` source type](https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#exec-environment-source) for more information on how that would work.

mdwheele:
> Hey, thanks for the idea @mdwheele! I'm going to close this now though, since we have a better way for you to include external `Puppetfiles` via your own policy. You'd want to write a short script that defined how to combine all the partials and then output a single YAML formatted list of modules to use in your environment. (You could even write your Puppetfiles in YAML, if you'd like)

No worries at all man! The `exec` source type looks like a much better (and safer) approach to this. We always knew that hack would eventually go away! 😀

PR #932, opened 2019-05-30
Ensure that 'git clone' does not use hardlinks

### LINKED ISSUE #931: Git clone should prevent usage of hardlinks on copy
Running `git clone` copies files with hardlinks and causes issues with SELinux `fixfiles` noting conflicting SELinux resources at application.

This can be remedied by passing `--no-hardlinks` to `git clone`. Unfortunately, this is not exposed to the `.gitconfig` file as far as I can tell so r10k needs to be updated to work around the issue.

trevor-vaughan:
Shell git uses hardlinks by default for local repo clones. This causes
issues with the r10k cache and SELinux contexts since the source and
destination will, quite often, be a home directory and some other
location on the system.

To prevent this (and other inadvertent issues with hardlinks in the r10k
cache), this patch adds the `--no-hardlinks` option to each call of `git
clone` when using shellgit.

Closes #931

### DIFF
--- a/lib/r10k/git/shellgit/bare_repository.rb
+++ b/lib/r10k/git/shellgit/bare_repository.rb
@@ -24,7 +24,7 @@ class R10K::Git::ShellGit::BareRepository < R10K::Git::ShellGit::BaseRepository
     proxy = R10K::Git.get_proxy_for_remote(remote)
 
     R10K::Git.with_proxy(proxy) do
-      git ['clone', '--mirror', remote, git_dir.to_s]
+      git ['clone', '--no-hardlinks', '--mirror', remote, git_dir.to_s]
     end
   end
 
--- a/lib/r10k/git/shellgit/working_repository.rb
+++ b/lib/r10k/git/shellgit/working_repository.rb
@@ -28,7 +28,7 @@ class R10K::Git::ShellGit::WorkingRepository < R10K::Git::ShellGit::BaseReposito
   #
   # @return [void]
   def clone(remote, opts = {})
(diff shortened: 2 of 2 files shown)


justinstoller:
Should this be configurable? It seems like a pretty meaningful change in default behavior. And one that many folks might not want. But I could be wrong?

trevor-vaughan:
@justinstoller The only drawback is that you may end up using additional disk space. That said, the current default behavior is actively unsafe as a cache.

Making it configurable wouldn't be a bad thing but making it non-configurable ensures that you have two actual independent copies that can't be harmed by someone going crazy with a script which is the current state of affairs.

Magisus:
Given that this does go against git's normal default behavior, I feel like it is pretty surprising to change globally for such a specific problem case. My instinct is to make it configurable and keep it off by default, unless it becomes clear people (not on SELinux) would prefer to avoid hardlinks generally.

justinstoller:
I think this is a great change. But because it's a change in the run time characteristics we should put it behind a flag, promote the use of it in 3.x, and switch the default in 4.

trevor-vaughan:
@justinstoller I'm fine with this.

I don't have time to figure out how to do that but 'allow edits from maintainers' is on so feel free to hack it up however you need to. Alternatively, push this through and then feature flag it. It's a tiny change in terms of code anyway.

adrienthebo:
@trevor-vaughan thanks for the contribution! I'd like to get this merged with a feature flag, is this something you're still interested in working on?

mwaggett:
Are we still thinking we might want to make this change for r10k 4.0.0?

Issue #930, opened 2019-05-30
nil access when not setting generate-types


ripienaar:
```
sh-4.2# /opt/puppetlabs/puppet/bin/r10k version
r10k 3.2.0
sh-4.2# /opt/puppetlabs/puppet/bin/r10k  deploy environment production -p -c r10k.yaml --trace

r10k: Runtime error: #<NoMethodError: undefined method `[]=' for nil:NilClass>
/opt/puppetlabs/puppet/lib/ruby/gems/2.4.0/gems/r10k-3.2.0/lib/r10k/action/runner.rb:49:in `setup_settins'
```

I set no `deploy` settings in my Puppetfile.

I dug a bit into this:

https://github.com/puppetlabs/r10k/blob/6dc614801f50102b719f2a9ff1112ef6541235ce/lib/r10k/action/runner.rb#L47-L49

Here you check `if @opts[:'puppet-path'] || @opts[:'generate-types']` and later check `unless @opts[:'generate-types'].nil?` - this fails when generate-types is set to `false`.

I am not sure what it is that sets generate-types to false, but if I add a debug `p @opts` there its false and so this code raises the exception.

ripienaar:
This is due to https://github.com/ddfreyne/cri/issues/94 which was released during the night, that sets forbidden to `false` and not `nil` anymore and so trips up this code (and quite a few other things, help handling too etc)

downgrading cri to 2.15.6 fixes it.

nmaludy:
This also bit me with the following:

```
$ bundle exec r10k puppetfile check

r10k: Runtime error: #<ArgumentError: R10K::Action::Puppetfile::Check cannot handle option 'help'>
```

aerickson:
I just encountered this issue also.

Locking the versions used seems like a good idea.

Pixelshifter:
Got this error as well when deploying a new git-branch as Puppet environment (r10k deploy environment $env -p). Locking the CRI gem version did indeed solve this for me (locked to 2.15.6).

aerickson:
@ripienaar This should be fixed in 3.3.0.

Pixelshifter:
@aerickson Yes, upgrading to 3.3.0 fixed it for me. Did you pin the CRI gem version?

PR #933, opened 2019-05-30
lock cri gem version

aerickson:
see https://github.com/puppetlabs/r10k/issues/930

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored',   '1.2'
-  s.add_dependency 'cri',       '~> 2.15.1'
+  s.add_dependency 'cri',       '2.15.6'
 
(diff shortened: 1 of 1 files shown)


aerickson:
Regarding the test failures, I believe they would fail on master if rerun (I can't seem to trigger a run).

The cri gem was updated last night and the last master build was before then.

justinstoller (approved):

justinstoller:
@Iristyle 	@underscorgan 	Do you understand why the docker tests are failing here? I don't see how this is related to the PR in question.

underscorgan:
so the conversation doesn't only exist in slack, this is failing since we're installing from gem not building from source. Have added a card to look into only running the tests when changes happen in the 'docker' directory.

aerickson:
@underscorgan Thanks for the clarification.

smortex:
The problem is fixed in [cri-2.15.8](https://github.com/ddfreyne/cri/pull/100).

PR #934, opened 2019-05-31
(RK-342) Update for behavior change in Cri options hash

donoghuc:
With the release of cri 2.15.7 flags are now always included in the options hash with the value set to `false`. This was causing errors in the `R10k::Util::Setops` setops method when executing puppetfile or deploy actions. This commit adds `help` to the `allowed_initialize_opts` in the `R10k::Action::Base` class

### DIFF
--- a/lib/r10k/action/base.rb
+++ b/lib/r10k/action/base.rb
@@ -24,6 +24,7 @@ module R10K
         {
           :config => true,
           :trace  => true,
+          :help => true,
         }
       end
     end


donoghuc:
For sure not a full solution. I think we will need to add some more (for example `no-force` would likely need to be added to; https://github.com/puppetlabs/r10k/blob/6dc614801f50102b719f2a9ff1112ef6541235ce/lib/r10k/action/deploy/display.rb#L83 ). I just put this up based on how i was hacking around it yesterday instead of trying to explain in in a comment on the ticket.

justinstoller:
@pcarlisle 	you said you were going to work with those involved on a longer term solution, is this part of that?

pcarlisle:
Unfortunately this is more complex and can't be fully solved on this version of cri. We're waiting on an upstream patch that's in progress which will give a more full solution to this. https://github.com/ddfreyne/cri/issues/94

PR #938, opened 2019-06-03
(maint) Don't run docker tests if there weren't any changes to the docker directory

underscorgan:


underscorgan:
made sure the build still triggers in https://github.com/puppetlabs/r10k/pull/939

nwolfe (approved):
I would've thought there was a built-in way to do a conditional build, like specifying a regex, but I don't see one!

underscorgan:
same. alas. If we find one can always make it better :)

PR #941, opened 2019-06-05
Ensure queue isn't empty before trying to get a new module

### LINKED ISSUE #940: pool_size is throwing errors
## Expected Behavior 
No errors while using `pool_size`.

## Actual Behavior

```
mbaur@puppet-development:~ $ r10k deploy environment -p production
ERROR	 -> Thread error during concurrent module deploy: queue empty
WARN	 -> Puppet Forge module 'puppet-staging' has been deprecated, visit https://forge.puppet.com/puppet/staging for more information.
ERROR	 -> Thread error during concurrent module deploy: queue empty
ERROR	 -> Thread error during concurrent module deploy: queue empty
ERROR	 -> Thread error during concurrent module deploy: queue empty
ERROR	 -> Thread error during concurrent module deploy: queue empty
```

## Steps to Reproduce the Problem

```
mbaur@puppet-development:~ $ cat r10k.yaml
---
:cachedir: /home/mbaur/.puppetlabs/opt/puppet/cache/r10k
:sources:
  control:
    basedir: /home/mbaur/.puppetlabs/etc/code/environments
    remote: https://github.com/example42/psick

:deploy:
  purge_whitelist:
  - resource_types

:pool_size: 5
mbaur@puppet-development:~ $ rm -rf /home/mbaur/.puppetlabs/opt/puppet/cache/r10k/* /home/mbaur/.puppetlabs/etc/code/environments/*
mbaur@puppet-development:~ $ r10k deploy environment -p production
ERROR	 -> Thread error during concurrent module deploy: queue empty
WARN	 -> Puppet Forge module 'puppet-staging' has been deprecated, visit https://forge.puppet.com/puppet/staging for more information.
ERROR	 -> Thread error during concurrent module deploy: queue empty
ERROR	 -> Thread error during concurrent module deploy: queue empty
ERROR	 -> Thread error during concurrent module deploy: queue empty
ERROR	 -> Thread error during concurrent module deploy: queue empty
```

## Specifications

Please add this info:

  1. Output of ```facter -p``` on the failing node (at least the OS related facts)
```
mbaur@puppet-development:~ $ facter -p os
{"name"=>"Ubuntu", "family"=>"Debian", "release"=>{"major"=>"16.04", "full"=>"16.04"}, "lsb"=>{"distcodename"=>"xenial", "distid"=>"Ubuntu", "distdescription"=>"Ubuntu 16.04.6 LTS", "distrelease"=>"16.04", "majdistrelease"=>"16.04"}}
```
  1. Version of Puppet and of the module
```
mbaur@puppet-development:~ $ puppet --version
6.4.2
```
Module Version see --> https://github.com/example42/psick/blob/production/Puppetfile

  1. The relevant Puppet code and eventually Hiera data

See `Steps to Reproduce the Problem`

  1. Ruby version

```
mbaur@puppet-development:~ $ ruby --version
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
```

Same for 2.5.3

baurmatt:
Fixes #940.

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -193,12 +193,9 @@ class Puppetfile
   def visitor_thread(visitor, mods_queue)
     Thread.new do
       begin
-        until mods_queue.empty?
-          mod = mods_queue.pop(true)
-          mod.accept(visitor)
-        end
+        while mod = mods_queue.pop(true) do mod.accept(visitor) end
       rescue ThreadError => e
-        logger.error _("Thread error during concurrent module deploy: %{message}") % {message: e.message}
+        logger.debug _("Module thread %{id} exiting: %{message}") % {message: e.message, id: Thread.current.object_id}
         Thread.exit
       end
     end
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -193,7 +193,10 @@ class Puppetfile
   def visitor_thread(visitor, mods_queue)
     Thread.new do
       begin
-        while mod = mods_queue.pop(true) do mod.accept(visitor) end
+        until mods_queue.empty?
+          mod = mods_queue.pop(true)
+          mod.accept(visitor)
+        end
       rescue ThreadError => e
         logger.error _("Thread error during concurrent module deploy: %{message}") % {message: e.message}
         Thread.exit
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -197,6 +197,9 @@ class Puppetfile
       rescue ThreadError => e
         logger.debug _("Module thread %{id} exiting: %{message}") % {message: e.message, id: Thread.current.object_id}
         Thread.exit
+      rescue => e
+        logger.error _("Error during concurrent deploy of module %{name}: %{message}") % {name: mod.name, message: e.message}
+        Thread.exit
       end
     end
   end
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -179,7 +179,21 @@ class Puppetfile
     logger.debug _("Updating modules with %{pool_size} threads") % {pool_size: pool_size}
     mods_queue = modules_queue(visitor)
     thread_pool = pool_size.times.map { visitor_thread(visitor, mods_queue) }
-    thread_pool.each(&:join)
+    thread_exception = nil
+
+    # If any threads raise an exception the deployment is considered a failure.
+    # In that event clear the queue, wait for other threads to finish their
+    # current work, then re-raise the first exception caught.
+    begin
+      thread_pool.each(&:join)
+    rescue => e
+      logger.error _("Error during concurrent deploy of a module: %{message}") % {message: e.message}
+      mods_queue.clear
+      thread_exception ||= e
+      retry
+    ensure
+      raise thread_exception unless thread_exception.nil?
+    end
   end
 
   def modules_queue(visitor)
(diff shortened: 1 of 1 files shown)


baurmatt:
Hey @oesgalha and @reidmv. As you originally implemented this option (Thanks for this btw! :)), could you please take a look if this makes sense? Thanks! :)

lib/r10k/puppetfile.rb
```diff
@@ -195,7 +195,7 @@ def visitor_thread(visitor, mods_queue)
       begin
         while mod = mods_queue.pop(true) do mod.accept(visitor) end
       rescue ThreadError => e
-        logger.error _("Thread error during concurrent module deploy: %{message}") % {message: e.message}
+        logger.debug _("Module thread %{id} exiting: %{message}") % {message: e.message, id: Thread.current.object_id}
```
Sharpie, line 212: After re-reading the docs a bit, it looks like `ThreadError` is raised by `.pop(true)` when the queue is empty. Logging this at DEBUG level seems reasonable. However, we should add an additional rescue clause at the end of the block to catch and log errors that occur during `mod.accept`:

```ruby
rescue => e
  logger.error _("Error during concurrent deploy of module %{mod_name}: %{message}") %
    {mode_name: mod.name,
     message: e.message}
  Thread.exit
end
```

Sharpie (changes requested):

lib/r10k/puppetfile.rb
reidmv: Added additional rescue block

Sharpie (approved):
Looks good 👍

reidmv:
@Sharpie actually, upon further thought I'm not sure catching non-ThreadErrors here is a good idea... we should at least re-raise. Seems like there may be a danger of threads all terminating before finishing the queue of modules and (if we're catching like this here) r10k happily continuing without having processed all modules.

Seems like it'd be better to let the exception float up further.

Thoughts?

oesgalha:
I think the behavior should be similar to the serial install.

If an error during `mod.accept` interrupts r10k in a serial install, then concurrent install should be interrupted too.

If this is the desired behavior, see [abort_on_exception](https://ruby-doc.org/core-2.6.3/Thread.html#method-i-abort_on_exception-3D) and this relevant doc:

> However, it's important to note that an exception that occurs in any thread except the main thread depends on abort_on_exception. This option is false by default, meaning that any unhandled exception will cause the thread to terminate silently when waited on by either join or value. You can change this default by either abort_on_exception= true or setting $DEBUG to true.

https://ruby-doc.org/core-2.6.3/Thread.html#class-Thread-label-Exception+handling

I think that re-raising with `Thread.main.raise(e)` should work too. An alternative to `abort_on_exception`.

Sharpie:
Yeah, if we want to interrupt the main thread, then a simple re-raise won't cut it if we're generating and raising exceptions inside the worker threads:

```
$ cat test.rb
#!/bin/env ruby

Thread.new do
  raise 'boom!'
end

puts 'you should never see this'

raise 'lets try that again'

puts 'you definitely should never see this'

$ ruby test.rb
you should never see this
test.rb:9:in `<main>': lets try that again (RuntimeError)
```

Sharpie:
Replacing `Thread.exit` with `Thread.main.raise(e)` the final `rescue => e` block seems to me like it would accomplish the goal of halting r10k.

One thing I'm not sure about is what would happen if the other threads are terminated abruptly while in the middle of pulling a `.git` update. I would hope that libgit2 is smart enough to handle this, but as they say: hope is not a strategy.

Sharpie:
@reidmv The approach of storing the exception and waiting for the threads to finish looks great!

baurmatt:
Hey @reidmv / @Sharpie, could you trigger a 3.3.1 release which includes this fix? Would love to use it in production! :)

Magisus:
I've just released this as 3.3.1. Enjoy!

PR #942, opened 2019-06-21
(maint) Build container from source

underscorgan:
Rather than installing the latest gem from rubygems, do a `gem build`
and install that gem so we're running from source.

### DIFF
--- a/docker/r10k/Dockerfile
+++ b/docker/r10k/Dockerfile
@@ -1,3 +1,11 @@
+FROM alpine:3.9 as build
+
+RUN apk add --no-cache ruby git
+RUN mkdir /workspace
+WORKDIR /workspace
+COPY . /workspace
+RUN gem build r10k.gemspec
+
 FROM alpine:3.9
 
 ARG vcs_ref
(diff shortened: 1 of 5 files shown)


underscorgan:
Running into a couple of things here now.
1. ~`COPY .` doesn't work on windows~
2. ~The way the r10k gemspec is written unless we manually copy in _all_ files from the source directory the gemspec will fail.~

nope, it was me missing the updates to the powershell build script that caused issues.

nwolfe:
@underscorgan Interesting. We've been using `COPY .` in the puppetdb image for a few months now: https://github.com/puppetlabs/puppetdb/blob/master/docker/puppetdb/Dockerfile#L10

We use a `.dockerignore` file in that case though...I wonder if there's something in that file that prevents us from hitting whatever bug you're running into 🤷‍♂

underscorgan:
Oh interesting. I'll investigate this more, thanks @nwolfe !

docker/r10k/Dockerfile
```diff
@@ -1,3 +1,11 @@
+FROM alpine:3.9 as build
+
+RUN apk add --no-cache ruby git
+RUN mkdir /workspace
+WORKDIR /workspace
+COPY . /workspace
```
Iristyle, line 6: Based on the Azure fail:

```
 ---> 7827868f48a6
Step 6/23 : RUN gem build r10k.gemspec
 ---> Running in c58a63ca78f4
ERROR:  Gemspec file not found: r10k.gemspec‌
##[error]docker : The command '/bin/sh -c gem build r10k.gemspec' returned a non-zero code: 1
```

I think `cwd` in Azure is not quite the same, so `.` might be a different set of files? Could verify with

```Dockerfile
RUN ls -l /workspace
```

underscorgan: I think this is due to a change in the shared build.ps1, testing

PR #943, opened 2019-07-01
Implemented default_branch_override flag to override Puppetfile :default_branch entry during environment deployments

chrisleicester:
### Description
With this change, a user can specify a default branch override using the `--default-branch-override` flag. This value will be passed through to the deployment and used in place of the Puppetfile's `:default_branch` entry. If no `:default_branch` was previously specified in the Puppetfile, the supplied default branch override will be used anyways.

### How this feature is to be used
When running r10k, specify a default_branch_override using the `--default-branch-override` flag. 
> `r10k deploy environment -p production --default-branch-override temporary_branch_237292`
 
### How important is this feature to our environment?
The impetus behind this feature is to address a bug in **Continuous Delivery for Puppet Enterprise**, where deployments utilizing a combination of the Control Branch AND [Temporary Branch pattern](https://puppet.com/docs/continuous-delivery/2.x/deployment_policies.html#concept-5197) for module deployments would cause modules to be deployed to the wrong branch, or fail completely.

(Because the control_branch pattern requires that both the control repo and module repo have branches with matching names, and the temporary branch pattern creates a brand new branch by definition, there is no way for the control branch pattern to succeed, _unless_ a default_branch is specified in the module's Puppetfile specifying same name as the new temporary branch). The only way to be sure the `default_branch` entry in the Puppetfile matches the temporary branch name is to supply is as an override.

Note: This bug also causes Impact Analysis runs to fail, as they use the Temporary Branch pattern as well.

### How useful this feature would be for other users?
I can't imagine another use case that makes overriding the default_branch via r10k simpler than just updating your Puppetfile, but its always possible.

### Potential caveats?
When a `default_branch_override` is specified for a deployment, the `default_branch` param will be set on _every module_ in the Puppetfile during deployment. Generally speaking, this will have no effect on the deployments of those modules, as the desired ref will generally always exist (ie. modules pulled from the forge), **and** the name of the `default_branch_override` will likely not exist in that repo.

### DIFF
--- a/lib/r10k/cli/deploy.rb
+++ b/lib/r10k/cli/deploy.rb
@@ -53,6 +53,7 @@ scheduled. On subsequent deployments, Puppetfile deployment will default to off.
           DESCRIPTION
 
           flag :p, :puppetfile, 'Deploy modules from a puppetfile'
+          required nil, :'default-branch-override', 'Specify a branchname to override the default branch in the puppetfile'
 
           runner R10K::Action::CriRunner.wrap(R10K::Action::Deploy::Environment)
         end
--- a/lib/r10k/module/git.rb
+++ b/lib/r10k/module/git.rb
@@ -23,6 +23,11 @@ class R10K::Module::Git < R10K::Module::Base
   #   @return [String]
   attr_reader :desired_ref
 
+  # @!attribute [r] default_ref
+  #   @api private
+  #   @return [String]
+  attr_reader :default_ref
+
   def initialize(title, dirname, args, environment=nil)
     super
 
--- a/doc/dynamic-environments/usage.mkd
+++ b/doc/dynamic-environments/usage.mkd
@@ -60,6 +60,18 @@ Update a single environment and force an update of modules:
 This will update the given environment and update all contained modules. This is
 useful if you want to make sure that a given environment is fully up to date.
 
+- - -
+
+Update a single environment and specify a default branch override:
+
+    r10k deploy environment my_working_environment --puppetfile --default-branch-override default_branch_override
+
+This will update the given environment and update all contained modules, overrideing 
+the :default_branch entry in the Puppetfile of each module. This is used primarily to allow
+automated r10k solutions using the control_branch pattern with a temporary branch deployment to 
+ensure the deployment is pushed to the correct module repository branch. Note that the :default_branch
+is only ever utilized if the desired ref cannot be located.
+
 ### Deploying modules
 
 Update a single module across all environments:
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -109,7 +109,7 @@ module R10K
         end
 
         def visit_puppetfile(puppetfile)
-          puppetfile.load
+          puppetfile.load(@opts[:'default-branch-override'])
 
           yield
 
@@ -146,7 +146,7 @@ module R10K
         end
 
         def allowed_initialize_opts
-          super.merge(puppetfile: :self, cachedir: :self, :'no-force' => :self)
+          super.merge(puppetfile: :self, cachedir: :self, :'no-force' => :self, :'default-branch-override' => :self)
         end
       end
     end
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -55,17 +55,20 @@ class Puppetfile
     @loaded = false
   end
 
-  def load
+  def load(default_branch_override = nil)
     if File.readable? @puppetfile_path
-      self.load!
+      self.load!(default_branch_override)
     else
       logger.debug _("Puppetfile %{path} missing or unreadable") % {path: @puppetfile_path.inspect}
     end
   end
 
-  def load!
+  def load!(default_branch_override = nil)
+    @default_branch_override = default_branch_override
+
     dsl = R10K::Puppetfile::DSL.new(self)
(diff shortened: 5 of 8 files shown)


reidmv:
UX comment: per [puppet-nogui](https://github.com/puppetlabs/puppet-nogui/blob/master/patterns/api_style_guide.md#word-separators), it is preferred that Puppet (Inc.) CLI flags use "-" word separators, rather than "_".

To align with nogui, the CLI flag should be `--default-branch-override`.

I'd also suggest dropping the one-character version of this flag. This should be an uncommon invocation almost always performed by an automaton. Allowing only the long-form gives two benefits then, by A) reserving the human-friendly short `-d` for possible future use cases, while B) making it much clearer what's going on when the flag does need to be invoked.

justinstoller:
Code-wise this looks great, Chris.

I agree with Reid and Maggie regarding the use of dashes instead of underscores in the actual UI (It also brings it inline with `--no-force`, `--puppet-path`, and `--generate-types`). Once that's address, I'd be happy to merge.

lib/r10k/cli/deploy.rb
```diff
@@ -53,7 +53,7 @@ def self.command
           DESCRIPTION
 
           flag :p, :puppetfile, 'Deploy modules from a puppetfile'
-          required :d, :default_branch_override, 'Specify a branchname to override the default branch in the puppetfile'
+          required :default-branch-override, 'Specify a branchname to override the default branch in the puppetfile'
```
reidmv: Should this be:
```
required nil, :'default-branch-override', 'Specify ...'
```
Just guessing based on other examples in the file that maybe the `nil` is missing here, and for symbols with dashes this'll require quotes, yes?

What's the difference between `required` and `flag`? Guessing it's non-intuitive but as long as I'm commenting thought I'd ask. :)

chrisleicester:
Thanks for the feedback, everyone.

@justinstoller I implemented the requested changes. What is the process for getting this merged in + the changes merged up into the other supported r10k release versions?

lib/r10k/cli/deploy.rb
chrisleicester: Good catch! Thank you.

As for the difference between `flag` and `required`:

`flag`: indicates a true/false value
`required` implies that a _parameter_ is required (which is counter-intuitive, and apparently has [since been deprecated in the library we use](https://github.com/ddfreyne/cri#command-line-options)).

justinstoller:
One nit that I just noticed: The git commit is not in our group's standard format. Aren't super particular about this with community contributors but for internal contributions it'd be nice to conform to this https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#making-changes which is the basis for most of our projects in PE/Platform's git hygiene.
eg.
```
CDPE-1653: modified default branch override CLI flag to be '--default-branch-override' instead of '-d' or '--default_branch_override' in order to match existing CLI flag conventions in r10k
```
would be something like:
```
(CDPE-1653) Conform branch override to cli conventions

Per nogui[1], use dashes instead of underscores in cli flags and do
not use single character flag for rarely used options that humans are
not expected to interact with.

1. https://github.com/puppetlabs/puppet-nogui/blob/master/patterns/api_style_guide.md#word-separators
```

Once merged, we open PRs for merging up to master and/or maintenance branches. I'm happy to open the PRs for that. What versions are you hoping to release this from (2.6.x goes to Irving, 3.0.x to Johnson, 3.2.x to Kearney, master is Lovejoy)? We have Jenkins jobs that do the release process for us. All those branches should auto promote into PE.

chrisleicester:
@justinstoller Thanks for the heads up. I hadn't seen the `making changes` part of the contributing doc.

I can update each commit message individually, or rebase/squash them all into a single commit. Either option is fine with me, just let me know your preference.

As for the releases, we need the changes to be in each release you referenced (Irving, Johnson, Kearney, Lovejoy).

justinstoller:
I don't have a strong preference, but I could see them being one logical - not tooo large - commit, and that's probably easiest.

chrisleicester:
@justinstoller the commits have been squashed and I believe the feature is ready to be merged whenever you are.

lib/r10k/module/git.rb
```diff
@@ -23,6 +23,11 @@ def self.implement?(name, args)
   #   @return [String]
   attr_reader :desired_ref
 
+  # @!attribute [r] default_ref
+  #   @api private
+  #   @return [String]
+  attr_reader :default_ref
```
Magisus, line 29: Where does this get set?

Magisus: Oh I see it. Nevermind. We just were never retrieving it form the outside before, and now we are in the test. Sometimes I hate how Ruby doesn't make you declare instance variables anywhere centralized...

Magisus (approved):

Magisus:
Looks like the commit message still refers to the short-hand `-d` flag, but I don't think we allow that anymore. Can you update that?

chrisleicester:
@Magisus Awesome ok. Updated. thank you!

chrisleicester:
@justinstoller @Magisus It looks like this was merged into 2.6.x, but none of the other branches we discussed. Is there anything I can do to help facilitate and/or test this?

Magisus:
We'll get the mergeups rolling today. I don't think you should need to do anything else.

chrisleicester:
@Magisus awesome thank you. I'll follow up re: code-manager once the r10k merge-ups are complete (https://github.com/puppetlabs/code-manager/pull/413)

PR #949, opened 2019-07-10
Merge 3.0.x -> 3.2.x

Magisus:


Magisus:
There is a [PR](https://github.com/puppetlabs/r10k/pull/942/files#diff-11ee4a521ee318acb249451574624884R7) up to fix the container test failure on master, I'm asking if those tests should be fixed on this branch too or just turned off. Regardless, I think we can ignore that failure for this mergeup.

tvpartytonight (approved):
👍

PR #952, opened 2019-07-11
Add R10K::CLI tests

nabertrand:
These tests verify that each `Cri` command and option behaves as expected and sets options and args correctly within the `Action` instances. Running the tests also showed that several of the `deploy` options were only valid in certain subcommands, so those options were moved into subcommands as appropriate. The other option would be to add support for those options in each of the `deploy` subcommands.

### DIFF
--- a/lib/r10k/cli/deploy.rb
+++ b/lib/r10k/cli/deploy.rb
@@ -21,17 +21,6 @@ module R10K::CLI
 (https://puppetlabs.com/blog/git-workflow-and-puppet-environments/).
         DESCRIPTION
 
-        required nil, :cachedir, 'Specify a cachedir, overriding the value in config'
-        flag nil, :'no-force', 'Prevent the overwriting of local module modifications'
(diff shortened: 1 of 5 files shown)


nabertrand:
There seems to be a transient Travis build failure with jruby. Closing and re-opening to trigger rebuild.

bastelfreak (approved):

bastelfreak:
hey people, can we get this merged/reviewed?

adrienthebo (approved):
LGTM! @nabertrand thanks for your patience!

adrienthebo:
@dhollinger @justinstoller I've verified this locally, can one of you check out this PR as well?

justinstoller (approved):

Magisus:
This appears to be causing failures when used with code-manger. One thing it does not appear to take into account is https://github.com/puppetlabs/r10k/pull/1020, which we are using with a new feature to allow code-manager to do a deploy of a single module.

PR #954, opened 2019-07-15
(CDPE-1813) Add module deploy info to .r10k-deploy.json file

chrisleicester:
With this change, module information will be added to the output file '.r10k-deploy.json' for external consumption (including the module name, version, and git sha associated with the commit used if pulling the module from a git repository.


chrisleicester:
opened against wrong base branch in r10k.

PR #955, opened 2019-07-15
(CDPE-1813) Add module deploy info to .r10k-deploy.json file

chrisleicester:
With this change, module information will be added to the output file '.r10k-deploy.json' for external consumption (including the module name, version, and git sha associated with the commit used if pulling the module from a git repository.

In particular, we need the module deploy sha in order to detect changes to the module when running Impact Analysis. Currently, the only way to detect changes is it compare the module versions.

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -125,11 +125,22 @@ module R10K
         end
 
         def write_environment_info!(environment, started_at, success)
+          module_deploys = []
+          if environment.puppetfile && environment.puppetfile.modules
+            environment.puppetfile.modules.each do |mod|
+              name = mod.name
+              version = mod.version
+              sha = mod.repo.head rescue nil
+              module_deploys.push({:name => name, :version => version, :sha => sha})
+            end
+          end
+
           File.open("#{environment.path}/.r10k-deploy.json", 'w') do |f|
             deploy_info = environment.info.merge({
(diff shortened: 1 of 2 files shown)


binford2k (changes requested):
Can you add some tests for this?

chrisleicester:
@binford2k due to the way the existing r10k unit tests are written, we never _actually_ create the .r10k-deploy.json file during testing. Therefore, there is currently no way to validate if the file is being written correctly (this fact existed prior to me modifying the code). That being said, I have added integration testing to the [corresponding code-manager code changes](https://github.com/puppetlabs/code-manager/pull/418/files#diff-64617a27cc613499d1a1c3cf5619a098R215) that does a full deploy, then validates the contents of the `.r10k-deploy.json` file created by r10k.

Given that r10k can (and is) used independently of code-manager, this is not an ideal solution, but it does verify that the r10k functionality works E2E.

If this still isn't satisfactory, we can always expose the `write_environment_info` method as a public method, and unit test it that way. Currently its private. 

If you have any other suggestions on how to validate this code, please let me know!

chrisleicester:
@binford2k tests added. Please review when you have a chance.

binford2k (approved):

Magisus (approved):
Seems straightforward and useful.

Magisus:
I'll merge this when I'm done with tagging shenanigans in the pe-r10k packaging project.

PR #960, opened 2019-07-22
(RK-340) Prefer SHA256 checksum for verifying modules

Magisus:
FIPS mode does not allow md5 to be used, to this updates r10k to always prefer using SHA256 when verifying modules. If the Forge does not have SHA256 data and FIPS mode is not enabled, it will fall back to verifying with md5. If FIPS mode is enabled, it will error.

### DIFF
--- a/lib/r10k/forge/module_release.rb
+++ b/lib/r10k/forge/module_release.rb
@@ -111,6 +111,7 @@ module R10K
       # module release checksum given by the Puppet Forge. On mismatch, remove
       # the cached copy.
       #
+      # @raise [R10K::Error] when no SHA256 is available and FIPS mode is on
       # @return [void]
       def verify
         logger.debug1 "Verifying that #{@tarball_cache_path} matches checksum"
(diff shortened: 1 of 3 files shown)


Magisus:
I have manually tested this on a Redhat7 machine with FIPS enabled.
Before:
```
[root@wbv035rc23t0t9g ~]# /opt/puppetlabs/puppet/bin/r10k deploy environment production -p --verbose
INFO	 -> Using Puppetfile '/etc/puppetlabs/code/environments/production/Puppetfile'
INFO	 -> Deploying environment /etc/puppetlabs/code/environments/production
INFO	 -> Environment production is now at 8b156e5005757d380269b9d805bd88350411d7cf
INFO	 -> Deploying Puppetfile content /etc/puppetlabs/code/environments/production/modules/stdlib
md5_dgst.c(82): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
Aborted
```
After:
```
[root@afm5zw4g3j3o4er ~]# /opt/puppetlabs/puppet/bin/r10k deploy environment production --verbose -p
INFO	 -> Using Puppetfile '/etc/puppetlabs/code/environments/production/Puppetfile'
INFO	 -> Deploying environment /etc/puppetlabs/code/environments/production
INFO	 -> Environment production is now at 799c083cac01c58dfbf72bee52b8386a1bbcd98a
INFO	 -> Deploying Puppetfile content /etc/puppetlabs/code/environments/production/modules/apache
```

lib/r10k/forge/module_release.rb
```diff
@@ -108,12 +115,71 @@ def download
       def verify
         logger.debug1 "Verifying that #{@tarball_cache_path} matches checksum"
 
-        md5_of_tarball = Digest::MD5.hexdigest(File.read(@tarball_cache_path, mode: 'rb'))
+        contents = File.read(@tarball_cache_path, mode: 'rb')
+        sha256_of_tarball = Digest::SHA256.hexdigest(contents)
```
joshcooper: One optimization is to call `Digest::SHA256.file(@tarball_cache_path).hexdigest` so that you don't have to read the entire file into memory. It should handle binary mode on Windows.

Magisus: Thanks I'll give that a shot

Magisus:
Updated with Josh's suggestion

lib/r10k/forge/module_release.rb
```diff
@@ -108,12 +115,70 @@ def download
...
 
-        md5_of_tarball = Digest::MD5.hexdigest(File.read(@tarball_cache_path, mode: 'rb'))
+        sha256_of_tarball = Digest::SHA256.file(@tarball_cache_path).hexdigest
+
+        if @sha256_file_path.exist?
+          verify_from_sha256_file(sha256_of_tarball)
+        else
+          begin
+            verify_sha256_from_forge(sha256_of_tarball)
+          rescue R10K::Error
```
justinstoller: This is just using an exception for control flow and should be done differently.

lib/r10k/forge/module_release.rb
```diff
@@ -108,12 +115,70 @@ def download
...
+            if R10K::Util::Platform.fips?
+              raise R10K::Error, "Could not verify module, no SHA256 checksum available, and MD5 checksums not allowed in FIPS mode"
+            end
+
+            logger.debug1 "No SHA256 checksum available, falling back to MD5"
+
+            md5_of_tarball = Digest::MD5.file(@tarball_cache_path).hexdigest
+
+            if @md5_file_path.exist?
+              verify_from_md5_file(md5_of_tarball)
```
justinstoller: If we parameterized these methods could we condense them? It seems that they all do largely the same thing but are reading different instance variables (and logging a formulaic, but slightly different message).

lib/r10k/forge/module_release.rb
```diff
@@ -205,6 +270,13 @@ def cleanup_md5_file_path
           md5_file_path.delete
         end
       end
+
+      # Remove the module release sha256.
+      def cleanup_sha256_file_path
```
justinstoller: This convention is wild.

lib/r10k/util/platform.rb
```diff
@@ -14,6 +16,16 @@ def self.platform
         end
       end
 
+      # We currently only suport FIPS mode on redhat 7, where it is
+      # toggled via a file.
+      def self.fips?
+        if File.exist?(FIPS_FILE)
+          File.read(FIPS_FILE).chomp == "1"
```
justinstoller, line 23: I'm kinda surprised the r10k user would be able to read this file

justinstoller: I'll be damned, it's world readable:
```
root@l3ieludaqizuhkz ~]# ls -la /proc/sys/crypto/
total 0
dr-xr-xr-x 1 root root 0 Jul 29 20:30 .
dr-xr-xr-x 1 root root 0 Jul 29 20:28 ..
-r--r--r-- 1 root root 0 Jul 29 20:30 fips_enabled
```

Magisus: Maybe it can't? Is the r10k user usually different than the puppet user? Because that's how puppet is doing it... see https://github.com/puppetlabs/puppet/blob/ba4d1a1aba0095d3c70b98fea5c67434a4876a61/lib/puppet/util/platform.rb#L24-L26. (This also worked in my manual testing, but there I was running as root on a pooler machine, which I guess might not be typical?)

lib/r10k/forge/module_release.rb
Magisus: Yeah I was thinking about that. I'll look into it.

lib/r10k/util/platform.rb
Magisus: Works for me :P

justinstoller:
I don't love the approach taken with this ...but it is consistent with the rest of the file, so... idk, I'd like to see some changes that I'd consider general programming best practices around code repetition and exception handling, but I'm open to the argument that it should go in largely as-is for consistency's sake.

Magisus:
I'm happy to do the refactor. I also don't like the amount of duplication involved, I just kind of wanted to see how the fall-back logic would work before doing a refactor, because I wasn't sure how different it was going to make the two cases. But I think I can come up with a decent abstraction with better control flow that will make this a lot less ugly.

lib/r10k/forge/module_release.rb
Magisus: What do you mean, specifically?

justinstoller: the abstraction over a call to `#delete`, and an existence check that most callers will already know the answer to.

not your code, but i dislike the conventions around the `cleanup_*_path` methods, and prefer methods that are _not_ part of an imperative, high level, public api to only act on their explicit inputs.

Magisus: I can ... cleanup... that too ;)

lib/r10k/forge/module_release.rb
```diff
@@ -104,54 +111,79 @@ def download
...
-      def verify_from_md5_file(md5_of_tarball)
-        md5_from_file = File.read(@md5_file_path).strip
-        if md5_of_tarball != md5_from_file
-          logger.error "MD5 of #{@tarball_cache_path} (#{md5_of_tarball}) does not match checksum #{md5_from_file} in #{@md5_file_path}. Removing both files."
-          cleanup_cached_tarball_path
-          cleanup_md5_file_path
+      def verify_from_file(tarball_checksum, checksum_file_path)
+        checksum_from_file = File.read(checksum_file_path).strip
+        if tarball_checksum != checksum_from_file
+          logger.error "Checksum of #{@tarball_cache_path} (#{tarball_checksum}) does not match checksum #{checksum_from_file} in #{checksum_file_path}. Removing both files."
```
Magisus, line 160: Do you think it's worth passing in the checksum type here to maintain putting `MD5` vs `SHA256` in the error message?

Magisus:
Refactored, had one question, let me know what you think.

lib/r10k/forge/module_release.rb
justinstoller: I don't think so, since a) I can't think of a use case that would involve scripting against checksum mismatches and b) the type of the checksum should still be inferable from the path (eg. `puppetlabs-concat-6.0.0.md5` or however that path works out). So I'd say the risk of breaking folks is very low and there's no loss in data for the end user.

But I'd be curious what the community thinks. @binford2k do you know if folks would be upset if we changed the checksum mismatch error message?

justinstoller (approved):

lib/r10k/forge/module_release.rb
binford2k: I've not ever seen anyone write tooling that depends on this. And I don't see a way to do it without changing the message, so I see no blocker here.

PR #962, opened 2019-07-24
Merge 2.6.x -> 3.0.x

Magisus:


Magisus:
@chrisleicester this is the mergeup of your extra deploy info changes, feel free to merge if it looks right.

lib/r10k/action/deploy/environment.rb
```diff
@@ -125,11 +125,22 @@ def visit_module(mod)
         end
 
         def write_environment_info!(environment, started_at, success)
+          module_deploys = []
+          if environment.puppetfile && environment.puppetfile.modules
```
chrisleicester: @Magisus I just realized something while reviewing this:

I believe the `if environment.puppetfile && environment.puppetfile.modules` check is completely unecessary, and could actually cause issues (The intention here is to check to see if the `puppetfile` of `puppetfile.modules` attributes are null, but when testing I ran across the error `environment.puppetfile function is not defined`, actually causing a crash. This was, however when generating mock data to validate data.

Given that each environment has a puppetfile function that returns the puppetfile object, and each puppetfile object has a function that returns its modules, it seems that while this check is unnecessary, it won't actually cause any issues down the line.

I think the proper approach would be to wrap the logic in `write_environment_info!` in a begin/rescue, but I question the value of that give how much overhead is involved. What's your take on it?

Magisus: Looking at those classes, I agree, it's extremely unlikely those things will ever be nil (and this is one of the reasons Ruby's whole mocking paradigm always makes me uncomfortable). I think you're right, given that, it would be clearer to take out. If you put up a PR against 2.6.x, I'll merge it and include it in this mergeup. Good catch!

chrisleicester: awesome! The PR is ready for your review here: https://github.com/puppetlabs/r10k/pull/963

Magisus:
Updated this with the `environment_write` change, @chrisleicester merge if you're happy with it now :)

chrisleicester:
@Magisus thanks! merged

PR #966, opened 2019-07-27
(maint) Consistently compare values to nil

smortex:
`overrides[:deploy]` is initialized as an empty Hash if any of _puppet-path_ and _generate-types_ is set.

`overrides[:deploy][:puppet_path]` and `overrides[:deploy][:generate_types]` are respectively set when _puppet-path_ and _generate-types_ are non-nil.

`false` being not nil, when _generate-types_ is set to `false`, `overrides[:deploy]` will not be initialized and an excpetion will be raised when setting `overrides[:deploy][:generate_types]` to false.

Always compare values to nil to avoid such a situation.

### DIFF
--- a/lib/r10k/action/runner.rb
+++ b/lib/r10k/action/runner.rb
@@ -44,7 +44,7 @@ module R10K
 
         overrides = {}
         overrides[:cachedir] = @opts[:cachedir] unless @opts[:cachedir].nil?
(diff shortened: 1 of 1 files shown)


smortex:
FWIW, the Azure Pipelines failures seems unrelated.

Dan33l (approved):
LGTM

PR #969, opened 2019-08-01
(RK340) Require puppet_forge 2.3.0 for sha256 checksum support

justinstoller:


Magisus (approved):

Magisus:
This is good, but I don't think it's sufficient to fix the test failures we're seeing. I didn't realize that they [autogenerate methods](https://github.com/puppetlabs/forge-ruby/blob/master/lib/puppet_forge/v3/base.rb#L23-L30) based on JSON fields, which means we need to be checking `respond_to?` instead of (in addition to) `nil?` when doing the fallback. Since we stub module downloading in the tests, none of them have SHA256 file fields. https://github.com/puppetlabs/r10k/blob/master/integration/tests/user_scenario/basic_workflow/single_env_switch_forge_git_module.rb#L23-L32

That is to say, oddly enough, we don't actually depend on the changes in that gem anywhere in our module downloading, since we implement our own verification.

justinstoller:
Gotcha, this updates us to pull in a version of puppet_forge that also supports sha256 verification (even though we implement the verification part ourselves in r10k).

Then I would assume the issues we're seeing internally come from our forge fixture not being updated with the sha256 checksum information that Jesse has put on the forge?

Magisus:
That's what I'm thinking, I'll check in with the Forge team. In the meantime, I'll fix the fallback logic.

PR #970, opened 2019-08-01
(RK-340) Check that the forge_release object has SHA256 file

Magisus:
Previously, we were only verifying that the Forge Release object did not
return nil from its `file_sha256` method. But actually, this method is
not even defined if the module does not provide a SHA256 checksum. So this
commit updates the verify logic to also check that the `file_sha256`
method exists.


Magisus:
This mirrors what the `puppet_forge` gem itself does when verifying, see https://github.com/puppetlabs/forge-ruby/blob/master/lib/puppet_forge/v3/release.rb#L48

tvpartytonight (approved):

PR #975, opened 2019-08-06
(maint) Minor Azure Yaml consistency updates

Iristyle:


azure-pipelines.yml
```diff
@@ -16,6 +16,8 @@ pool:
 variables:
   NAMESPACE: puppet
   CONTAINER_NAME: r10k
+  CONTAINER_BUILD_PATH: .
+  LINT_IGNORES: DL3008 DL3018 DL4000 DL4001 DL3028
```
nwolfe, line 20: You trying to move all codes out of `Lint-Dockerfile` ? Seems like a good thing

nwolfe (approved):

azure-pipelines.yml
Iristyle: Yeah, exactly... decided we should move that to the top as data and not make any assumptions in `Lint-Dockerfile` as defaults.

PR #978, opened 2019-08-20
(MAINT) Standardize docker test

nwolfe:
Use docker-compose so we can use our shared helper methods, and apply
other consistency updates.

----

Depends on https://github.com/puppetlabs/pupperware/pull/114


docker/Gemfile
```diff
@@ -4,5 +4,5 @@ gem 'rspec'
 gem 'rspec_junit_formatter'
 gem 'pupperware',
     :git => 'https://github.com/puppetlabs/pupperware.git',
-    :branch => 'master',
+    :branch => 'fix-stderr-support',
```
underscorgan, line 7: This should go back to master before merging

underscorgan (commented):
We should revisit this if we ever end up with more complicated entrypoints, but this seems reasonable

nwolfe:
Going to spend a few minutes seeing if we can't tease apart the entrypoint and command so we don't have to overwrite entrypoint in the tests.

nwolfe:
Nevermind, not putting r10k on compose.

PR #980, opened 2019-08-27
Make Digest subclass lookup thread-save

### LINKED ISSUE #979: Bug: Digest::Base cannot be directly inherited in Ruby
## Describe the Bug
r10k sometimes throws an `Digest::Base cannot be directly inherited in Ruby` error. This leads to module not being deployed and therefore other errors (failing CI pipelines in our case).

This showed up after we enabled the `pool_size` feature. (Currently using `pool_size: 10`)

## Expected Behavior
r10k doesn't fail and successfully deploys all modules.

## Steps to Reproduce
Currently I can't reliably reproduce this. We had 6 (out of probably thousands) jobs failing with this error within ~24h.

## Environment
 - Ruby: 2.5.3
 - r10k Version: 3.3.1
 - Platform: Debian 9.8 running in a docker container onto of a OpenStack VM.

baurmatt:
Fixes #979.

### DIFF
--- a/lib/r10k/forge/module_release.rb
+++ b/lib/r10k/forge/module_release.rb
@@ -116,7 +116,7 @@ module R10K
       def verify
         logger.debug1 "Verifying that #{@tarball_cache_path} matches checksum"
 
-        sha256_of_tarball = Digest::SHA256.file(@tarball_cache_path).hexdigest
+        sha256_of_tarball = Digest(:SHA256).file(@tarball_cache_path).hexdigest
 
         if @sha256_file_path.exist?
           verify_from_file(sha256_of_tarball, @sha256_file_path)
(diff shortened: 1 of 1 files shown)


Magisus (approved):
Good catch!

Magisus:
From notes I found, this will be effective in Ruby >= 2.2. At this point r10k requires 2.3, so sounds like we're good.

baurmatt:
Wow that was fast! Thanks a lot @Magisus! :) Any chance we can get this bug fix in a new Rubygem release? ;)

Magisus:
This is actually causing a problem in our FIPS-compliant builds, so I want to see if I can get that sorted before I release this. But hopefully soon!

baurmatt:
Oh, if I can help fixing the problem, let me know! :)

Magisus:
Finally got the bottom of our issues, and it turns out they were unrelated to this PR! I'll get this released.

PR #983, opened 2019-09-27
Add Yaml environment source

reidmv:
Feature demonstration: https://youtu.be/P0lljPd_9L8

Two major ideas here:

1. Make it so that a YAML object enumerating environments and their versions/sources can be consumed, rather than consulting Git branches in a repository for this information
2. Make it so that for environments defined this way, a list of modules can also be supplied _directly_. No need to involve a Puppetfile.

Example environments.yaml file:

```yaml
env_with_only_modules:
  type: bare
  modules:
    puppetlabs-stdlib: 6.0.0
    puppetlabs-concat: 6.1.0
    puppetlabs-exec: 0.5.0
    puppetlabs-ntp: 8.1.0
    reidmv-xampl:
      git: https://github.com/reidmv/reidmv-xampl.git
      ref: 62d07f2
    puppetlabs-pe_xl:
      git: https://github.com/puppetlabs/puppetlabs-pe_xl.git
      ref: bcbd7b9

env_with_control_repo_base:
  type: git
  remote: file:///Users/reidmv/Workspace/repos/control-repo/.git
  ref: dddbd8d
  modules:
    puppetlabs-pe_xl:
      git: file:///Users/reidmv/Workspace/repos/puppetlabs-pe_xl/.git
      ref: bcbd7b9
    reidmv-xampl:
      git: file:///Users/reidmv/Workspace/repos/reidmv-xampl/.git
      ref: 62d07f2
```

### DIFF
--- a/lib/r10k/source/hash.rb
+++ b/lib/r10k/source/hash.rb
@@ -1,0 +1,158 @@
+# This class implements an environment source based on recieving a hash of
+# environments
+#
+# @since 3.4.0
+#
+# DESCRIPTION
+#
+# This class implements environments defined by a hash having the following
+# schema:
+#
+#     ---
+#     type: object
+#     additionalProperties:
+#       type: object
+#       properties:
+#         type:
+#           type: string
+#         basedir:
+#           type: string
+#         modules:
+#           type: object
+#           additionalProperties:
+#             type: object
+#         moduledir:
+#           type: string
+#       additionalProperties:
+#         type: string
+#
+# The top-level keys in the hash are environment names. Keys in individual
+# environments should be the same as those which would be given to define a
+# single source in r10k.yaml. Additionally, the "modules" key (and moduledir)
+# can be used to designate module content for the environment, independent of
+# the base source parameters.
+#
+# Example:
+#
+#     ---
+#     production:
+#       type: git
+#       remote: 'https://github.com/reidmv/control-repo.git'
+#       ref: '1.0.0'
+#       modules:
+#         geoffwilliams-r_profile: '1.1.0'
+#         geoffwilliams-r_role: '2.0.0'
+#
+#     development:
+#       type: git
+#       remote: 'https://github.com/reidmv/control-repo.git'
+#       ref: 'master'
+#       modules:
+#         geoffwilliams-r_profile: '1.1.0'
+#         geoffwilliams-r_role: '2.0.0'
+#
+# USAGE
+#
+# The following is an example implementation class showing how to use the
+# R10K::Source::Hash abstract base class. Assume an r10k.yaml file such as:
+#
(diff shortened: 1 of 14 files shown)


lib/r10k/source/hash.rb
```diff
@@ -0,0 +1,158 @@
...
+#       additionalProperties:
+#         type: string
+#
+# The top-level keys in the hash are environment names. Keys in individual
+# environments should be the same as those which would be given to define a
+# single source in r10k.yaml. Additionally, the "modules" key (and moduledir)
+# can be used to designate module content for the environment, independent of
+# the base source parameters.
+#
+# Example:
```
reidmv, line 35: Start reading here ☝️ for a summary of how it would work

binford2k:
@reidmv Where would said `environments.yaml` file live on disk? How would it be version controlled, or is that an exercise left to the reader?

Also, I'm a fan of the general idea.

dhollinger (approved):
LGTM

reidmv:
> Where would said `environments.yaml` file live on disk? How would it be version controlled, or is that an exercise left to the reader?

@binford2k the `environments.yaml` file would live on disk on the master. Management and version control of the file is left as an exercise to the reader.

For shops that like keeping the content in Git, it could be as simple as a one-file Git repo with a hook to read the latest file on commit.

For shops that prefer to be API-driven, _either_ a script could query the API for a hash and write the file as part of a deploy action, _or_ r10k itself could read a hash from the API directly; see the [R10k::Source::Hash](https://github.com/puppetlabs/r10k/pull/983/files#diff-806801e1a469fa7fb24d8bff0f5aaa18) abstract class for suggestions on how that would be implemented in a later iteration.

This is a plumbing feature that user interfaces (such as PE's Code Manager) would build on.

binford2k (approved):

ekohl:
Makes me wonder if you want `environments.d/*.yaml` so people can share an environment. Branching would be as simple as a `cp`. Also less chance of merge conflicts.

binford2k:
@ekohl Reid's working on an `environments.d/` source now and will submit a separate PR

justinstoller:
Why doesn't this work for SVN?

reidmv:
> Why doesn't this work for SVN?

As this is currently an experimental feature, SVN support is untested. Focus has been on testing with Git sources.

Technically, I think it _should_ partly work for SVN sources. What will likely work is listing an SVN source as an environment in a YAML file. What won't work is specifying individual modules to deploy to that SVN source, as the SVN source type has not been converted from using the Base environment class to using the WithModules environment class. See [here](https://github.com/puppetlabs/r10k/pull/983/files#diff-4532eeb4585e014fdae4f028078fae6eL9-R9).

PR #993, opened 2019-10-31
Add support for defining custom puppet.conf when generating types

treydock:


justinstoller (dismissed):
The request makes sense. I'm a little unclear on the option handling - and I'm unclear if that's my own misunderstanding of what's currently required (we've had regressions in upstream CRI behavior and so have had to do wonky things in the past - or an issue in the code.

However, it copies the functionality of `--puppet-path`, which imo is what we want. So, assuming this works correctly for the submitter, I'm good with it.

adrienthebo:
@treydock thanks for submitting this pull request! This pull request has some outstanding merge conflicts, are you still interested in getting this PR merged?

treydock:
I resolved conflicts

lib/r10k/cli/deploy.rb
```diff
@@ -60,6 +60,13 @@ def self.command
               exit 1
             end
           end
+          option nil, :'puppet-conf', 'Path to puppet.conf', argument: :required do |value, cmd|
+            unless File.exist? value
```
mwaggett: Should we check readability instead of existence?

spec/unit/settings_spec.rb
```diff
@@ -129,6 +129,18 @@
...
     end
+
+    describe 'puppet_conf' do
+      it 'when file raises no error' do
+        expect(File).to receive(:exist?).with('/nonexistent').and_return(true)
+        expect { subject.evaluate('puppet_conf' => '/nonexistent') }.not_to raise_error
+      end
+
+      it 'when not file raises error' do
+        expect(File).to receive(:exist?).with('/nonexistent')
```
mwaggett: Would it be worth explicitly returning `false` here?

lib/r10k/cli/deploy.rb
treydock: Switched the `puppet-conf` checks for `exist?` to `readable?`.

spec/unit/settings_spec.rb
treydock: This was fixed as part of changing to readable? check.

mwaggett (dismissed):
Looks good to me, @treydock ! We've tracked down the test failure and it's being fixed in another PR, so once that goes in, we'll retest this and then we should be good to go! Thanks for your contribution!

mwaggett:
hey @treydock , sorry for all the churn, but would you mind resolving conflicts again and rebasing off master? We recently merged changes that fix breakages we were seeing. Thanks so much!

treydock:
@mwaggett I've rebased

mwaggett (dismissed):
Looks great!

lib/r10k/cli/deploy.rb
```diff
@@ -31,6 +31,13 @@ def self.command
             exit 1
           end
         end
+        option nil, :'puppet-conf', 'Path to puppet.conf', argument: :required do |value, cmd|
+          unless File.readable? value
+            $stderr.puts "The specified puppet.conf #{value} could not be read."
```
adrienthebo: We can use the `logger` method as it's defined on commands; the following should work:

```suggestion
            logger.fatal "The specified puppet.conf #{value} could not be read."
```

(We should also apply the same change for `puppet-path`, but that's best tracked in another PR.)

lib/r10k/cli/deploy.rb
```diff
@@ -31,6 +31,13 @@ def self.command
             exit 1
           end
         end
+        option nil, :'puppet-conf', 'Path to puppet.conf', argument: :required do |value, cmd|
+          unless File.readable? value
```
adrienthebo: We have validation also defined on the setting itself, so this isn't strictly necessary. If we drop this validation, we get the following:

```
r10k deploy -c r10k.yaml --puppet-conf not_a_file environment
ERROR    -> Validation failed for 'global' settings group:
    deploy:
        Validation failed for 'deploy' settings group:
            puppet_conf:
                The specified puppet.conf not_a_file is not readable
```

Is the above error sufficiently clear or are we better served by this more specific error messaging?

adrienthebo (dismissed):
I've got a couple of suggestions, but no blockers. @treydock if you see value in making the suggested changes please do; either way we can merge this in the next couple of days. Thanks for the contribution!

lib/r10k/cli/deploy.rb
treydock: I'm not sure how to make the error more specific. The check is if the file is readable and the error states it's not readable. Not much else to check or tell the user about.

lib/r10k/cli/deploy.rb
treydock: Made this change.

lib/r10k/cli/deploy.rb
mwaggett: @treydock I think what @adrienthebo is saying is that we can drop this `readable?` check altogether because there is validation elsewhere that will produce a similar error message.

mwaggett:
@treydock Would you mind adding an entry to the CHANGELOG for this? Also looks like we need another rebase/merge-conflict-resolution. 😓

lib/r10k/cli/deploy.rb
treydock: I removed the validation from this particular location.

treydock:
@mwaggett I added a changelog entry and rebased against master.

mwaggett (dismissed):

dhollinger (dismissed):
Love it!

mwaggett (approved):

mwaggett:
This was already approved by @dhollinger but the review got dismissed when I force-pushed to rebase and resolve conflicts in the changelog. so I'm merging even though there is only one recorded review.

PR #996, opened 2019-11-21
(maint) Run r10k as the `puppet` user

underscorgan:
This way permissions won't get weird when r10k and puppetserver are
mounting a shared code volume

### DIFF
--- a/docker/r10k/Dockerfile
+++ b/docker/r10k/Dockerfile
@@ -31,6 +31,11 @@ LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
       org.label-schema.schema-version="1.0" \
       org.label-schema.dockerfile="/Dockerfile"
 
+# Add a puppet user to run r10k as for consistency with puppetserver
+COPY docker/r10k/adduser.sh /
+RUN chmod a+x /adduser.sh && \
+    /adduser.sh
+
 RUN apk add --no-cache ruby openssh-client git ruby-rugged curl ruby-dev make gcc musl-dev
 COPY --from=build /workspace/r10k.gem /
 RUN gem install --no-doc r10k.gem json etc && \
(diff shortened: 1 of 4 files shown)


nwolfe (approved):

docker/r10k/Dockerfile
```diff
@@ -31,6 +31,10 @@ LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
       org.label-schema.schema-version="1.0" \
       org.label-schema.dockerfile="/Dockerfile"
 
+# Add a puppet user to run r10k as for consistency with puppetserver
+# The alpine image uses the 'ping' group with gid 999, so use that instead of
+# a 'puppet' group
+RUN adduser puppet -D -G ping -u 999
```
Iristyle: I have no idea if Alpine has any guarantees around group id to name mappings staying consistent across alpine versions?

Since we really care about the numeric value in the end, should we just use numeric values everywhere?

docker/spec/fixtures/Puppetfile
```diff
@@ -1,2 +1,2 @@
-moduledir 'test/modules'
+moduledir '/tmp/modules'
```
glennsarti, line 1: Not sure if this is a concern but, that's gone from a xplat path to a non-windows path. Given R10K is xplat, should this be changed?

docker/r10k/Dockerfile
underscorgan: ah, good call. we should probably do something safer here where we see if there's a `999` group (apparently that wasn't in 3.3 for example), add it if it doesn't exist, and add the user to  whatever group 999 is if it does exist

docker/spec/fixtures/Puppetfile
underscorgan: @glennsarti good catch. I don't think this is a concern since this Puppetfile is installed on the container, which will always be linux-y

docker/r10k/Dockerfile
underscorgan: @Iristyle updated!

docker/r10k/Dockerfile
```diff
@@ -39,7 +44,10 @@ RUN gem install --no-doc r10k.gem json etc && \
 COPY docker/r10k/docker-entrypoint.sh /
 RUN chmod +x /docker-entrypoint.sh
 COPY docker/r10k/docker-entrypoint.d /docker-entrypoint.d
+RUN chown -R puppet: /docker-entrypoint.d /docker-entrypoint.sh
```
Iristyle, line 47: Ah cool, never knew about the default group trick with `:<missing>`

PR #997, opened 2019-11-26
(maint) Updates for container versioning

underscorgan:
Containers built from source will now be tagged with `edge`, and
containers tagged with a version number will have the published gem
installed.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -6,13 +6,21 @@ build_date := $(shell date -u +%FT%T)
 hadolint_available := $(shell hadolint --help > /dev/null 2>&1; echo $$?)
 hadolint_command := hadolint --ignore DL3008 --ignore DL3018 --ignore DL3028 --ignore DL4000 --ignore DL4001
 hadolint_container := hadolint/hadolint:latest
-pwd := $(shell pwd)
-export BUNDLE_PATH = $(pwd)/.bundle/gems
-export BUNDLE_BIN = $(pwd)/.bundle/bin
-export GEMFILE = $(pwd)/Gemfile
+export BUNDLE_PATH = $(PWD)/.bundle/gems
+export BUNDLE_BIN = $(PWD)/.bundle/bin
+export GEMFILE = $(PWD)/Gemfile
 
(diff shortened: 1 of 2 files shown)


docker/Makefile
```diff
@@ -6,13 +6,21 @@ build_date := $(shell date -u +%FT%T)
...
 
-version := $(shell command grep VERSION ../lib/r10k/version.rb | awk '{print $$3}' | sed "s/'//g")
-dockerfile := Dockerfile
+ifeq ($(IS_RELEASE),true)
+	VERSION ?= $(shell echo $(git_describe) | sed 's/-.*//')
+	LATEST_VERSION ?= latest
+	dockerfile := Dockerfile-release
+	dockerfile_context := r10k
+else
+	VERSION ?= edge
```
Iristyle, line 19: Should be tagging this way in the Windows builds as well?

underscorgan: I could go either way. I hadn't worried about it on windows since we never publish releases from there but could update it for more consistency.

Iristyle: Let's get this merged and add a card about the consistency for later (FWIW, I think it makes it easier to compare results between CI systems when things are built / tagged the same way).

Iristyle:
Travis:

```
Successfully built a6acdfeea08f
Successfully tagged puppet/r10k:edge
```

Azure:

```
Successfully built 6764b09c767f
Successfully tagged puppet/r10k:3.3.3
```

Ignoring the inconsistency for now and merging.

docker/Makefile
nwolfe: Regarding consistency...

I still think we should explore moving this code into Ruby, so it's not duplicated between Bash and PowerShell.

At first our Docker invocations were trivial enough the duplication was fine, but at this point we have enough logic that surrounds those calls that it seems worthwhile to at least explore what that'd look like.

Iristyle: Yeah, I'm still not in favor of rewriting the PS1 / Makefile bits in Ruby and adding another dependency for builds.  Using Ruby is fine for specs, but adding it to builds adds more complexity as it's still necessary to bootstrap the build with the PS1 / Makefile anyway.

PR #1002, opened 2019-12-13
Add yamldir environment source type

reidmv:
This is an alternative UX to the yaml environment source type, but functionally similar. If it's proven to be a better UX we may abandon the yaml environment source type in favor of yamldir.

Credit to @ekohl for the idea!


ekohl (approved):
Nice!

PR #1006, opened 2020-01-02
Run spec tests on Ruby 2.6 as well

bastelfreak:
As more and more distributions support Ruby 2.6 is makes sense to also
run our test on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ before_install: gem install bundler -v '< 2' --no-document
 matrix:
   include:
     - stage: r10k tests
+      rvm: 2.6.5
+    - stage: r10k tests
       rvm: 2.5.0
     - stage: r10k tests
       rvm: 2.4.0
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -12,6 +12,8 @@ jdk:
 before_install: gem install bundler -v '< 2' --no-document
 matrix:
   include:
+    - stage: r10k tests
+      rvm: 2.6.0
```
mwaggett: I think it makes more sense to test on the latest Z in a given series. Looks like that would be [2.6.5](https://www.ruby-lang.org/en/downloads/releases/).

bastelfreak: oh my bad, I thought travis would automatically switch to the latest Z release. But that's only the case for Major.Minor syntax. I will update the PR.

bastelfreak: okay, `2.6` only brings us `2.6.3`. I will explicitly set `2.6.5`

mwaggett (approved):

PR #1007, opened 2020-01-07
(RK-355) Port to the maintained colored2 gem

binford2k:
See https://github.com/defunkt/colored/pull/11 for commentary.

### DIFF
--- a/bin/r10k
+++ b/bin/r10k
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 
 require 'r10k/cli'
-require 'colored'
+require 'colored2'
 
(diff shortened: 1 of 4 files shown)


mwaggett (approved):

Magisus:
This will require an update to the vanagon project before it can be promoted to PE. https://github.com/puppetlabs/pe-r10k-vanagon/blob/master/configs/components/rubygem-colored.rb

Magisus:
Doing the update to pe-r10k breaks with the current version of r10k, so I'm going to merge this and let it promote.

PR #1008, opened 2020-01-07
(maint) Update changelog for 3.4.0 release

Magisus:


CHANGELOG.mkd
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+3.4.0
+-----
+
+  - Switch to supported `colored2` gem
+  - Add YAML environment source [#983](https://github.com/puppetlabs/r10k/pull/983)
```
mwaggett: might just be me and my unfamiliarity with r10k, but I don't really know what this means..

Magisus: It's a huge addition, which is why I just went with linking the PR instead of trying to explain in the changelog... But I can try if you think it's worth it.

mwaggett: yeeeah.. I guess this sort of goes back to the to-changelog-or-not-to-changelog discussion..

imo, if there's a changelog, I should be able to get a brief understanding of the changes without having to look at PRs or commit messages. If the changelog is linking to PRs or mimicking commit messages, I don't see the point in having it at all.

because I'm new to r10k, I don't know if "environment source" is common terminology that folks will be familiar with or whether anyone looking at this will need to look at the PR to actually understand (which I personally don't love having to do). but I won't block on this.

mwaggett (approved):

CHANGELOG.mkd
Magisus: Fair. Maybe @reidmv would like to suggest a good way of synthesizing this?

reidmv: > Add new environment source types YAML and YAMLdir. These source types use configuration files to create environments, rather than VCS branches, and support deploying modules to an environment without modifying a control-repo.

How's this?

CHANGELOG.mkd
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+3.4.0
+-----
+
+  - Switch to supported `colored2` gem
+  - Add YAML environment source types [#983](https://github.com/puppetlabs/r10k/pull/983), [#1002](https://github.com/puppetlabs/r10k/pull/1002). YAML source types use configuration files to define environments, rather than VCS branches, and support deploying modules to an environment without modifying a control-repo. YAML environment source types are experimental in this release
```
mwaggett, line 8: Thank you @reidmv ! This is much clearer to me. 😊

mwaggett (approved):

PR #1011, opened 2020-01-08
Pin gettext as they have decided to make a breaking change

cyberious:
- gettext 3.3.0 has made a breaking change of suddenly requiring ruby >= 2.5.0
this breaks anybody on ruby 2.4.9 with Puppet 5 series supported versions


cyberious:
Also 2.6.0 might have been pushed with 5.5 series we are using 3.0 for control build testing and pulling all modules. Also given that R10k 3.3.0 has a ruby requirement of >=2.3.0 this would break semvar for you to allow gettext 3.3.0 which is requiring a greater version of Ruby than you currently have required.

Magisus:
We are planning to pin this in gettext-setup directly, which I think might mean we don't need to do it explicitly here. See https://github.com/puppetlabs/gettext-setup-gem/pull/75.

cyberious:
Thanks @Magisus didn't realize gettext_setup was `puppetlabs`

PR #1012, opened 2020-01-08
(maint) Update gettext-setup gem to 1.0

beechtom:
The gettext-setup gem now pins gettext to < 3.3.0. The latest release
of gettext-setup is 1.0, so is not captured by the current version
constraint for the gem.


mwaggett (approved):

mwaggett (changes requested):
looks like there's a transitive dependency that conflicts with this

Magisus:
We've retagged this update as 0.32, closing this.

PR #1013, opened 2020-01-20
Add Ruby 2.7 to testmatrix

bastelfreak:
Ruby 2.7 got released some time ago and the first distributions ship it,
so I think it makes sense to also execute the tests on Ruby 2.7.


justinstoller (approved):
Awesome. Thank you, Tim.

PR #1016, opened 2020-01-23
(RE-13192) Migrate saturn -> Artifactory

mwaggett:


Magisus (approved):
No luck with the defaults then, I take it?

mwaggett:
the issue is that certain version info is needed to construct these paths and, since I'm pretty new to beaker world, I wasn't sure what info would be available during different processes and I didn't want to change the api for any methods that other projects use. ¯\_(ツ)_/¯

PR #1019, opened 2020-01-27
Allow `>= 2.15.10` versions of cri

bastelfreak:
cri 2.15.10 is the latest version and people already use it with r10k so
I think we should also test against it / allow it. It claims it's
compatible with Ruby 2.3.

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored2',   '3.1.2'
-  s.add_dependency 'cri',       '2.15.6'
+  s.add_dependency 'cri', ['>= 2.15.10', '< 3.0.0']
 
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'


r10k.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored2',   '3.1.2'
-  s.add_dependency 'cri',       '2.15.6'
+  s.add_dependency 'cri',       '>= 2.15.6'
```
mwaggett: We might want to do `'~> 2.15'` or  `['>= 2.15.6', '< 3.0.0']` if we're worried about maintaining support for certain Ruby versions (major version bumps have bitten us in the past).

bastelfreak: updated it :)

mwaggett (approved):

pcarlisle:
This was pinned because newer versions were broken, so if we move forward it at least needs to set the minimum to after that issue was fixed, which i think is 2.15.8 based on a quick skim of the release notes.

mwaggett:
ooh, good to know. In that case, maybe we should just set the latest as our new minimum?

pcarlisle:
sure, if it works then that's fine

bastelfreak:
@pcarlisle @mwaggett updated it as requested.

mwaggett (approved):

PR #1020, opened 2020-01-31
(RK-360) Allow overriding the cachedir in `deploy module`

Magisus:
Previously, we added the ability to configure the cachedir from the CLI
for the `deploy environment` command, since Code Manager configures a
separate cachedir for each of its r10k workers. Now that we are also
using the `deploy module` command in Code Manager, it needs to respect
these special cachedirs the same way. The `--cachedir` flag can now be
used with the `deploy module` command.

### DIFF
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -72,7 +72,11 @@ module R10K
         end
 
         def allowed_initialize_opts
-          super.merge(environment: true, 'no-force': :self, 'generate-types': :self, 'puppet-path': :self)
+          super.merge(environment: true,
+                      cachedir: :self,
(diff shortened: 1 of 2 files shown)


justinstoller (approved):
Seems eminently sensible.

I'm curious how many of the actions would/could use cachedir, but it looks like Base has _very_ little in it. We can add auditing options to the "things we'd like to clean up in this repo" list.

Magisus:
Yeah it was a little difficult to follow the thread of exactly where this gets plumed through. It winds up kicking in [here](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/initializers.rb#L35-L36) mostly, I think.

mwaggett (approved):

PR #1021, opened 2020-02-05
(RK-358) Loosen puppet_forge dependency

Magisus:
This matches the loosening we did recently on later branches. The later
forge versions contain minor bug fixes.

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'
 
-  s.add_dependency 'puppet_forge', '~> 2.2.8'
+  s.add_dependency 'puppet_forge', '~> 2.3.0'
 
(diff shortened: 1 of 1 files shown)


r10k.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'
 
-  s.add_dependency 'puppet_forge', '~> 2.2.8'
+  s.add_dependency 'puppet_forge', '~> 2.3.0'
```
mwaggett, line 31: Would it be worth setting this to `'~> 2.3'` so we don't need to bump when there's a Y release?

Magisus: I wondered... there's definitely a mix of pin levels in here. I went with this because it matches what we did in master, but I could go either way.

mwaggett: I'm all for consistency, so we can just leave this for now.

mwaggett (approved):

PR #1025, opened 2020-02-20
Update documentation links

baurmatt:
There are two more in `doc/dynamic-environments/master-configuration.mkd` but I could figure out where http://docs.puppetlabs.com/puppet/latest/reference/environments.html#enabling-directory-environments should point to now.

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -48,7 +48,7 @@ cachedir: '/var/cache/r10k'
 
 
 The cachedir setting defaults to `~/.r10k`. If the HOME environment variable is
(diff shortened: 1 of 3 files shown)


justinstoller (approved):

justinstoller:
Thanks, @baurmatt ! I looked into the other references you mentioned and https://puppet.com/docs/puppet/3.8/environments_creating.html is still a valid link but the distinction between "directory" environments just isn't a distinction we do anymore. I'm tempted to remove all of the pre Puppet 3.6 docs that are no longer applicable and instead point folks to https://puppet.com/docs/puppet/latest/environments_creating.html

I can make those changes, but I'd love your feedback if you think those would be helpful for community members?

mwaggett:
+1 for removing docs links for EOL stuff

baurmatt:
@justinstoller Sounds good!

PR #1032, opened 2020-03-05
(OSOKR-23) Add GitHub Action to tag and release

mwaggett:
This commit adds a GitHub Action to tag the r10k repo, build the r10k gem, and
publish the gem to rubygems. This Action will be triggered whenever r10k's
version file gets updated. The commit message to update the version file must
include `#minor` or `#major` in order to bump the Y or X version. Otherwise,
the Z version will get bumped. For more info, see
https://github.com/marketplace/actions/github-tag-bump

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,9 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.PUPPET_RELEASE_GH_TOKEN }}
           DEFAULT_BUMP: patch
           WITH_V: false
+          # Uncomment this if the tag and version file become out-of-sync and
+          # you need to tag at a specific version.
+          # CUSTOM_TAG:
       - name: Build gem
         uses: scarhand/actions-ruby@master
         with:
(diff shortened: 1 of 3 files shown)


Magisus:
So it sounds like this means that the tag we create is not actually based on the number in the file, right? It's based on an increment of the current latest tag?

mwaggett:
@Magisus That's correct. I can try to find a different approach if we don't like that.

mwaggett:
There is a `CUSTOM_TAG` environment variable to set, but we'd either need to 1. manually update that in addition to the version file, or 2. parse the version file, which could be, uh, yucky to do inside of a yaml file.. Thoughts?

Magisus:
(side note, there appears to be an extraneous 'E' in the ticket title)

mwaggett:
(oh ya, whoops, will fix)

dhollinger (dismissed):
Nice and simple

mwaggett:
sorry @dhollinger, I made a small tweak after you ✅'d

Magisus (dismissed):

Magisus (approved):

PR #1034, opened 2020-03-06
Merge 2.6.x -> master + 3.4.1 release prep

mwaggett:


CHANGELOG.mkd
```diff
@@ -1,6 +1,17 @@
 CHANGELOG
 =========
 
+Unreleased
+----
+
+3.4.1
+----
+- Add support for Ruby 2.7
```
Magisus, line 9: I notice some of the changelog entries below have a `Changes` label. Do we think that's necessary? If so, should there be one here?

Magisus (approved):

CHANGELOG.mkd
mwaggett: yeah, I wasn't sure. This changelog is inconsistent all over the place, so I just tried to be consistent with nearby entries lol.
But if we want, I can update to standardize it.

Magisus: Nah that's okay. Seems like not a necessary label.

tvpartytonight (approved):

PR #1039, opened 2020-03-11
Make writing the env deployment file close to atomic

### LINKED ISSUE #813: Writing .r10k-deploy.json is not atomic
r10k rewrites one file: .r10k-deploy.json. This process is not done atomically by open(O_TRUNC), write, close. This means that the file is empty for some time and the empty string is no valid JSON.

We have a failure rate of about one in 1000 Puppet runs, where the config_version script is unable to determine the version as the file is empty at this time.

Please write this file to a temporary location and replace the old one atomically.

binford2k:
This isn't quite atomic, but it's really close. And AFAIK, it's as close
as we can get to atomic in pure Ruby since `flock` is advisory.

Fixes #813

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -157,7 +157,10 @@ module R10K
             logger.debug("Unable to get environment module deploy data for .r10k-deploy.json at #{environment.path}")
           end
 
-          File.open("#{environment.path}/.r10k-deploy.json", 'w') do |f|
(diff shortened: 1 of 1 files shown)


mwaggett (approved):

dhollinger (approved):

adrienthebo (approved):
`flock` is advisory, but if it's r10k fighting with itself it might still be handy.

That said we do need a better story about not running r10k in parallel.

PR #1042, opened 2020-04-03
Add `exec` environment source

reidmv:
Add a new environment source type, exec, to enable simple implementation of external environment sources.

The exec environment source runs an external command which is expected to return on stdout content compatible with the YAML environment source data format. The command may return the data in JSON or YAML form. The exec environment source is similar in purpose to Puppet's exec node terminus, used to implement external node classifiers (ENCs). R10k's exec source type allows the the implementation of external environment sources.

Video demo:
https://youtu.be/npfqRjaIxGo

The video goes for 20 minutes and includes how CD4PE/PE can use R10k with this feature. The R10k-specific part of the demo is presented early on.

### DIFF
--- a/lib/r10k/source/exec.rb
+++ b/lib/r10k/source/exec.rb
@@ -15,13 +15,11 @@ class R10K::Source::Exec < R10K::Source::Hash
     super(name, basedir, options)
   end
 
-  def environments
-    if @environments.nil?
-      set_environments_hash(run_environments_command)
-      super
-    end
-    @environments
+  def environments_hash
+    @environments_hash ||= set_environments_hash(run_environments_command)
   end
+
+  private
 
   def run_environments_command
     subproc = R10K::Util::Subprocess.new([@command])
@@ -33,7 +31,7 @@ class R10K::Source::Exec < R10K::Source::Hash
       environments = JSON.parse(procresult.stdout)
     rescue JSON::ParserError => json_err
       begin
-        environments = YAML.load(procresult.stdout)
+        environments = YAML.safe_load(procresult.stdout)
       rescue Psych::SyntaxError => yaml_err
         raise R10K::Error, _("Error parsing command output for exec source %{name}:\n" \
                              "Not valid JSON: %{j_msg}\n" \
@@ -43,8 +41,8 @@ class R10K::Source::Exec < R10K::Source::Hash
     end
 
     unless R10K::Source::Hash.valid_environments_hash?(environments)
-      raise R10K::Error, _("Environment source ${name} command %{cmd} did not return valid environment data.\n" \
-                           'Returned: %{data}') % {name: name, cmd: command, data: environments}
+      raise R10K::Error, _("Environment source %{name} command %{cmd} did not return valid environment data.\n" \
+                           'Returned: %{data}') % {name: name, cmd: @command, data: environments}
     end
 
     # Return the resulting environments hash
--- a/lib/r10k/source/exec.rb
+++ b/lib/r10k/source/exec.rb
@@ -1,0 +1,53 @@
+require 'r10k/util/subprocess'
+require 'json'
+require 'yaml'
+
+class R10K::Source::Exec < R10K::Source::Hash
+  R10K::Source.register(:exec, self)
+
+  def initialize(name, basedir, options = {})
+    unless @command = options[:command]
+      raise ConfigError, _('Environment source %{name} missing required parameter: command') % {name: name}
+    end
+
+    # We haven't set the environments option yet. We will do that by
+    # overloading the #environments method.
+    super(name, basedir, options)
+  end
+
+  def environments
+    if @environments.nil?
+      set_environments_hash(run_environments_command)
+      super
+    end
+    @environments
+  end
+
+  def run_environments_command
+    subproc = R10K::Util::Subprocess.new([@command])
+    subproc.raise_on_fail = true
+    subproc.logger = self.logger
+    procresult = subproc.execute
+
+    begin
+      environments = JSON.parse(procresult.stdout)
+    rescue JSON::ParserError => json_err
+      begin
+        environments = YAML.load(procresult.stdout)
+      rescue Psych::SyntaxError => yaml_err
+        raise R10K::Error, _("Error parsing command output for exec source %{name}:\n" \
+                             "Not valid JSON: %{j_msg}\n" \
+                             "Not valid YAML: %{y_msg}\n" \
+                             "Stdout:\n%{out}") % {name: name, j_msg: json_err.message, y_msg: yaml_err.message, out: procresult.stdout}
+      end
+    end
+
+    unless R10K::Source::Hash.valid_environments_hash?(environments)
+      raise R10K::Error, _("Environment source ${name} command %{cmd} did not return valid environment data.\n" \
+                           'Returned: %{data}') % {name: name, cmd: command, data: environments}
+    end
+
+    # Return the resulting environments hash
+    environments
+  end
+end
--- a/lib/r10k/source/hash.rb
+++ b/lib/r10k/source/hash.rb
@@ -120,6 +120,16 @@
 #
 class R10K::Source::Hash < R10K::Source::Base
 
+  include R10K::Logging
+
+  # @param hash [Hash] A hash to validate.
+  # @return [Boolean] False if the hash is obviously invalid. A true return
+  #   means _maybe_ it's valid.
+  def self.valid_environments_hash?(hash)
+    # TODO: more robust schema valiation
+    hash.is_a?(Hash)
+  end
+
   # @param name [String] The identifier for this source.
   # @param basedir [String] The base directory where the generated environments will be created.
   # @param options [Hash] An additional set of options for this source. The
@@ -131,18 +141,33 @@ class R10K::Source::Hash < R10K::Source::Base
   # @option options [Hash] :environments The hash definition of environments
   def initialize(name, basedir, options = {})
     super(name, basedir, options)
+  end
 
-    @environments_hash = options.delete(:environments) || {}
-
-    @environments_hash.keys.each do |name|
-      R10K::Util::SymbolizeKeys.symbolize_keys!(@environments_hash[name])
-      @environments_hash[name][:basedir] = basedir
-      @environments_hash[name][:dirname] = R10K::Environment::Name.new(name, {prefix: @prefix, source: @name}).dirname
+  # Set the environment hash for the source. The environment hash is what the
+  # source uses to generate enviroments.
+  # @param hash [Hash] The hash to sanitize and use as the source's environments.
+  #   Should be formatted for use with R10K::Environment#from_hash.
+  def set_environments_hash(hash)
+    @environments_hash = hash.reduce({}) do |memo,(name,opts)|
+      R10K::Util::SymbolizeKeys.symbolize_keys!(opts)
+      memo.merge({ 
+        name => opts.merge({
+          :basedir => @basedir,
+          :dirname => R10K::Environment::Name.new(name, {prefix: @prefix, source: @name}).dirname
+        })
+      })
     end
   end
 
+  # Return the sanitized environments hash for this source. The environments
+  # hash should contain objects formatted for use with R10K::Environment#from_hash.
+  # If the hash does not exist it will be built based on @options.
+  def environments_hash
+    @environments_hash ||= set_environments_hash(@options.fetch(:environments, {}))
+  end
+
   def environments
-    @environments ||= @environments_hash.map do |name, hash|
+    @environments ||= environments_hash.map do |name, hash|
       R10K::Environment.from_hash(name, hash)
     end
   end
--- a/spec/unit/source/exec_spec.rb
+++ b/spec/unit/source/exec_spec.rb
@@ -1,0 +1,72 @@
+require 'spec_helper'
+require 'r10k/source'
+require 'json'
+require 'yaml'
+
+describe R10K::Source::Exec do
+
+  let(:environments_hash) do
+    {
+      'production' => {
+        'remote'  => 'https://git.example.com/puppet/control-repo.git',
+        'ref'     => 'release-141',
+        'modules' => {
+          'puppetlabs-stdlib' => '6.1.0',
+          'puppetlabs-ntp' => '8.1.0',
+          'example-myapp1' => {
+            'git' => 'https://git.example.com/puppet/example-myapp1.git',
+            'ref' => 'v1.3.0'
+          }
+        }
+      },
+      'development' => {
+        'remote'  => 'https://git.example.com/puppet/control-repo.git',
+        'ref'     => 'master',
+        'modules' => {
+          'puppetlabs-stdlib' => '6.1.0',
+          'puppetlabs-ntp' => '8.1.0',
+          'example-myapp1' => {
+            'git' => 'https://git.example.com/puppet/example-myapp1.git',
+            'ref' => 'v1.3.1'
+          }
+        }
+      }
+    }
+  end
+
+  describe 'initialize' do
+    context 'with a valid command' do
+      context 'that produces valid output' do
+        it 'accepts json' do
+          allow_any_instance_of(R10K::Util::Subprocess)
+            .to receive(:execute)
+            .and_return(double('result', stdout: environments_hash.to_json))
+
+          source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
+          expect(source.environments.map(&:name)).to contain_exactly('production', 'development')
+        end
+
+        it 'accepts yaml' do
+          allow_any_instance_of(R10K::Util::Subprocess)
+            .to receive(:execute)
+            .and_return(double('result', stdout: environments_hash.to_yaml))
+
+          source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
+          expect(source.environments.map(&:name)).to contain_exactly('production', 'development')
+        end
+
+      end
+
+      context 'that produces invalid output' do
+        it 'raises an error for non-json, non-yaml data' do
+          allow_any_instance_of(R10K::Util::Subprocess)
+            .to receive(:execute)
+            .and_return(double('result', stdout: "one:\ntwo\n"))
+
+          source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
+          expect { source.environments }.to raise_error(/Error parsing command output/)
+        end
+      end
+    end
+  end
+end
--- a/spec/unit/source/exec_spec.rb
+++ b/spec/unit/source/exec_spec.rb
@@ -66,6 +66,15 @@ describe R10K::Source::Exec do
           source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
           expect { source.environments }.to raise_error(/Error parsing command output/)
         end
+
+        it 'raises an error for yaml data that is not a hash' do
+          allow_any_instance_of(R10K::Util::Subprocess)
+            .to receive(:execute)
+            .and_return(double('result', stdout: "[one, two]"))
+
+          source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
+          expect { source.environments }.to raise_error(R10K::Error, /Environment source execsource.*did not return valid environment data.*one.*two.*/m)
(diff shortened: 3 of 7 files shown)


lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,40 @@
...
+                           "stdout:\n%{out}\n" \
+                           "stderr:\n%{err}") % {name: name, cmd: command, code: status.exitstaus.to_s, out: stdout, err: stderr}
+    end
+
+    begin
+      environments = JSON.parse(stdout) rescue YAML.load(stdout)
+    rescue Psych::SyntaxError => e
+      raise R10K::Error, _('Error parsing command output for exec source %{name}; ' \
+                           "tried JSON first, then YAML; YAML parsing failed with: %{msg}\n" \
+                           "stdout:\n%{out}") % {name: name, msg: e.message, out: stdout}
```
mwaggett: any reason why JSON is tried first? would there be any reason we'd want that error message too?

mwaggett: would it be worth giving a way to specify which format to expect?

reidmv: Yeah, this could get cleaned up to expose the JSON parsing error too if all fails. Mostly just done this way the first time around for code brevity.

JSON is tried first because it will correctly read valid JSON, and will fail on anything else. YAML goes second because it will actually parse JSON as valid input _most of the time_, but has edge cases that could produce something weird. Trying to parse as JSON first avoids that.

Design-wise, we should not let the user specify. We can figure it out easily so there's no benefit to making the user think about it or care during configuration.

reidmv: Modified in new commit to show parsing error results of JSON and YAML attempts, if both fail.

mwaggett: ahh okay, that makes sense to me. Thanks!

mwaggett:
This looks good to me! Just wondering if there should be accompanying tests.

Also, someone more familiar with this codebase should look at this too, since I'm still new to this one. 😅

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,45 @@
...
+      raise R10K::Error, _("Error executing command for exec source %{name}: %{msg}") % {name: name, msg: e.message}
+    end
+
+    unless status.success?
+      raise R10K::Error, _("Error executing exec command for source %{name}: %{cmd}. Command exited %{code}\n" \
+                           "stdout:\n%{out}\n" \
+                           "stderr:\n%{err}") % {name: name, cmd: command, code: status.exitstaus.to_s, out: stdout, err: stderr}
+    end
+
+    begin
```
adrienthebo, line 30: Supporting two different serialization formats adds complexity, can we start with JSON support to start with and add YAML support if the demand manifests for that format?

If we do need to support multiple serialization formats, that should be configured at the source level, with something like this:

```
sources:
  exec:
    command: "..."
    format: "yaml"
```

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,45 @@
...
+class R10K::Source::Exec < R10K::Source::Hash
+  R10K::Source.register(:exec, self)
+
+  def initialize(name, basedir, options = {})
+    unless command = options[:command]
+      raise ConfigError, _('Environment source %{name} missing required parameter: command') % {name: name}
+    end
+
+    begin
+      stdout, stderr, status = Open3.capture3(command)
```
adrienthebo: Is there a use case for calling `open3` over using [`R10K::Util::Subprocess](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/util/subprocess.rb)?

adrienthebo (dismissed):
Looks good overall, thanks @reidmv! If possible let's use the r10k subprocess utility for command execution, simplify format handling if possible, and I'd like to see some basic test coverage for this. Once those concerns are concerned we should be good to merge this.

Expanding on the format handling, I'm mildly opposed to supporting multiple formats without a concrete use case; this is an easy simplification and avoids the mild heuristics that multiple formats entail. For comparison the puppet exec node terminus only supports YAML so we have prior art for supporting a single format. That being said, I don't feel strongly - if there's a compelling case for multiple formats, then we can proceed as-is.

Summary of the above:

* [x] Command execution with `Subprocess`
* [x] Format handling simplification (or documentation for multiple format use cases)
* [x] Test coverage

Ping me in the Puppet community Slack if you'd like to discuss this more. Thanks!

lib/r10k/source/exec.rb
reidmv: Nope, just not familiar with R10K::Util::Subprocess. Modifying to use that...

lib/r10k/source/exec.rb
reidmv: We could start with JSON. That is a choice we could make.

Design-wise, regardless of what we decide we should not let the user specify. We can figure out JSON vs. YAML trivially so there's never going to be a benefit to making the user think about it or care during configuration. We should optimize for keeping the configuration simple.

My design preference is to support both. YAML is prevalent in the R10K ecosystem, and is the input format accepted by the conceptually related ENC feature in Puppet. However, the use case that's driving this feature is to retrieve environment information from REST APIs, which typically return JSON. The code complexity to just accept both is minimal. This is effectively a user interface so anything we can do to make dealing with it simpler for the user is a win.

spec/unit/source/exec_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+            .and_return(double('result', stdout: environments_hash.to_yaml))
+
+          source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
+          expect(source.environments.map(&:name)).to contain_exactly('production', 'development')
+        end
+
+      end
+
+      context 'that produces invalid output' do
+        it "rejects schema invalid data"
```
mwaggett: Did you mean to add something else here?

reidmv: Nope, that's a pending test. It doesn't currently reject schema invalid data, but it seems like it should. Not yet implemented.

lib/r10k/source/exec.rb
reidmv: To account for yaml being more permissive than JSON (e.g. the text string "200 OK" is valid yaml), it would be best to implement validation of the data read though. For today/tonight I added the following pending test, but the PR doesn't currently implement it.

```ruby
it "rejects schema invalid data"
```

spec/unit/source/exec_spec.rb
reidmv: Replaced this pending test with a slightly improved one. Only "schema" validation being enforced right now is that data returned needs to be a hash. I didn't see any more robust schema validation libraries already in use in R10K, and don't think we necessarily need to add one just for this. Not right now anyway. Maybe later.

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,43 @@
...
+        raise R10K::Error, _("Error parsing command output for exec source %{name}:\n" \
+                             "Not valid JSON: %{j_msg}\n" \
+                             "Not valid YAML: %{y_msg}\n" \
+                             "Stdout:\n%{out}") % {name: name, j_msg: json_err.message, y_msg: yaml_err.message, out: procresult.stdout}
+      end
+    end
+
+    unless R10K::Source::Hash.valid_environment_hash?(environments)
+      raise R10K::Error, _("Environment source ${name} command %{cmd} did not return valid environment data.\n" \
+                           'Returned: %{dat}') % {name: name, cmd: command, dat: environments}
```
mwaggett: nitpick, but why not just `data`? personally, I like to only abbreviate super obvious things, e.g. `directory` -> `dir`, `command` -> `cmd`.

spec/unit/source/exec_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+            .and_return(double('result', stdout: environments_hash.to_yaml))
+
+          source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
+          expect(source.environments.map(&:name)).to contain_exactly('production', 'development')
+        end
+
+      end
+
+      context 'that produces invalid output' do
+        it "rejects strings as schema invalid data" do
```
mwaggett: Is there a `hash_spec` that this should go in, since that's the class doing the actual validation?

reidmv: @mwaggett wanna go merge #1046? It would be easy to move this test over there if that PR gets merged 😁

lib/r10k/source/exec.rb
reidmv: Haha, sure. Changed to "data".

spec/unit/source/exec_spec.rb
mwaggett: done 😁

reidmv: Test moved

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,43 @@
+require 'r10k/util/subprocess'
+require 'json'
+require 'yaml'
+
+class R10K::Source::Exec < R10K::Source::Hash
+  R10K::Source.register(:exec, self)
+
+  def initialize(name, basedir, options = {})
```
adrienthebo, line 8: All of the logic within this class is happening within the constructor, so you can't create an instance of this class without triggering side effects. Instead of using inheritance to use the `R10K::Source::Hash` class we can use composition, with something like the following:

```ruby
class R10K::Source::Exec < R10K::Source::Base
  def initialize(...)
    # ..
    @inner = null
  end

  def environments
    inner.environments
  end

  def desired_contents
    inner.desired_contents
  end
  
  private

  def inner
    if @inner.nil?
      envs = run_command_and_get_envs_hash
      @inner = R10K::Source::Hash.new(@name, @basedir, environments: envs)
    end
    @inner
  end
end
```

Thoughts?

reidmv: Pushed some new commits that should allow an instance of the class to be created without side-effects. Running of the configured command is now delayed until the source's `#environments` method is called.

lib/r10k/source/hash.rb
```diff
@@ -131,18 +141,26 @@ class R10K::Source::Hash < R10K::Source::Base
...
     super(name, basedir, options)
+  end
 
-    @environments_hash = options.delete(:environments) || {}
-
-    @environments_hash.keys.each do |name|
-      R10K::Util::SymbolizeKeys.symbolize_keys!(@environments_hash[name])
-      @environments_hash[name][:basedir] = basedir
-      @environments_hash[name][:dirname] = R10K::Environment::Name.new(name, {prefix: @prefix, source: @name}).dirname
+  def set_environments_hash(hash)
```
adrienthebo, line 150: Non-blocker: documentation would be good here.

lib/r10k/source/hash.rb
```diff
@@ -131,18 +141,26 @@ class R10K::Source::Hash < R10K::Source::Base
...
+      memo.merge({ 
+        name => opts.merge({
+          :basedir => @basedir,
+          :dirname => R10K::Environment::Name.new(name, {prefix: @prefix, source: @name}).dirname
+        })
+      })
     end
   end
 
+  def environments_hash
```
adrienthebo, line 165: Non-blocker: documentation would be good here.

lib/r10k/source/hash.rb
```diff
@@ -131,18 +141,26 @@ class R10K::Source::Hash < R10K::Source::Base
...
+    @environments_hash ||= set_environments_hash(@options.delete(:environments) || {})
```
adrienthebo: Do we need to remove this from the `@options` hash, or can this be an `@options.fetch(:environments, {})`?

adrienthebo (dismissed):
LGTM! There are a couple of nice to have changes (mainly documentation) that would be good to add before merge, but I'm happy with this as-is.

lib/r10k/source/hash.rb
```diff
@@ -120,6 +120,16 @@
...
 
+  include R10K::Logging
+
+  # @param hash [Hash] A hash to validate.
+  # @return [Boolean] False if the hash is obviously invalid. A true return
+  #   means _maybe_ it's valid.
+  def self.valid_environments_hash?(hash)
+      # TODO: more robust schema valiation
+      hash.is_a?(Hash)
+  end
```
adrienthebo:
```suggestion
  def self.valid_environments_hash?(hash)
    # TODO: more robust schema valiation
    hash.is_a?(Hash)
  end
```

lib/r10k/source/hash.rb
reidmv, line 165:
```suggestion
  # Return the sanitized environments hash for this source. The environments
  # hash should contain objects formatted for use with R10K::Environment#from_hash.
  # If the hash does not exist it will be built based on @options.
  def environments_hash
```

lib/r10k/source/hash.rb
reidmv:
```suggestion
    @environments_hash ||= set_environments_hash(@options.fetch(:environments, {}))
```

lib/r10k/source/hash.rb
reidmv, line 150:
```suggestion
  # Set the environment hash for the source. The environment hash is what the
  # source uses to generate enviroments.
  # @param hash [Hash] The hash to sanitize and use as the source's environments.
  #   Should be formatted for use with R10K::Environment#from_hash.
  def set_environments_hash(hash)
```

lib/r10k/source/hash.rb
reidmv: Nope, we don't need to remove it; fetch would be better!

reidmv:
Hmm. The failing test here looks like it originated from #952. Not sure what to do about that; I don't even get the failure locally. #952 passed tests before Ruby 2.7.0 was added to the matrix.

reidmv:
Tracked it down. The test failure is due to an rspec-mocks Ruby 2.7 bug in its `#and_call_original` method. See https://github.com/rspec/rspec-mocks/issues/1306.

reidmv:
@adrienthebo ready for review, including bug workaround for rspec-mocks 3.9.1

mwaggett:
lgtm! did you want to squash any of your commits?

reidmv:
Squashed down to three commits. 1) Improve hash source type, 2) add exec source type, and 3) the rspec-mocks test bug workaround.

mwaggett (dismissed):

spec/unit/cli_spec.rb
```diff
@@ -178,6 +178,17 @@ def check_command(command, parents = [])
 
   include_context 'cli'
 
+  ## TODO: when rspec-mocks fixes #1306, delete this block
```
adrienthebo: We should be able to drop this once #1047 is merged.

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,53 @@
...
+    subproc = R10K::Util::Subprocess.new([@command])
+    subproc.raise_on_fail = true
+    subproc.logger = self.logger
+    procresult = subproc.execute
+
+    begin
+      environments = JSON.parse(procresult.stdout)
+    rescue JSON::ParserError => json_err
+      begin
+        environments = YAML.load(procresult.stdout)
```
adrienthebo: This should be `YAML.safe_load`.
```suggestion
        environments = YAML.safe_load(procresult.stdout)
```

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,53 @@
...
+        raise R10K::Error, _("Error parsing command output for exec source %{name}:\n" \
+                             "Not valid JSON: %{j_msg}\n" \
+                             "Not valid YAML: %{y_msg}\n" \
+                             "Stdout:\n%{out}") % {name: name, j_msg: json_err.message, y_msg: yaml_err.message, out: procresult.stdout}
+      end
+    end
+
+    unless R10K::Source::Hash.valid_environments_hash?(environments)
+      raise R10K::Error, _("Environment source ${name} command %{cmd} did not return valid environment data.\n" \
+                           'Returned: %{data}') % {name: name, cmd: command, data: environments}
```
adrienthebo: Blocker: the error handling in this code will fail if called.

1. The `${name}` variable expansion isn't a valid template expansion token
2. `@command` doesn't have an attr_reader so this code will throw an exception during error handling:

```suggestion
      raise R10K::Error, _("Environment source %{name} command %{cmd} did not return valid environment data.\n" \
                           'Returned: %{data}') % {name: name, cmd: @command, data: environments}
```

I've added a test for this as well:

```diff
diff --git i/lib/r10k/source/exec.rb w/lib/r10k/source/exec.rb
index 8460699..87a3a0e 100644
--- i/lib/r10k/source/exec.rb
+++ w/lib/r10k/source/exec.rb
@@ -43,8 +43,8 @@ class R10K::Source::Exec < R10K::Source::Hash
     end
 
     unless R10K::Source::Hash.valid_environments_hash?(environments)
-      raise R10K::Error, _("Environment source ${name} command %{cmd} did not return valid environment data.\n" \
-                           'Returned: %{data}') % {name: name, cmd: command, data: environments}
+      raise R10K::Error, _("Environment source %{name} command %{cmd} did not return valid environment data.\n" \
+                           'Returned: %{data}') % {name: name, cmd: @command, data: environments}
     end
 
     # Return the resulting environments hash
diff --git i/spec/unit/source/exec_spec.rb w/spec/unit/source/exec_spec.rb
index 51fc069..fa36f7c 100644
--- i/spec/unit/source/exec_spec.rb
+++ w/spec/unit/source/exec_spec.rb
@@ -66,6 +66,15 @@ describe R10K::Source::Exec do
           source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
           expect { source.environments }.to raise_error(/Error parsing command output/)
         end
+
+        it 'raises an error for yaml data that is not a hash' do
+          allow_any_instance_of(R10K::Util::Subprocess)
+            .to receive(:execute)
+            .and_return(double('result', stdout: "[one, two]"))
+
+          source = described_class.new('execsource', '/some/nonexistent/dir', command: '/path/to/command')
+          expect { source.environments }.to raise_error(R10K::Error, /Environment source execsource.*did not return valid environment data.*one.*two/)
+        end
       end
     end
   end
```

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def environments
+    if @environments.nil?
+      set_environments_hash(run_environments_command)
+      super
+    end
+    @environments
+  end
+
+  def run_environments_command
```
adrienthebo, line 24: Non-blocker: this method should be private or protected.

lib/r10k/source/exec.rb
```diff
@@ -0,0 +1,53 @@
...
+      raise ConfigError, _('Environment source %{name} missing required parameter: command') % {name: name}
+    end
+
+    # We haven't set the environments option yet. We will do that by
+    # overloading the #environments method.
+    super(name, basedir, options)
+  end
+
+  def environments
+    if @environments.nil?
```
adrienthebo: Non-blocker: `R10K::Source::Hash` has `#environments` which lazily sets `@environments` from `#environments_hash`. Instead of overriding `#environments` we can override `#environments_hash`, which drops the need for the conditional call to `super`.

This is a good candidate for a refactor in a second PR, so we can skip this work in this PR.

adrienthebo (dismissed):
We have a hard blocker for error handling, a minor change to YAML parsing, and some optional minor improvements. Let's get the blocker cleared out so we can get this merged, and then can do some refactoring in a followup PR.

spec/unit/cli_spec.rb
mwaggett: I think the error this fixes is unrelated to what #1047 fixes.

lib/r10k/source/exec.rb
mwaggett: thanks for looking at this more closely than I did 😅

reidmv:
Applied suggestions from code review, dropped rspec-mocks cli tests workaround, rebased on master.

mwaggett (approved):

lib/r10k/source/exec.rb
reidmv: Made method private.

lib/r10k/source/exec.rb
reidmv: Ended up implementing this suggestion along with the hard blocker fixes

adrienthebo (approved):
LGTM!

PR #1043, opened 2020-04-13
Add Supercronic - to be used in K8s

Xtigyro:
`supercronic` is needed - so r10k can be scheduled in a sidecar container in K8s. 

That will allow Puppet Server, deployed through the Helm chart, to run on different K8s nodes (even distributed in different cloud zones).

### DIFF
--- a/docker/r10k/Dockerfile
+++ b/docker/r10k/Dockerfile
@@ -19,7 +19,7 @@ ARG pupperware_analytics_stream="dev"
 ARG supercronic_version="0.1.9"
 ARG supercronic_sha1sum="5ddf8ea26b56d4a7ff6faecdd8966610d5cb9d85"
 ARG supercronic="supercronic-linux-amd64"
-ARG supercronic_url="https://github.com/aptible/supercronic/releases/download/v$supercronic_version/supercronic-linux-amd64"
+ARG supercronic_url="https://github.com/aptible/supercronic/releases/download/v$supercronic_version/$supercronic"
 
 LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
       org.label-schema.vendor="Puppet" \
@@ -44,6 +44,7 @@ LABEL org.label-schema.version="$version" \
       org.label-schema.build-date="$build_date"
 
 COPY --from=build /workspace/r10k.gem /
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 RUN chmod a+x /adduser.sh && \
 # Add a puppet user to run r10k as for consistency with puppetserver
     /adduser.sh && \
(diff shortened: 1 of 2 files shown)


underscorgan:
Hiya @Xtigyro ! Thanks for this. Have you tried using k8s CronJobs instead of supercronic? Mostly curious if that would be a workable solution https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/

Xtigyro:
> 
> 
> Hiya @Xtigyro ! Thanks for this. Have you tried using k8s CronJobs instead of supercronic? Mostly curious if that would be a workable solution https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/

Hey @underscorgan!

At present the Helm chart uses exactly that. One for r10k that syncs a Puppet Control Repo and another one for Hiera data. 
-- https://github.com/puppetlabs/pupperware/blob/master/k8s/templates/r10k-code.cronjob.yaml
-- https://github.com/puppetlabs/pupperware/blob/master/k8s/templates/r10k-hiera.cronjob.yaml

Using a K8s CronJob is an efficient way for achieving regular runs of a task. However, a CronJob object cannot be included in a K8s Deployment. Which means that the Puppet Server K8s Deployment cannot control the r10k cronjobs. However, r10k and Puppet Server must have common storage for the synced Puppet code. To achieve that, we've utilized in the chart optional K8s node selectors or pod affinity contstraints for the K8s pods of the r10k cronjobs and the Puppet Server. That is suboptimal though working fine because you always schedule the Puppet Server pod and the r10k cronjob pods on the same K8s node. Up until recently the chart supported running just one Puppet Server and everything was OK. The chart at present supports running multiple Puppet masters and it still optionally (but also needed) enforces scheduling all of them on one K8s node because of the aforementioned limitation. It also leads to a lack of a true high availability.

Putting r10k control repo and hiera data cronjobs in sidecar containers allows the K8s deployment of Puppet server to control them. It will also solve the mandatory common storage between them - because by definition they will run in the same pod. That will accommodate the chart to schedule in the future (still in development) the different Puppet masters on different K8s nodes even when those nodes are managed in the cloud and sit in different cloud zones. It will also simplify considerably and  without much hassle the usage of the chart in complex cloud topologies.

underscorgan:
ah, ok, gotcha. Thanks for the explanation!! Also I wanted to let you know I haven't forgotten about the fixes I'm going to help with here, just got sidetracked by some other things. Hopefully tomorrow :)

underscorgan:
I have a PR up against your branch that should resolve lint issues and make sure the changes are consistent in both Dockerfiles

Xtigyro:
> 
> 
> I have a PR up against your branch that should resolve lint issues and make sure the changes are consistent in both Dockerfiles

Merged. Sweet - thank you, Morgan!

underscorgan (approved):
looks good to me! 🎉

Xtigyro:
The major revamp of the Puppet Server Helm chart - which is 90 % ready to be released - relies on this PR, guys.

I would like to ask if we could merge this and release a new version of the `r10k` Docker image with it?

Iristyle:
Superseded by #1062

Xtigyro:
@Iristyle Is it safe to delete the forked `r10k` repo in my account?

PR #1045, opened 2020-04-15
Fix Hash source to support prefix correctly

reidmv:
Previously, sources derived from R10k::Source::Hash did not correctly
handle prefixes. This commit fixes that.

### DIFF
--- a/lib/r10k/source/hash.rb
+++ b/lib/r10k/source/hash.rb
@@ -135,10 +135,9 @@ class R10K::Source::Hash < R10K::Source::Base
     @environments_hash = options.delete(:environments) || {}
 
     @environments_hash.keys.each do |name|
-      # TODO: deal with names that aren't valid
-      ::R10K::Util::SymbolizeKeys.symbolize_keys!(@environments_hash[name])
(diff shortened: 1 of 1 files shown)


lib/r10k/source/hash.rb
```diff
@@ -135,16 +135,15 @@ def initialize(name, basedir, options = {})
...
       @environments_hash[name][:basedir] = basedir
-      @environments_hash[name][:dirname] = name
+      @environments_hash[name][:dirname] = R10K::Environment::Name.new(name, {prefix: @prefix, source: @name}).dirname
     end
   end
 
   def environments
-    @environments ||= @environments_hash.map do |name, hash|
-      R10K::Environment.from_hash(name, hash)
+    @environments ||= @environments_hash.map do |key, hash|
```
mwaggett: why change this from `name`?

reidmv: Good question; there is no good reason. Leftover crumb from previous experiments. Cleaning that up now...

mwaggett (approved):

adrienthebo (approved):
LGTM. See #1046 for test coverage for this feature.

PR #1047, opened 2020-04-22
Revert "Move deploy options into subcommands that actually support them" + Revert "Add R10K::CLI spec tests"

mwaggett:
Please add all notable changes to the "Unreleased" section of the CHANGELOG.

PR #1048, opened 2020-04-23
Add CLI support for deploy module cachedir option

nabertrand:
These changes are an addition to #952 and should allow #1020 to work correctly and prevent #1047 from needing to be merged.

### DIFF
--- a/lib/r10k/action/deploy/display.rb
+++ b/lib/r10k/action/deploy/display.rb
@@ -81,7 +81,14 @@ module R10K
         end
 
         def allowed_initialize_opts
-          super.merge(puppetfile: :self, detail: :self, format: :self, fetch: :self)
+          super.merge(cachedir: :self,
+                      detail: :self,
+                      fetch: :self,
+                      format: :self,
+                      'generate-types': :self,
+                      'no-force': :self,
+                      puppetfile: :self,
+                      'puppet-path': :self)
         end
       end
     end
--- a/lib/r10k/cli/deploy.rb
+++ b/lib/r10k/cli/deploy.rb
@@ -80,6 +80,7 @@ It will load the Puppetfile configurations out of all environments, and will
 try to deploy the given module names in all environments.
           DESCRIPTION
 
+          required nil, :cachedir, 'Specify a cachedir, overriding the value in config'
           required :e, :environment, 'Update the modules in the given environment'
           flag nil, :'generate-types', 'Run `puppet generate types` after updating an environment'
           flag nil, :'no-force', 'Prevent the overwriting of local module modifications'
--- a/lib/r10k/cli/deploy.rb
+++ b/lib/r10k/cli/deploy.rb
@@ -21,6 +21,17 @@ module R10K::CLI
 (https://puppet.com/docs/puppet/latest/environments_about.html).
         DESCRIPTION
 
+        required nil, :cachedir, 'Specify a cachedir, overriding the value in config'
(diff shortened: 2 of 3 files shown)


Magisus:
We were also seeing issues using `--cachedir` with `r10k deploy environment`. I'm working on replicating this with standalone r10k right now (we saw it in the context of code-manager). Will post back with what I find.

Magisus:
Narrowed down the issue! Placing the flags earlier in the command, between `deploy` and `environment`, no longer works. This is what we're doing in code-manager.
```
[root@r10k-test ~]# r10k deploy --cachedir ~/r10k environment production
deploy: unrecognised option -- cachedir
[root@r10k-test ~]# r10k deploy environment production --cachedir ~/r10k
[root@r10k-test ~]#
```
I consider this a general breaking change to the r10k API, that we should be wary of releasing.

nabertrand:
Thanks for tracking that down, @Magisus. I had moved some options into subcommands as not all subcommand Actions actually supported/used the options. I'll move them back to the deploy root so as to not introduce a breaking change.

nabertrand:
It looks like the Travis build failure is due to a warning caused by rspec-mocks introduced in ruby 2.7.0. @reidmv appears to have solved it in #1042: https://github.com/puppetlabs/r10k/pull/1042/files#diff-54f8ce220443a0a76cdb13549523a13dR181-R191. I'm not sure why the Azure builds are failing.

reidmv:
Re: rspec-mocks issue and the fix for that, I think #1042 is ready to merge, just needs a second reviewer. If for some reason that isn't the case I'm happy to pull out the rspec-mocks workaround into a separate PR.

adrienthebo:
@nabertrand I spent some time working on the Ruby 2.7 regression triggered by #952, and I'd like to take a different approach for CLI test coverage. The current approach uses a schema of sorts to describe the CLI which has the benefit of little repetition in test coverage, but this requires some trade-offs. When I looked at the tests and data it was non-trivial to see what parts of the CLI schema mapped to which test, and it was difficult to understand what the tests were actually validating. In addition the logic for building the tests was quite involved, and my experience is that generating test coverage with complex logic leads to increased costs for test development.

I'd like to step back a bit and re-add test coverage for the CLI gradually, starting with the top level `r10k` command and then gradually expanding that to the other commands. We can use shared examples and contexts to reuse logic but I would like to see test coverage clearly grouped for commands and options instead of the current schema style coverage. Is this something you'd be interested in working on?

nabertrand:
@adrienthebo I will admit the logic ended up quite complicated and understand the desire to use a simpler solution. I'll take a look to see if I can re-use some of my tests in a simpler fashion.

mwaggett:
It looks like this pull request hasn't seen any activity in several months. @nabertrand Is this something that you're still able to work on? We're happy to keep this pull request open if there's forward movement, and we'll leave it open for another two weeks from now. If there isn't any activity by then we may have to close this pull request due to lack of activity.

Closing the pull request wouldn't mean we don't consider this change valuable. However we try to keep all open pull requests moving towards completion, and closing stale pull requests help us focus on more active pull requests.

If you have any questions or concerns, please don't hesitate to ping us in the [Puppet Community Slack](http://slack.puppet.com/).

binford2k:
Thanks for the contribution @nabertrand but I'm going to close this for now due to inactivity. Feel free to open it up again if you end up having time to work on it. Thanks for everything you do!

PR #1057, opened 2020-04-30
(maint) Remove second non_module_object from test

mwaggett:
After increasing the default pool_size, a bug surfaced when attempting to
install multiple objects from the same git source. We attempt to cache both
objects, but the second attempt fails because we try to clone to the same
destination. We don't think this is common behavior and this test does not need
two objects to test what it's testing.

Please add all notable changes to the "Unreleased" section of the CHANGELOG.

### DIFF
--- a/integration/tests/purging/content_not_purged_at_root.rb
+++ b/integration/tests/purging/content_not_purged_at_root.rb
@@ -32,11 +32,6 @@ mod 'non_module_object_1',
   :install_path => './',
   :git => 'git://github.com/puppetlabs/control-repo.git',
   :branch => 'production'
-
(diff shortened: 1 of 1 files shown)


Magisus (approved):

Magisus:
I've filed https://github.com/puppetlabs/r10k/issues/1058 about the bug this uncovered. We can revisit this decision if there are any comments on that issue that this is a common workflow.

mwaggett:
Confirmed that this passes now.

justinstoller:
I'd like to block this until we get feedback on the associated issue regarding the scope of those affected.

Issue #1058, opened 2020-04-30
Git cache breaks with pool_size > 1 if managing the same git source twice


Magisus:
A bug recently surfaced in our beaker acceptance tests, that seems valid, but might not be a common workflow. It appeared after the merge of https://github.com/puppetlabs/r10k/pull/1044, updating the default pool size to 4.

If you are managing two objects (in this case non-module objects) backed by the same git repo, and are using a `pool_size` of more than 1, the caching mechanism for git behaves poorly on the initial deploy.

**Puppetfile**
```
mod 'non_module_object_1',
  :install_path => './',
  :git => 'git://github.com/puppetlabs/control-repo.git',
  :branch => 'production'
mod 'non_module_object_2',
 :install_path => '',
 :git => 'git://github.com/puppetlabs/control-repo.git',
 :branch => 'production'
```
 
**Error output** at `--verbose=debug2` from `r10k deploy environment -p`:
```
[2020-04-30 22:28:15 - INFO] Deploying Puppetfile content /etc/puppetlabs/code/environments/production/non_module_object_1
[2020-04-30 22:28:15 - DEBUG1] Creating new git cache for "git://github.com/puppetlabs/control-repo.git"
[2020-04-30 22:28:15 - DEBUG] Module thread 23963080 exiting: queue empty
[2020-04-30 22:28:15 - DEBUG2] Starting process: ["git", "clone", "--mirror", "git://github.com/puppetlabs/control-repo.git", "/var/cache/r10k/git---github.com-puppetlabs-control-repo.git"]
[2020-04-30 22:28:15 - INFO] Deploying Puppetfile content /etc/puppetlabs/code/environments/production/non_module_object_2
[2020-04-30 22:28:15 - DEBUG1] Creating new git cache for "git://github.com/puppetlabs/control-repo.git"
[2020-04-30 22:28:15 - DEBUG2] Starting process: ["git", "clone", "--mirror", "git://github.com/puppetlabs/control-repo.git", "/var/cache/r10k/git---github.com-puppetlabs-control-repo.git"]
[2020-04-30 22:28:15 - DEBUG] Module thread 23579460 exiting: queue empty
[2020-04-30 22:28:15 - DEBUG2] Finished process:
Command: git clone --mirror git://github.com/puppetlabs/control-repo.git /var/cache/r10k/git---github.com-puppetlabs-control-repo.git
Stderr:
[...]

PR #1059, opened 2020-05-01
[#1058] Serially deploy modules that share a cachedir

justinstoller:
When we raised the default threadpool for deploying modules above 1, we
began to see issues when caching a remote locally if that remote/cache
was shared by multiple modules within the Puppetfile.

To resolve this, when we create the module we group it with any other
module that may share its cachedir. We then pull modules by cachedir off
of the queue and serially process those.

Please add all notable changes to the "Unreleased" section of the CHANGELOG.

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -42,6 +42,11 @@ class Puppetfile
   #   @return [Boolean] Overwrite any locally made changes
   attr_accessor :force
 
+  # @!attribute [r] modules_by_vcs_cachedir
+  #   @api private Only exposed for testing purposes
+  #   @return [Hash{:none, String => Array<R10K::Module>}]
+  attr_reader :modules_by_vcs_cachedir
+
   # @param [String] basedir
   # @param [String] moduledir The directory to install the modules, default to #{basedir}/modules
   # @param [String] puppetfile_path The path to the Puppetfile, default to #{basedir}/Puppetfile
@@ -58,6 +63,7 @@ class Puppetfile
 
     @modules = []
     @managed_content = {}
+    @modules_by_vcs_cachedir = {}
     @forge   = 'forgeapi.puppetlabs.com'
 
     @loaded = false
@@ -137,6 +143,9 @@ class Puppetfile
     mod.origin = 'Puppetfile'
 
     @managed_content[install_path] << mod.name
+    cachedir = module_vcs_cachedir(mod)
+    @modules_by_vcs_cachedir[cachedir] ||= []
+    @modules_by_vcs_cachedir[cachedir] << mod
     @modules << mod
   end
 
@@ -180,6 +189,19 @@ class Puppetfile
 
   private
 
+  def module_vcs_cachedir(mod)
+    if mod.respond_to? :repo
+      repo = mod.repo
+      if repo.respond_to? :cache
+        cache = repo.cache
+        if cache.respond_to? :sanitized_dirname
+          return cache.sanitized_dirname
+        end
+      end
+    end
+    :none
+  end
+
   def serial_accept(visitor)
     visitor.visit(:puppetfile, self) do
       modules.each do |mod|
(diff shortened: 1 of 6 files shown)


Magisus (approved):
This looks like it'll do the trick, while maintaining as much parallelism as is safe.

Magisus:
I heard plans to add an acceptance test, which I think we should do. Was that going to be added to this PR? Or should we add this in, use the [existing test](https://github.com/puppetlabs/r10k/blob/master/integration/tests/purging/content_not_purged_at_root.rb) that failed because of this as-is to verify it, and add the new one later (in the interest of the release timeline)?

spec/unit/puppetfile_spec.rb
```diff
@@ -289,12 +308,38 @@ def expect_wrapped_error(orig, pf_path, wrapped_error)
...
+        expect(type).to eq :puppetfile
+        expect(other).to eq subject
+        block.call
+      end
+
+      mod1 = spy('module1')
+      mod2 = spy('module2')
+      mod3 = spy('module3')
+      mod4 = spy('module4')
+      mod5 = spy('module4')
```
mwaggett, line 331: is this intentionally still 'module4'? or should it be 5?

spec/unit/puppetfile_spec.rb
```diff
@@ -289,12 +308,38 @@ def expect_wrapped_error(orig, pf_path, wrapped_error)
...
+      mod6 = spy('module5')
+
+      expect(subject).to receive(:modules_by_vcs_cachedir)
+        .and_return({:none => [mod1, mod2],
+                     "foo-cachedir" => [mod3, mod4],
+                     "bar-cachedir" => [mod5, mod6]})
+
+      queue = subject.modules_queue(visitor)
+      expect(queue.length).to be 4
+      queue_array = 4.times.map { queue.pop }
```
mwaggett, line 341: Can this be `length.times.map` instead of hard-coding 4?

mwaggett:
If we're in a rush, I'm okay using the existing test, but I do think we should ultimately have one that's explicitly testing this. I'll leave it up to @justinstoller to decide whether it can be done quick enough to be included here.

adrienthebo:
I was under the mistaken assumption that serializing access to caches had been resolved as part of the parallel modules pull request. Given that this is still an issue we may want to consider dropping the default `pool_size` back to 1 while we verify that parallelization is safe. Part of this concern is that we have the [`WithModules`](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/environment/with_modules.rb#L55-L64) environment that has another entrypoint into interacting with modules that this pull request might not account for.

lib/r10k/puppetfile.rb
```diff
@@ -180,6 +189,19 @@ def accept(visitor)
 
   private
 
+  def module_vcs_cachedir(mod)
```
adrienthebo, line 192: As this method deals with properties of modules, we can push this detail down into the modules themselves instead of probing objects for methods and acting from there.

lib/r10k/puppetfile.rb
```diff
@@ -212,15 +234,22 @@ def concurrent_accept(visitor, pool_size)
   def modules_queue(visitor)
     Queue.new.tap do |queue|
       visitor.visit(:puppetfile, self) do
-        modules.each { |mod| queue << mod }
+        modules_by_cachedir = modules_by_vcs_cachedir.clone
```
adrienthebo, line 237: This might have some interesting interactions with https://github.com/puppetlabs/r10k/blob/master/lib/r10k/environment/with_modules.rb#L55-L64.

justinstoller:
How about this? https://github.com/puppetlabs/r10k/pull/1060

justinstoller:
Ugh, it looks like the `WithModules` implementation maybe just copied and pasted code from the `Puppetfile` implementation? We probably should refactor out that shared code.

lib/r10k/puppetfile.rb
dhollinger: I would feel safer if this was addressed before the PR is merged. Just to avoid any potential unforeseen issues.

dhollinger (commented):
I don't really have anything additional to add that @adrienthebo and @mwaggett haven't already pointed out. I would like to see the potential interaction with `WithModules` to be solved before merging though

justinstoller:
Yeah, sorry, I didn't mean to ghost, y'all. When we thought we could make this in before a deadline we gave it a go, and when it became apparent we weren't going to make it for that deadline it dropped way down our priority list.

lib/r10k/puppetfile.rb
Magisus: I may be missing something, but I'm not sure how this would matter in `WithModules`. That class doesn't have a `concurrent_accept` mode so far as I can tell, so its modules would be processed first, always in serial, and then the Puppetfile would be processed and this logic would handle the potential conflict.

I can't find any other other places that respect `pool_size` besides what's in the Puppetfile here. What did y'all think was potentially problematic?

Magisus:
I don't appear to be able to push to this branch. Is that something @justinstoller can change, or do I need to open a new PR to pick this up?

justinstoller:
I pushed this code to this branch so anyone can take it over: https://github.com/puppetlabs/r10k/tree/GH-1058_dont-concurrently-modify-same-cache

spec/unit/puppetfile_spec.rb
Magisus: I think it doesn't really matter, since calling `spy` will still create unique objects regardless of that string. But I'll update it.

Magisus:
Superseded by https://github.com/puppetlabs/r10k/pull/1081

PR #1044, opened 2020-04-15
Increase default pool_size to 4

npwalker:
Solves #1038.  pool_size was added as an optional setting
a while ago and people have been using it successfully
in the wild.  This commit increases the default to 4
which seems like a generally useful improvement.

Please add all notable changes to the "Unreleased" section of the CHANGELOG.

PR #1060, opened 2020-05-05
Revert increase in default pool size

justinstoller:
After updating the default pool_size, a race condition was
found when initializing our local git cache. See
#1058 for the issue in detail.

When attempting a fix we realized that there were multiple copies of the
code to iterate over modules to manage. See
#1059 for a first attempt at this
problem and discussion.

In order to meet our current timelines, the default will need to reverted
until we can fix this issue properly. Hence this commit and a55536c.

This commit was necessary because the CHANGELOG has been refactored since
the commit referencing the default was added.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -3,7 +3,6 @@ CHANGELOG
 
 Unreleased
 ----
-- Increase `pool_size` default to 4 from 1 [#1044](https://github.com/puppetlabs/r10k/pull/1044)
 - Add exec environment source type. The exec source type allows for the implementation of external environment sources [#1042](https://github.com/puppetlabs/r10k/pull/1042)
 
 3.4.1
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -10,7 +10,7 @@ class Puppetfile
 
   include R10K::Settings::Mixin
 
-  def_setting_attr :pool_size, 4
+  def_setting_attr :pool_size, 1
 
   include R10K::Logging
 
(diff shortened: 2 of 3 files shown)


mwaggett:
@justinstoller Can you add some info to the commit message (maybe linking to the issue) about _why_ we're doing this?

spec/unit/action/puppetfile/install_spec.rb
```diff
@@ -19,9 +19,7 @@ def installer(opts = {}, argv = [], settings = {})
 
   describe "installing modules" do
     let(:modules) do
-      (1..4).map do |idx|
-        R10K::Module::Base.new("author/modname#{idx}", "/some/nonexistent/path/modname#{idx}", nil)
-      end
+      Array.new(4, R10K::Module::Base.new('author/modname', "/some/nonexistent/path/modname", nil))
```
mwaggett: looks like we might want to keep this? https://github.com/puppetlabs/r10k/pull/1044/commits/0e198cac586cf79af5ad1ce0b1305a3eb02846e1

mwaggett:
We should also revert the changelog entry: https://github.com/puppetlabs/r10k/pull/1044/commits/b3408940684bc92b3c75b22b861f3549cc7b5a1f

spec/unit/action/puppetfile/install_spec.rb
justinstoller: Updated

justinstoller:
The changelog entry doesn't revert, there's been additional changes since then formatting the changelog line. I've added a new commit removing it.

Magisus (dismissed):

mwaggett:
@justinstoller if you update the commit message, I will merge

justinstoller:
Updated the commit message and put it in the PR description.

mwaggett (approved):

Magisus (approved):

PR #1063, opened 2020-05-11
(maint) Switch Travis to Ubuntu 20.04 builder

Iristyle:
- Focal fossa time!
 - No need to upgrade Docker for buildkit as image has 19.03 already,
   so this should remove ~30s of setup time
 - Switch to Ruby 2.6.6 for specs, which is known good and installed
 - docker-compose in the 19.03 release included here is 1.23.1, so
   continue to upgrade it on Travis to 1.25.5
 - Separate out Travis operations so that logs have timings for lint, build and test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,14 @@ matrix:
         - docker
       rvm: 2.6.6
       env:
+        - DOCKER_COMPOSE_VERSION=1.25.5
         # necessary to prevent overwhelming TravisCI build output limits
         - DOCKER_BUILD_FLAGS="--progress plain"
+      before_install:
+        - sudo rm /usr/local/bin/docker-compose
+        - curl --location https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname --kernel-name`-`uname --machine` > docker-compose
+        - chmod +x docker-compose
+        - sudo mv docker-compose /usr/local/bin
       script:
         - set -e
         - cd docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ notifications:
   email: false
 sudo: false
 jdk:
-  - openjdk8
+  - openjdk11
 before_install: gem install bundler -v '< 2' --no-document
 matrix:
   include:
@@ -23,20 +23,16 @@ matrix:
     - stage: r10k tests
       rvm: jruby
     - stage: r10k container tests
-      dist: bionic
+      dist: focal
       language: ruby
       services:
-        # bionic uses 18.06 but need 19.03+ for buildkit so upgrade later in relevant cell
         - docker
-      rvm: 2.6.5
+      rvm: 2.6.6
       env:
         # necessary to prevent overwhelming TravisCI build output limits
         - DOCKER_BUILD_FLAGS="--progress plain"
       script:
         - set -e
-        - sudo apt update -y && sudo apt install -y docker.io
-        - sudo systemctl unmask docker.service
-        - sudo systemctl start docker
         - cd docker
         - make lint
         - make build
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,11 @@ matrix:
         # necessary to prevent overwhelming TravisCI build output limits
         - DOCKER_BUILD_FLAGS="--progress plain"
       script:
-        - |
-          set -ex
-          sudo apt update -y && sudo apt install -y docker.io
-          sudo systemctl unmask docker.service
-          sudo systemctl start docker
-          cd docker && make lint build test
-          set +x
+        - set -e
+        - sudo apt update -y && sudo apt install -y docker.io
+        - sudo systemctl unmask docker.service
+        - sudo systemctl start docker
+        - cd docker
+        - make lint
+        - make build
+        - make test


mwaggett (dismissed):

Gemfile
```diff
@@ -2,7 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 gemspec
 
 group :extra do
-  gem 'rugged', '>= 0.24.0', '< 0.24.6', :platforms => :ruby
+  gem 'rugged', '<= 0.28.5', :platforms => :ruby
```
Iristyle: @adrienthebo just noticing we're pinning to an ancient rugged version. At first glance, it doesn't look like it can compile on Ubuntu 20.04 anymore

I'm also noticing that we're installing the pre-compiled rugged binary for the Alpine-based container, which is pulling in 0.27.5-r0 - i.e see https://travis-ci.org/github/puppetlabs/r10k/jobs/680807669#L524 -- and if we're pinning to the older rugged for a particular reason, it makes me wonder if the container works right with Rugged.

You can also verify directly against latest published image like

> docker run -it --rm --entrypoint /usr/bin/ruby puppet/r10k -e "require 'rugged'; puts Rugged::VERSION"
```0.27.5```

And to make things a little more fun, I see that there's a 1.0 release out there as of April 3rd - https://rubygems.org/gems/rugged/versions

Comment (Iristyle) [reply Gemfile]: As a test, that bump in the gemspec to 0.28.5 compiles and passes tests just fine on Ubuntu 20.04. Will try another bump to 1.0 for ha-has.

File: .travis.yml [hunk 2]
```diff
@@ -23,20 +23,23 @@ matrix:
...
         # necessary to prevent overwhelming TravisCI build output limits
         - DOCKER_BUILD_FLAGS="--progress plain"
+      before_install:
+        - sudo rm /usr/local/bin/docker-compose
+        - curl --location https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname --kernel-name`-`uname --machine` > docker-compose
+        - chmod +x docker-compose
+        - sudo mv docker-compose /usr/local/bin
       script:
-        - |
-          set -ex
```
underscorgan, line 37: the `set -e` might be important here. Want to make sure travis will report a failure if one of these steps returns non-zero.

mwaggett:
Based on https://github.com/puppetlabs/r10k/commit/6c55978d7b5e2c4c57732f3bb4967aa6a31a9a69, I would say we are probably safe to bump rugged at this point (assuming things build and work correctly).

.travis.yml
```diff
@@ -23,20 +23,23 @@ matrix:
...
       rvm: jruby
     - stage: r10k container tests
-      dist: bionic
+      dist: focal
       language: ruby
       services:
-        # bionic uses 18.06 but need 19.03+ for buildkit so upgrade later in relevant cell
         - docker
-      rvm: 2.6.5
+      jdk: openjdk9
```
mwaggett: Is this the version we want? PE ships with java 11, so I feel like we should test against that?

Gemfile
```diff
@@ -2,7 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 gemspec
 
 group :extra do
-  gem 'rugged', '>= 0.24.0', '< 0.24.6', :platforms => :ruby
+  gem 'rugged', '~> 1.0.0', :platforms => :ruby
```
mwaggett: If this works, would we want to loosen it to `~>1.0` so we get new Y versions?

.travis.yml
Iristyle: It is important - didn't think it was because of how the steps were separated with `-`, but I was wrong. It's added back.

Thanks!

Gemfile
Iristyle: @scotje it looks like you were the last one to pin rugged back in https://github.com/puppetlabs/r10k/commit/6c55978d7b5e2c4c57732f3bb4967aa6a31a9a69

I think we probably want to go to a 1.0 release now if you're not aware of any other reasons to block that?

.travis.yml
Iristyle: It's not used in the container matrix cell at all...  I just needed something that was available on the `focal` builder, because `openjdk8` is not and will immediately cause the cell to fail. 

I tried setting it to `jdk: ` in an effort to prevent installation in that cell, but it didn't work.

Gemfile
Iristyle: Up to maintainers... 

I think we want the rugged change in a standalone PR too.

Gemfile
scotje: Based on the commit message it looks like there was just a specific issue with the `0.24.6` release that we were avoiding. Don't have any more recent context for the 1.x changes.

.travis.yml
mwaggett: Is `openjdk11` available? I just want to minimize confusion for future us, since 9 isn't a version we've every explicitly supported. Otherwise, could you add a comment that the version isn't used / is meaningless?

underscorgan (dismissed):

Gemfile
Iristyle: Cool, that's what @mwaggett thought and it what it looked like to me too. Thanks!

.travis.yml
Iristyle: Yeah, I can try to set that globally too. I was also minimizing the blast radius of my changes to just the docker cell, but updating everything to openjdk11 is probably a reasonable solution too.

Will update PR to try it out.

Gemfile
Iristyle: Separate PR is up that fully documents the Rugged dep bump - https://github.com/puppetlabs/r10k/pull/1064

underscorgan (approved):

.travis.yml
```diff
@@ -23,20 +23,23 @@ matrix:
...
         - docker
-      rvm: 2.6.5
+      rvm: 2.6.6
       env:
+        - DOCKER_COMPOSE_VERSION=1.25.5
         # necessary to prevent overwhelming TravisCI build output limits
         - DOCKER_BUILD_FLAGS="--progress plain"
+      before_install:
+        - sudo rm /usr/local/bin/docker-compose
+        - curl --location https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname --kernel-name`-`uname --machine` > docker-compose
```
mwaggett, line 37: What does this get run on? It seems like `uname` doesn't want to take long flags on OSX??

Iristyle: Only gets run on Ubuntu in Travis...

mwaggett: lol 🤦‍♀️

mwaggett (approved):

PR #1064, opened 2020-05-12
(maint) Upgrade Rugged gem to ~> 1.0

Iristyle:
This is a blocker for using Ubuntu 20.04 builders in Travis as part of #1063

 - The old version was pinned to a max of 0.24.5 in 2017 due to some
   build-time errors in 6c55978d7b5e2c4c57732f3bb4967aa6a31a9a69

   0.24.5 was published Jan 6, 2017 and has been updated regularly,
   receiving 350+ commits since:

   https://github.com/libgit2/rugged/compare/v0.24.5...v1.0.0

 - The gemspec doesn't specify a hard dependency on rugged, so updating
   the Gemfile only impacts local development and CI systems.

   Rugged 1.0 specifies compatibility back to Ruby 1.9.3:
	 rubygems.org/gems/rugged/versions/1.0.0

 - In reality, the Alpine "edge" (nightly) container and the latest
   tagged "release" (3.3.3) have been installing the Alpine 3.9
   precompiled binary to avoid compiling rugged. That vesion is actually
   0.27.5:

   docker run -it --rm --entrypoint /usr/bin/ruby puppet/r10k:edge -e "require 'rugged'; puts Rugged::VERSION"
   0.27.5

   docker run -it --rm --entrypoint /usr/bin/ruby puppet/r10k:3.3.3 -e "require 'rugged'; puts Rugged::VERSION"
   0.27.5

 - The shipped container won't pick up a new version of Rugged until
   we either:

   * compile it in a multi-stage build
   * update Alpine version to 3.11 or similar - 3.11 ships 0.28.4.1

 - In PE scenarios, the vanagon project specifies a dependency on rugged
   0.27.7, and this PR doesn't change that:

   puppetlabs/pe-r10k-vanagon:configs/components/rubygem-rugged.rb@master#L3

 - Also note that a newer version of Rugged is required to be able to
   build on Ubuntu 20.04 LTS Focal Fossa (and a PR is in progress to
   switch the Travis jobs over to this builder for the sake of a newer
   pre-installed Docker version)

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -3,6 +3,7 @@ CHANGELOG
 
 Unreleased
 ----
+- Upgrade Rugged Gemfile dependency for local development to ~> 1.0
 
 
 3.5.0
@@ -394,7 +395,7 @@ Puppetfile should be installed to. See the [Puppetfile documentation](https://gi
 
 You can now configure how r10k purges unmanaged content after a deployment. The
 default behavior should be unchanged but there is a new "purge\_levels" configuration
-option that can be used to enable new behavior or de-activate certain existing 
+option that can be used to enable new behavior or de-activate certain existing
 behaviors. See the relevant [configuration documentation](https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd#purge_levels) for more details.
 
 (RK-223) Ability to track control repo branch from content declarations.
@@ -441,7 +442,7 @@ fix.)
 Previously, r10k only supported the use of HTTP proxies for connecting to the Puppet
 Forge. With these changes, r10k can now be configured to use an HTTP proxy for both
 Forge and Git operations. Configuration can be specified globally, for Forge or Git
-only, or on a per-Git repository basis. See [configuration documentation](https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd) 
+only, or on a per-Git repository basis. See [configuration documentation](https://github.com/puppetlabs/r10k/blob/master/doc/dynamic-environments/configuration.mkd)
 for more details.
 
 ### Bug Fixes
@@ -462,9 +463,9 @@ for SSH.
 (RK-241) "deploy display" action does not properly format wrapped exceptions
 
 The "deploy display" action was not capturing and logging exceptions in the same way as
-other related actions. This meant that in many cases, when an error occurred, the 
+other related actions. This meant that in many cases, when an error occurred, the
 underlying cause was not being shown. Specifically, the "deploy display" action was
-not benefitting from the improved error messaging for unreadable SSH keys which was 
+not benefitting from the improved error messaging for unreadable SSH keys which was
 added in r10k 2.2.0 as part of RK-220.
 
 2.2.1
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 gemspec
 
 group :extra do
-  gem 'rugged', '>= 0.24.0', '< 0.24.6', :platforms => :ruby
+  gem 'rugged', '~> 1.0', :platforms => :ruby
 end
 
 group :development do


mwaggett:
oh wait, I just found this: https://github.com/puppetlabs/pe-r10k-vanagon/commit/065e62269d432287fa9ac3d33ecd0f5acb9027de

we'll have to investigate where that^ is still an issue

Iristyle:
Good find @mwaggett ... lets definitely validate that before moving forward with any upgrades.  Hopefully in all the time that's passed, those issues are no longer present.

adrienthebo:
I've been running with the following local patch to resolve issues with Ruby 2.5+ and Rugged:

```diff
diff --git i/Gemfile w/Gemfile
index 3cf7721..892d7ea 100644
--- i/Gemfile
+++ w/Gemfile
@@ -2,7 +2,11 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 gemspec
 
 group :extra do
-  gem 'rugged', '>= 0.24.0', '< 0.24.6', :platforms => :ruby
+  if RUBY_VERSION =~ /^2\.[567]/
+    gem 'rugged', '~> 1.0', :platforms => :ruby
+  else
+    gem 'rugged', '>= 0.24.0', '< 0.24.6', :platforms => :ruby
+  end
 end
 
 group :development do
```

This works for me, approved pending resolution of the pe-r10k-vanagon issue.

adrienthebo (dismissed):

justinstoller:
(I don't think that pe-r10k-vanagon reference is public)

fwiw, the issue is that libgit2 implemented their own proxy support so their error messages if something proxy-related is really bad and doesn't differentiate between different types of issues we specifically call out in testing (assumably because users have had a hard time debugging those issues).

justinstoller:
I don't know if Adrian is using a proxy in his setup?

I don't see any mention of proxies in the release notes for libgit2 post the change we're concerned about.

I also don't think our internal packaging cares about the :extra group in the Gemfile? I would think that Finch's patch would the correct solution for this.

adrienthebo:
> I don't know if Adrian is using a proxy in his setup?
> 
> I don't see any mention of proxies in the release notes for libgit2 post the change we're concerned about.
> 
> I also don't think our internal packaging cares about the :extra group in the Gemfile? I would think that Finch's patch would the correct solution for this.

I don't have visibility into pe-r10k-vanagon. Are we seeing CI failures for proxy tests?

Iristyle:
Since:

* Rugged 1.0 is marked as compatible with Ruby 1.9.3
* r10k doesn't take a hard dependency on rugged here in the gemspec
* the Gemfile doesn't impact what we actually ship

I'm inclined to the simpler pin to `~> 1.0` rather than choosing based on `RUBY_VERSION`

I will update the commit message real quick / repush... and I think we're safe to merge.

mwaggett:
@adrienthebo yeah, but this was from a year ago:

> Previously we had to move forward from our no-longer maintained versions of libgit2 & rugged so we could begin building packages on RHEL8 (an issue with ruby 2.5 and more recent llvms caused rugged to fail to compile pre-0.27.0.

> In rolling forward we moved to the latest release of both libgit2/rugged (0.28.1). However the 0.28 series has only been out for two months at this point and still may be working some kinks out. We specifically are seeing issues with their new re-written handing of proxies.

> We may need to review how we configure proxies in r10k to account for libgit2 0.28.0+ proxy handling, and that effort is unknown, at a minimum it in its current configuration it no longer gives any proxy related errors or warning on misconfiguration. That classifies at least as a regression in usability. Unfortunately, with r10k's current low priority rolling to a well supported, stable, and known good build is the preferred fix for now.

@justinstoller wrote that^ so he might have more info

justinstoller:
> I don't have visibility into pe-r10k-vanagon. Are we seeing CI failures for proxy tests?

We did in April of 2019, when we tried to upgrade to 0.28. I tried to summarize the content of the commit message being referenced but here it is in full:

```
(maint) roll back rugged/libgit2 versions

Previously we had to move forward from our no-longer maintained versions
of libgit2 & rugged so we could begin building packages on RHEL8 (an
issue with ruby 2.5 and more recent llvms caused rugged to fail to
compile pre-0.27.0.

In rolling forward we moved to the latest release of both libgit2/rugged
(0.28.1). However the 0.28 series has only been out for two months at
this point and still may be working some kinks out. We specifically are
seeing issues with their new re-written handing of proxies.

We may need to review how we configure proxies in r10k to account for
libgit2 0.28.0+ proxy handling, and that effort is unknown, at a minimum
it in its current configuration it no longer gives any proxy related
errors or warning on misconfiguration. That classifies at least as a
regression in usability. Unfortunately, with r10k's current low priority
rolling to a well supported, stable, and known good build is the preferred
fix for now.
```

The proxy tests live in this repo if you want to look at them. Re-reading my summary it might not have been well worded. libgit2 dropped libcurl as a dependency and wrote their own proxy handling code. IIRC, that code doesn't tell you that it attempted to use a proxy or what specific issue occurred. You just get a generic connection failure with no good way to debug.

Are you using a proxy with your setup, Finch? Have you had to do any debugging of it?

Anyways, that's info that would be nice to know for our own internal prioritization, but I don't think our internal packaging repo cares or should block a FOSS Gemfile change.

mwaggett:
I'm doing a quick build of pe-r10k-vanagon for sanity, but otherwise lgtm

mwaggett (approved):

Iristyle:
Think we're just waiting on a 👍 from @adrienthebo or @dhollinger to get this merged... and then we can get #1063 in

adrienthebo (approved):

PR #1066, opened 2020-05-19
(maint) Update the Forge API base url

binford2k:
The Forge API is now primarily served from the `puppet.com` domain. DNS
still resolves from the old domain, but the legacy `*.puppetlabs.com`
certificate [has been removed](https://headwayapp.co/puppet-forge-updates/removed-legacy-*-puppetlabs-com-ssl-certificate-143941).
For most users, this will have no effect because the main `puppet.com`
certificate includes `puppetlabs.com` as a SAN, but in some outdated or
otherwise broken SSL stacks, this might prevent SSL validation.

To remove that edge case, this PR updates the baseurl to the current
domain name.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,8 @@ CHANGELOG
 Unreleased
 ----
 
+- Change default API base url to forgeapi.puppet.com
+  [#1066](https://github.com/puppetlabs/r10k/pull/1066)
 
 3.5.0
 -----
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -163,7 +163,7 @@ interactions. See the global proxy setting documentation for more information an
 #### baseurl
 
 The 'baseurl' setting indicates where Forge modules should be installed from.
-This defaults to 'https://forgeapi.puppetlabs.com'
+This defaults to 'https://forgeapi.puppet.com'
 
 ```yaml
 forge:
--- a/integration/tests/basic_functionality/proxy_specified_in_configuration.rb
+++ b/integration/tests/basic_functionality/proxy_specified_in_configuration.rb
@@ -50,7 +50,7 @@ forge:
 CONF
 
 #Verification
-squid_log_regex = /CONNECT forgeapi.puppetlabs.com:443/
+squid_log_regex = /CONNECT forgeapi.puppet.com:443/
 
 #Teardown
 teardown do
--- a/integration/tests/basic_functionality/proxy_with_pe_only_module.rb
+++ b/integration/tests/basic_functionality/proxy_with_pe_only_module.rb
@@ -55,7 +55,7 @@ site_pp_path = File.join(git_environments_path, 'manifests', 'site.pp')
 site_pp = create_site_pp(master_certname, '  include peonly')
 
 #Verification
-squid_log_regex = /CONNECT forgeapi.puppetlabs.com:443/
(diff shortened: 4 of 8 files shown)


mwaggett:
@binford2k Is this still something we should do? We ended up needing to revert this change from the 2.6.x stream because it broke something in CI. Wondering if we should try again or just close this.

binford2k:
🤷‍♀️ 

At some point yes, we totally should. But it's clearly less of a problem than I thought it would be, so it's low priority.

CHANGELOG.mkd
```diff
@@ -7,7 +7,8 @@ Unreleased
 - Add 'strip\_component' environment source configuration setting, to allow deploying Git branches named like "env/production" as Puppet environments named like "production". [#1128](https://github.com/puppetlabs/r10k/pull/1128)
 - A warning will be emitted when the user supplies conflicting arguments to module definitions in a Puppetfile, such as when specifying both :commit and :branch [#1130](https://github.com/puppetlabs/r10k/pull/1130)
 - Add optional standard module and environment specification interface: name, type, source, version. These options can be used when specifying environments and/or modules in a yaml/exec source, as well as when specifying modules in a Puppetfile. Providing the standard interface simplifies integrations with external services [#1131](https://github.com/puppetlabs/r10k/pull/1131)
-
```
mwaggett, line 10: (nit) don't need this newline

mwaggett (commented):
so yeah I'm hoping to get a chance to try this again soon. I forget exactly which tests broke when we tried merging this into an older branch, but I'll poke around (it was something in PE tests).

mwaggett:
would be good to squash these commits though

binford2k:
@mwaggett is the CI breakage still a thing?

mwaggett:
@binford2k yes. hoping to get a chance to look into it soon.

mwaggett:
@binford2k Were you ever able to get Forge folks to look into the CI breakage? It's (at least some of) the integration tests you updated in this PR that fail when you change the url from `puppetlabs` to `puppet`.

caseywilliams:
Adding a bit of a summary here:

This breaks PE integration tests, we believe because of the use of [beaker-puppet's deprecated  `stub_forge_on` method](https://github.com/voxpupuli/beaker-puppet/blob/376a735afee0735dc7e7547310a060a738d6f5da/lib/beaker-puppet/helpers/puppet_helpers.rb#L652), which [was removed recently](https://github.com/voxpupuli/beaker-puppet/pull/160) - this won't be fixed in a future release of beaker-puppet. IMO, these tests should be either refactored or at least temporarily disabled if we want to get this change merged.

mwaggett:
*The [PR](https://github.com/voxpupuli/beaker-puppet/pull/160) to remove `stub_forge_on` has not been merged yet, fwiw

I'm extremely hesitant to "temporarily" disable because there's a 95% chance that that means "permanently". If we don't think those tests are providing value, I'm happy to get rid of them, but we should make that decision consciously. 
As for refactoring, were you just thinking of using the generic `stub_hosts_on` method or did you have something else in mind?

caseywilliams:
I don't have any strong opinions about how to solve this (and I'm not familiar with this codebase myself), so however the PE teams want to address this works for me. I asked around and I'm not aware of any plans to fully remove `forgeapi.puppetlabs.com` anytime soon, so whenever works for you is fine. `forgeapi.puppet.com` _is_ the official URL, though, so it should eventually become the default in r10k.

PR #1080, opened 2020-07-24
[#1058] Avoid concurrent cache updates when downloading modules on multiple threads

Magisus:
Moved from https://github.com/puppetlabs/r10k/pull/1059

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -143,7 +143,7 @@ class Puppetfile
     mod.origin = 'Puppetfile'
 
     @managed_content[install_path] << mod.name
-    cachedir = module_vcs_cachedir(mod)
+    cachedir = mod.cachedir
     @modules_by_vcs_cachedir[cachedir] ||= []
     @modules_by_vcs_cachedir[cachedir] << mod
     @modules << mod
@@ -188,19 +188,6 @@ class Puppetfile
   end
 
   private
-
-  def module_vcs_cachedir(mod)
-    if mod.respond_to? :repo
-      repo = mod.repo
-      if repo.respond_to? :cache
-        cache = repo.cache
-        if cache.respond_to? :sanitized_dirname
-          return cache.sanitized_dirname
-        end
-      end
-    end
-    :none
-  end
 
   def serial_accept(visitor)
     visitor.visit(:puppetfile, self) do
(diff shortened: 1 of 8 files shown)


Magisus:
This needs a rebase, but our branch protection rules prevent force-pushing. I'm going to re-open this from my fork.

lib/r10k/puppetfile.rb
```diff
@@ -212,15 +221,22 @@ def concurrent_accept(visitor, pool_size)
   def modules_queue(visitor)
     Queue.new.tap do |queue|
       visitor.visit(:puppetfile, self) do
-        modules.each { |mod| queue << mod }
+        modules_by_cachedir = modules_by_vcs_cachedir.clone
+        modules_without_vcs_cachedir = modules_by_cachedir.delete(:none) || []
+
+        modules_without_vcs_cachedir.each {|mod| queue << Array(mod) }
```
mwaggett, line 227: I'm a confused by this.. `modules_without_vcs_cachedir` could only be `:none` or `[]`, right? so why would those need to go in the queue?

Magisus: `:none` isn't going in the queue, `:none` is a key containing a list of all the modules that don't have a cachedir. They're added one-by-one to queue so that they can all get processed in parallel. The other keys in the maps are cachedir names, and their contents are lists of all module corresponding to a given cachedir. These lists are added to the queue, whole, so that the modules within them are processed by a single thread, serially.

mwaggett: oh right it's a map, not an array

Magisus:
I'm not sure why these tests are failing yet, they were added since the original PR. Looking into it.

lib/r10k/puppetfile.rb
mwaggett: so then wouldn't we want the values of `{none: [<blah>]}`, instead of the whole map? like we do for `modules_by_cachedir` below?

mwaggett: maybe I should remind myself how Ruby works

mwaggett: ignore me

Magisus:
Our branch protection doesn't allow force pushing to branches on this repo, so I've opened https://github.com/puppetlabs/r10k/pull/1081 from my fork.

PR #1081, opened 2020-07-27
[#1058] Avoid concurrent cache updates when downloading modules on multiple threads

Magisus:
Supersedes #1080

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -3,12 +3,10 @@ CHANGELOG
 
 Unreleased
 ----------
-## Changes
 
-(#823) Add filter_command configuration option for git repositories
-(Thanks to [mhumpula](https://github.com/mhumpula) for the [feature](https://github.com/puppetlabs/r10k/pull/823/conflicts).)
-
-----
+- Add filter_command configuration option for git repositories. (Thanks to [mhumpula](https://github.com/mhumpula) for the feature.) [#823](https://github.com/puppetlabs/r10k/pull/823)
+- Increase default pool_size to 4, allowing modules to be downloaded on 4 threads concurrently. [#1038](https://github.com/puppetlabs/r10k/issues/1038)
+- Ensure that modules that share a cachedir download serially, to avoid cache corruption. [#1058](https://github.com/puppetlabs/r10k/issues/1058)
 
 3.5.2
 -----
--- a/integration/tests/git_source/git_source_repeated_remote.rb
+++ b/integration/tests/git_source/git_source_repeated_remote.rb
@@ -1,0 +1,68 @@
+require 'git_utils'
+require 'r10k_utils'
+require 'master_manipulator'
+test_name 'Verify the same remote can be used in more than one object'
+
+env_path = on(master, puppet('config print environmentpath')).stdout.rstrip
+r10k_fqp = get_r10k_fqp(master)
+git_environments_path = '/root/environments'
+git_repo_path = '/git_repos'
+git_repo_name = 'environments'
+git_control_remote = File.join(git_repo_path, "#{git_repo_name}.git")
+code_dir = "#{env_path}/production"
+
+last_commit = git_last_commit(master, git_environments_path)
+git_provider = ENV['GIT_PROVIDER']
+r10k_config_path = get_r10k_config_file_path(master)
+r10k_config_bak_path = "#{r10k_config_path}.bak"
+#In-line files
+r10k_conf = <<-CONF
+cachedir: '/var/cache/r10k'
+git:
+  provider: '#{git_provider}'
+sources:
+  control:
+    basedir: "#{env_path}"
+    remote: "#{git_control_remote}"
+CONF
+
+# Install the same module in two different places
+puppetfile = <<-EOS
+mod 'prod_apache',
+  :git => 'git://github.com/puppetlabs/puppetlabs-apache.git',
+  :branch => 'master'
+
+mod 'test_apache',
+  :git => 'git://github.com/puppetlabs/puppetlabs-apache.git',
+  :branch => 'master'
+EOS
+
+teardown do
+  step 'Restore Original "r10k" Config'
+  on(master, "mv #{r10k_config_bak_path} #{r10k_config_path}")
+
+  clean_up_r10k(master, last_commit, git_environments_path)
+end
+
+step 'Stub the forge'
+stub_forge_on(master)
+
+step 'Backup Current "r10k" Config'
+on(master, "mv #{r10k_config_path} #{r10k_config_bak_path}")
+
+step 'Update the "r10k" Config'
+create_remote_file(master, r10k_config_path, r10k_conf)
+
+step 'Ask r10k to deploy'
+on(master, "#{r10k_fqp} deploy environment -p")
+
+step 'Add puppetfile with repeated remote'
+create_remote_file(master, "#{git_environments_path}/Puppetfile", puppetfile)
+git_add_commit_push(master, 'production', 'add Puppetfile', git_environments_path)
+
+step 'Deploy r10k'
+on(master, "#{r10k_fqp} deploy environment -p")
+
+step 'Verify module was installed in both places'
+on(master, "test -d #{code_dir}/modules/prod_apache")
+on(master, "test -d #{code_dir}/modules/test_apache")
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -143,7 +143,7 @@ class Puppetfile
     mod.origin = 'Puppetfile'
 
     @managed_content[install_path] << mod.name
-    cachedir = module_vcs_cachedir(mod)
+    cachedir = mod.cachedir
     @modules_by_vcs_cachedir[cachedir] ||= []
     @modules_by_vcs_cachedir[cachedir] << mod
     @modules << mod
@@ -188,19 +188,6 @@ class Puppetfile
   end
 
   private
-
-  def module_vcs_cachedir(mod)
-    if mod.respond_to? :repo
-      repo = mod.repo
-      if repo.respond_to? :cache
-        cache = repo.cache
-        if cache.respond_to? :sanitized_dirname
-          return cache.sanitized_dirname
-        end
-      end
-    end
-    :none
-  end
 
   def serial_accept(visitor)
     visitor.visit(:puppetfile, self) do
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -10,7 +10,7 @@ class Puppetfile
 
   include R10K::Settings::Mixin
 
-  def_setting_attr :pool_size, 1
+  def_setting_attr :pool_size, 4
 
   include R10K::Logging
 
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -42,6 +42,11 @@ class Puppetfile
   #   @return [Boolean] Overwrite any locally made changes
   attr_accessor :force
 
+  # @!attribute [r] modules_by_vcs_cachedir
+  #   @api private Only exposed for testing purposes
+  #   @return [Hash{:none, String => Array<R10K::Module>}]
+  attr_reader :modules_by_vcs_cachedir
+
   # @param [String] basedir
   # @param [String] moduledir The directory to install the modules, default to #{basedir}/modules
   # @param [String] puppetfile_path The path to the Puppetfile, default to #{basedir}/Puppetfile
@@ -58,6 +63,7 @@ class Puppetfile
 
     @modules = []
     @managed_content = {}
+    @modules_by_vcs_cachedir = {}
     @forge   = 'forgeapi.puppetlabs.com'
 
     @loaded = false
@@ -137,6 +143,9 @@ class Puppetfile
     mod.origin = 'Puppetfile'
 
     @managed_content[install_path] << mod.name
+    cachedir = module_vcs_cachedir(mod)
+    @modules_by_vcs_cachedir[cachedir] ||= []
+    @modules_by_vcs_cachedir[cachedir] << mod
     @modules << mod
   end
 
@@ -180,6 +189,19 @@ class Puppetfile
 
   private
 
+  def module_vcs_cachedir(mod)
+    if mod.respond_to? :repo
+      repo = mod.repo
+      if repo.respond_to? :cache
+        cache = repo.cache
+        if cache.respond_to? :sanitized_dirname
+          return cache.sanitized_dirname
+        end
+      end
+    end
+    :none
+  end
+
   def serial_accept(visitor)
     visitor.visit(:puppetfile, self) do
       modules.each do |mod|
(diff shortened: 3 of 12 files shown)


Magisus:
Still interested in thoughts on https://github.com/puppetlabs/r10k/pull/1059#discussion_r459774931.

CHANGELOG.mkd
```diff
@@ -5,8 +5,13 @@ Unreleased
...
-(Thanks to [mhumpula](https://github.com/mhumpula) for the [feature](https://github.com/puppetlabs/r10k/pull/823/conflicts).)
+- (#823) Add filter_command configuration option for git repositories
+(Thanks to [mhumpula](https://github.com/mhumpula) for the [feature](https://github.com/puppetlabs/r10k/pull/823).)
+- Increase default pool_size to 4, allowing modules to be downloaded on 4 threads concurrently. [#1038](https://github.com/puppetlabs/r10k/issues/1038)
+
+## Bug Fix
+
+- Ensure that modules that share a cachedir download serially, to avoid cache corruption. [#1058](https://github.com/puppetlabs/r10k/issues/1058)
 
 ----
```
mwaggett: while you're in here, would you mind removing this?

CHANGELOG.mkd
```diff
@@ -5,8 +5,13 @@ Unreleased
 ----------
 ## Changes
 
-(#823) Add filter_command configuration option for git repositories
-(Thanks to [mhumpula](https://github.com/mhumpula) for the [feature](https://github.com/puppetlabs/r10k/pull/823/conflicts).)
+- (#823) Add filter_command configuration option for git repositories
+(Thanks to [mhumpula](https://github.com/mhumpula) for the [feature](https://github.com/puppetlabs/r10k/pull/823).)
+- Increase default pool_size to 4, allowing modules to be downloaded on 4 threads concurrently. [#1038](https://github.com/puppetlabs/r10k/issues/1038)
+
+## Bug Fix
```
mwaggett: Looks like we haven't been categorizing our changes since 3.3.3.
I know the `## Changes` above was introduced by a community PR.. 
Do we want to go back to categorizing like this? If so, do we want to fix the last few versions or nah?

CHANGELOG.mkd
```diff
@@ -5,8 +5,13 @@ Unreleased
 ----------
```
mwaggett, line 5: oh yeah, and would you mind putting this back to just `----`?

mwaggett (dismissed):

CHANGELOG.mkd
Magisus: No let's not.

CHANGELOG.mkd
Magisus: When rendered, this makes a line just like the shorter version. When unrendered, I kinda like that it's the same length as the word 🤷‍♀️ But I can change it if you want.

mwaggett: oh interesting. then yeah, sure, we can keep it like this.

CHANGELOG.mkd
mwaggett: Let's not fix old versions or let's not re-start doing it this way at all?

Magisus: Not re-start doing it.

CHANGELOG.mkd
```diff
@@ -3,12 +3,11 @@ CHANGELOG
...
 Unreleased
 ----------
-## Changes
 
-(#823) Add filter_command configuration option for git repositories
-(Thanks to [mhumpula](https://github.com/mhumpula) for the [feature](https://github.com/puppetlabs/r10k/pull/823/conflicts).)
-
-----
+- (#823) Add filter_command configuration option for git repositories
+(Thanks to [mhumpula](https://github.com/mhumpula) for the [feature](https://github.com/puppetlabs/r10k/pull/823).)
```
mwaggett: (sorry I keep throwing this at your PR even though they're unrelated to your work, but) do we want to fix this so it better-matches how we've been formatting things. We seem to only mention tickets out front and we put issues and PRs on the end. ¯\_(ツ)_/¯

Magisus: If we're going to be prescriptive about this, seems only fair that the format be explicitly stated somewhere.

mwaggett: fair. I can add something to our PR template that has the changelog reminder.

mwaggett (dismissed):

CHANGELOG.mkd
mwaggett: https://github.com/puppetlabs/r10k/pull/1082

Magisus:
This does increase the default pool_size again, I think that's what we wanted?

Magisus:
There was a request way back when for an acceptance test that demonstrates this bug (and its fix) so I'm working on that now.

Magisus:
I verified that this test failed with a pool size greater than one, before this fix. I had forgotten, but I'll note here (as well as in the commit message) for posterity that the error only occurs in shellgit.

integration/tests/git_source/git_source_repeated_remote.rb
```diff
@@ -0,0 +1,65 @@
+require 'git_utils'
+require 'r10k_utils'
+require 'master_manipulator'
+test_name 'Verify the same remote can be used in more than one object'
+
+env_path = on(master, puppet('config print environmentpath')).stdout.rstrip
+r10k_fqp = get_r10k_fqp(master)
+git_environments_path = '/root/environments'
+git_repo_path = '/git_repos'
+git_repo_name = 'environments'
```
mwaggett, line 10: this is.. confusing.. can we call this something else, like 'control-repo' or something?

mwaggett: or is this just so that we don't have to configure (what file-sync calls) a submodules dir?

integration/tests/git_source/git_source_repeated_remote.rb
```diff
@@ -0,0 +1,65 @@
...
+CONF
+
+puppetfile = <<-EOS
+mod 'non_module_object_1',
+  :install_path => './',
+  :git => 'git://github.com/puppetlabs/control-repo.git',
+  :branch => 'production'
+
+mod 'non_module_object_2',
+  :install_path => '',
```
mwaggett: what is the behavior when `:install_path` is an empty string?
I'm assuming `:install_path` is where to install the module and having a different path than the one for `non_module_object_1` is how we ensure these are treated as separate modules?
Could they not just go in `./mod1` and `./mod2` or something?

integration/tests/git_source/git_source_repeated_remote.rb
```diff
@@ -0,0 +1,65 @@
...
+git:
+  provider: '#{git_provider}'
+sources:
+  control:
+    basedir: "#{env_path}"
+    remote: "#{git_control_remote}"
+CONF
+
+puppetfile = <<-EOS
+mod 'non_module_object_1',
```
mwaggett: what does it mean for this to be a 'non module object'?

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: These are specifically not modules. 

And yeah I imagine the install path could be anything. This test is copied almost exactly from the other one that failed, minus the assertions for purging. I didn't dig much into the choices made there (the commits for these test are notoriously bare-bones, and their tickets tend to be in a different tracking system than JIRA, so the spelunking is difficult), but I can.

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: The probably don't have to be for this test to work, I might change that.

But some useful links on that topic: https://tickets.puppetlabs.com/browse/CODEMGMT-649, https://puppet.com/docs/pe/2018.1/puppetfile.html#declare_git_repositories_in_the_puppetfile

integration/tests/git_source/git_source_repeated_remote.rb
mwaggett: I don't mind you leaving it, if that's the existing pattern we've established. just trying to understand how all things things work.

if they're not modules, what are they?

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: This corresponds to the control repo, hosted locally instead of somewhere like Github. We could call it something else, but this is the name used throughout in these tests, so I'm a little reluctant to change it just here.

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: Also, `install_path` appears to be relative to the location of the Puppetfile. So this is equivalent to the `./` in the previous example, which was likely intentional in the original test, which was verifying that things at the top level (i.e. not in the module dir of an environment) would get purged.

Magisus: In this specific test, they're whole control repos. But according to the docs (linked in the other comment) they're most commonly hiera data in a real install.

integration/tests/git_source/git_source_repeated_remote.rb
mwaggett: right. I'm fine leaving it if that's what we do elsewhere, I feel like it just conflates things in a way that is kind of confusing. but I assume that's it's being used so that paths look the way we expect them to, with an `environments` directory.

Magisus:
I'll update the test to be a slightly less-weird setup. There's no reason most of this can't be a more usual workflow and still demonstrate this bug.

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: The name of this path shouldn't have any bearing on where things get placed, since r10k is just treating this directory/git repo like a remote. It's likely called `environments` because ultimately a control repo is just the collection of environments, each of its branches corresponding to a different one. Likely this helped the original test writer feel less confused about what was happening, even if we've gotten super used to other names (like "control repo") now.

integration/tests/git_source/git_source_repeated_remote.rb
mwaggett: so would 'content that isn't a forge module' be more accurate? like, this could be anything that lives in a git repo, right? even what we call a "module"? sorry, this is just background stuff that I'm trying to understand.

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: This is what the control-repo fixture is called throughout these tests (an I think maybe in frankenbuilder too) so I left it.

mwaggett: hmm. then I like it even less lol. but it's fine, we can leave it for consistency.

Magisus: The `env_path` var above is what holds the location that content will be deployed to, assuming it doesn't have its `install_dir` overridden.

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: I'm going to change this to not muddy the waters in this test. But yeah, in this case, it's something that isn't structured like a Puppet module, which was a specific feature that got added. Previously, things had to look like modules for r10k to deploy them (I gather from skimming the ticket), whether they came from the Forge or from git. But now, the thing on the other end of a git source can be something else besides a Puppet module if the user wants.

integration/tests/git_source/git_source_repeated_remote.rb
mwaggett: ahh okay. so then for our purposes, we don't need to set `install_dir` at all, right?

Magisus: Right. And I'm changing that, because I'm updating this to install the same module under two different names, but both in the regular modulepath.

Magisus:
@mwaggett see if that's a little clearer. I left the source repo alone for consistency, like we discussed, but updated what's in the Puppetfile to be less unusual and added a couple of assertions to better illustrate the result of the deploy.

integration/tests/git_source/git_source_repeated_remote.rb
```diff
@@ -0,0 +1,68 @@
...
+require 'master_manipulator'
+test_name 'Verify the same remote can be used in more than one object'
+
+env_path = on(master, puppet('config print environmentpath')).stdout.rstrip
+r10k_fqp = get_r10k_fqp(master)
+git_environments_path = '/root/environments'
+git_repo_path = '/git_repos'
+git_repo_name = 'environments'
+git_control_remote = File.join(git_repo_path, "#{git_repo_name}.git")
+code_dir = '/etc/puppetlabs/code/environments/production'
```
mwaggett: Should the `env_path` be incorporated into this somehow? Is it `"#{env_path}/production"` or `"/etc/puppetlabs/code/#{env_path}/production"`? Looks like the former, based on https://puppet.com/docs/puppet/5.5/environments_creating.html#environmentpath .

Magisus: The `env_path` is the absolute path, so `/etc/puppetlabs/code/environments`. But yeah it can be incorporated.

integration/tests/git_source/git_source_repeated_remote.rb
```diff
@@ -0,0 +1,68 @@
...
+
+# Install the same module in two different places
+puppetfile = <<-EOS
+mod 'prod_apache',
+  :git => 'git://github.com/puppetlabs/puppetlabs-apache.git',
+  :branch => 'master'
+
+mod 'test_apache',
+  :git => 'git://github.com/puppetlabs/puppetlabs-apache.git',
+  :branch => 'master'
```
mwaggett, line 37: Would it be more realistic for this to be the 'test' branch?

Is it worth testing both ^the more realistic case and a case where the sources are literally identical? Or is that kind of redundant/unnecessary?

Magisus: That module only has a `master` and a `release` branch in its repo, and I'm not sure if release is permanent. I need to make sure it's something that this test is unlikely to fail on.

Magisus: Users in the r10k community were unable to come up with a _real_ use case that would trigger this bug. So no matter what we do here, it's not likely to match up to any known real workflow.

integration/tests/git_source/git_source_repeated_remote.rb
mwaggett: cool, yeah, I think if we include `env_path`, it's clearer that the `git_repo_name` has no relationship to that path.

integration/tests/git_source/git_source_repeated_remote.rb
mwaggett: gotcha, okay 👍

Magisus: I could find a different module that has more branches, or create a fixture, but we use `apache` elsewhere in these tests so I went with it.

mwaggett: if we don't think the two cases are different, I'm fine leaving it

integration/tests/git_source/git_source_repeated_remote.rb
Magisus: updated

mwaggett (approved):

Magisus:
@adrienthebo I'd appreciate a look at this if you have a minute, since the original concern with `WithModules` was yours, and Justin and I don't see an issue, but maybe we're missing something? https://github.com/puppetlabs/r10k/pull/1059#discussion_r459774931

lib/r10k/puppetfile.rb
```diff
@@ -212,15 +221,22 @@ def concurrent_accept(visitor, pool_size)
   def modules_queue(visitor)
     Queue.new.tap do |queue|
       visitor.visit(:puppetfile, self) do
-        modules.each { |mod| queue << mod }
+        modules_by_cachedir = modules_by_vcs_cachedir.clone
```
dhollinger, line 224: Am I missing something here or should `modules_by_vcs_cachedir` be an instance variable or is this appropriate since it's available as an `attr_reader`?

dhollinger (commented):
Just had one question, otherwise looks good to me

lib/r10k/puppetfile.rb
Magisus: Correct, this calls the `attr_reader` method for which the default implementation is to simply return the instance variable of the same name. We added the reader so we could retrieve that value in the tests. This could just as easily be `@modules_by_vcs_cachedir.clone`, since we're in the class where the variable is definied, but as I understand it, it's generally good practice to use the reader method if one is provided. Makes it easier to override with a custom getter later if you end up needing to do something special when retrieving the value.

dhollinger (approved):

PR #1090, opened 2020-09-23
RFC: Add `--environment` option to `puppetfile install`

alexjfisher:
I use Onceover and it uses R10k via `r10k puppetfile install`.
Unfortunately, this is a problem if modules in your Puppetfile use `:branch => :control_branch`.
See https://github.com/puppetlabs/r10k/blob/master/doc/puppetfile.mkd#control-repo-branch-tracking

```
Cannot track control repo branch for content some_module when not part of a 'deploy' action, will use default if available.
```

This change adds a `--environment` option to the `r10k puppetfile install` command.
Not sure I'm going about it the right way, or perhaps someone can suggest a better solution.

Feedback appreciated.

### DIFF
--- a/lib/r10k/module/git.rb
+++ b/lib/r10k/module/git.rb
@@ -108,6 +108,8 @@ class R10K::Module::Git < R10K::Module::Base
 
     if @desired_ref == :control_branch && @environment && @environment.respond_to?(:ref)
       @desired_ref = @environment.ref
+    elsif @desired_ref == :control_branch && @environment.is_a?(String)
+      @desired_ref = @environment
     end
   end
 end
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -52,12 +52,14 @@ class Puppetfile
   # @param [String] puppetfile_path The path to the Puppetfile, default to #{basedir}/Puppetfile
   # @param [String] puppetfile_name The name of the Puppetfile, default to 'Puppetfile'
   # @param [Boolean] force Shall we overwrite locally made changes?
-  def initialize(basedir, moduledir = nil, puppetfile_path = nil, puppetfile_name = nil, force = nil )
+  # @param [String] The environment the Puppetfile is in.
+  def initialize(basedir, moduledir = nil, puppetfile_path = nil, puppetfile_name = nil, force = nil, environment = nil )
     @basedir         = basedir
     @force           = force || false
     @moduledir       = moduledir  || File.join(basedir, 'modules')
     @puppetfile_name = puppetfile_name || 'Puppetfile'
     @puppetfile_path = puppetfile_path || File.join(basedir, @puppetfile_name)
+    @environment     = environment
 
     logger.info _("Using Puppetfile '%{puppetfile}'") % {puppetfile: @puppetfile_path}
 
--- a/lib/r10k/action/puppetfile/install.rb
+++ b/lib/r10k/action/puppetfile/install.rb
@@ -10,7 +10,7 @@ module R10K
 
         def call
           @visit_ok = true
-          pf = R10K::Puppetfile.new(@root, @moduledir, @puppetfile, nil , @force)
+          pf = R10K::Puppetfile.new(@root, @moduledir, @puppetfile, nil , @force, @environment)
           pf.accept(self)
           @visit_ok
         end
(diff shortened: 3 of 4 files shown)


alexjfisher:
Obviously, if (after I've fixed it up, added tests, docs etc), this is accepted, I'd need to make a small change to onceover too.
@dylanratcliffe Any thoughts?

lib/r10k/puppetfile.rb
```diff
@@ -52,12 +52,14 @@ class Puppetfile
...
   # @param [Boolean] force Shall we overwrite locally made changes?
-  def initialize(basedir, moduledir = nil, puppetfile_path = nil, puppetfile_name = nil, force = nil )
+  # @param [String] The environment the Puppetfile is in.
+  def initialize(basedir, moduledir = nil, puppetfile_path = nil, puppetfile_name = nil, force = nil, environment = nil )
     @basedir         = basedir
     @force           = force || false
     @moduledir       = moduledir  || File.join(basedir, 'modules')
     @puppetfile_name = puppetfile_name || 'Puppetfile'
     @puppetfile_path = puppetfile_path || File.join(basedir, @puppetfile_name)
+    @environment     = environment
```
alexjfisher, line 62: The use of `@environment` is what I'm not sure about.  There's already an `@environment` attr_accessor in this class.   I don't *think* it's being used anywhere, and it's expecting an `R10k::Environment` where I'm just wanting to use a plain string.

lib/r10k/module/git.rb
```diff
@@ -108,6 +108,8 @@ def parse_options(options)
 
     if @desired_ref == :control_branch && @environment && @environment.respond_to?(:ref)
       @desired_ref = @environment.ref
+    elsif @desired_red == :control_branch && @environment.is_a?(String)
```
dylanratcliffe: Typo here

```suggestion
    elsif @desired_ref == :control_branch && @environment.is_a?(String)
```

lib/r10k/puppetfile.rb
dylanratcliffe: Yeah I have no idea how/if that could even have worked before...

dylanratcliffe (commented):
Overall I'd say this functionality should exist. Would also be useful for testing and things like that. Possibly I'd argue we should call it "control_branch" or something rather than environment since environment is such an overloaded term...

We would need to test that `attr_accessor` on an older version and see what on earth it does, and if nothing then we'd be fine with this implementation.

Is this functionality just for Onceover's sake? Because Onceover does do nasty hacks if you're using this in your Puppetfile (it literally edits the puppetfile on the fly). Does that not work?

lib/r10k/module/git.rb
alexjfisher: Oops.  Somewhere between testing and removing random debug/pry stuff I must have accidentally mutated an `f` into `d`...

alexjfisher:
> Is this functionality just for Onceover's sake? Because Onceover does do nasty hacks if you're using this in your Puppetfile (it literally edits the puppetfile on the fly). Does that not work?

Yes, my use-case was just Onceover.  If you're telling me that onceover is already supposed to have rewritten the Puppetfile, then I guess it's not working.  I'll try to debug this.  (If I couldn't get r10k to do what I wanted, I had considered mangling my Puppetfile in the jenkins job I've setup.)

justinstoller:
I don't have opinions on once over or whether this is the approach vs updating that library. However, should this approach be preferred I would like to second Dylan's comment 'we should call it "control_branch" or something rather than environment', for the same reasons.

alexjfisher:
> > Is this functionality just for Onceover's sake? Because Onceover does do nasty hacks if you're using this in your Puppetfile (it literally edits the puppetfile on the fly). Does that not work?
> 
> Yes, my use-case was just Onceover. If you're telling me that onceover is already supposed to have rewritten the Puppetfile, then I guess it's not working. I'll try to debug this. (If I couldn't get r10k to do what I wanted, I had considered mangling my Puppetfile in the jenkins job I've setup.)

@dylanratcliffe Did https://github.com/dylanratcliffe/onceover/pull/144 ever work??
Even in 3.3.0, (when I think the feature was first added), r10k is run against the actual Puppetfile, not the copy??
https://github.com/dylanratcliffe/onceover/blob/v3.3.0/lib/onceover/testconfig.rb#L237

alexjfisher:
I think I've figured out what went wrong... https://github.com/dylanratcliffe/onceover/issues/282

mwaggett:
It looks like this pull request hasn't seen any activity in several months. @alexjfisher Is this something that you're still planning to work on? We're happy to keep this pull request open if there's forward movement, and we'll leave it open for another two weeks from now. If there isn't any activity by then we may have to close this pull request due to lack of activity.

Closing the pull request wouldn't mean we don't consider this change valuable. However we try to keep all open pull requests moving towards completion, and closing stale pull requests help us focus on more active pull requests.

If you have any questions or concerns, please don't hesitate to ping us in the [Puppet Community Slack](http://slack.puppet.com/).

mwaggett:
Since this hasn't seen any activity in many months, I'm going to go ahead and close this pull request for the time being. Please feel free to re-open it, or open a fresh PR, if you have a renewed interest in this work. Thanks for contributing!

PR #1091, opened 2020-10-10
Docker: Added option to include your own scripts

dvaerum:
In the [The Puppet Server container](https://github.com/puppetlabs/puppetserver/blob/3b6d9181bc79671220ad938a193d938e95f0a70e/docker/puppetserver/docker-entrypoint.sh#L13-L19) there is the option to mount your own script to config to do some specific configuration for Puppet Server.

It finds it very useful to have this feature for r10k, so I hope you will accept this pull request 😃 I also added the option to have ones own script executed as root.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
+- Added the option to mount your on custom script into the r10k (Docker) container, just like you can with [the puppetserver container](https://github.com/puppetlabs/puppetserver/blob/3b6d9181bc79671220ad938a193d938e95f0a70e/docker/puppetserver/docker-entrypoint.sh#L13-L19) [#1091](https://github.com/puppetlabs/r10k/pull/1091)
 - Add 'strip\_component' environment source configuration setting, to allow deploying Git branches named like "env/production" as Puppet environments named like "production". [#1128](https://github.com/puppetlabs/r10k/pull/1128)
 - A warning will be emitted when the user supplies conflicting arguments to module definitions in a Puppetfile, such as when specifying both :commit and :branch [#1130](https://github.com/puppetlabs/r10k/pull/1130)
 - Add optional standard module and environment specification interface: name, type, source, version. These options can be used when specifying environments and/or modules in a yaml/exec source, as well as when specifying modules in a Puppetfile. Providing the standard interface simplifies integrations with external services [#1131](https://github.com/puppetlabs/r10k/pull/1131)
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,7 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
-- Added the option to mount your on custom script into the r10k (Docker) container, just like you can with [the puppetserver container](https://github.com/puppetlabs/puppetserver/blob/3b6d9181bc79671220ad938a193d938e95f0a70e/docker/puppetserver/docker-entrypoint.sh#L13-L19) [#1091](https://github.com/puppetlabs/r10k/pull/1091)
+- Added the option to mount your own custom script into the r10k (Docker) container, just like you can with [the puppetserver container](https://github.com/puppetlabs/puppetserver/blob/3b6d9181bc79671220ad938a193d938e95f0a70e/docker/puppetserver/docker-entrypoint.sh#L13-L19) [#1091](https://github.com/puppetlabs/r10k/pull/1091)
 - Add 'strip\_component' environment source configuration setting, to allow deploying Git branches named like "env/production" as Puppet environments named like "production". [#1128](https://github.com/puppetlabs/r10k/pull/1128)
 - A warning will be emitted when the user supplies conflicting arguments to module definitions in a Puppetfile, such as when specifying both :commit and :branch [#1130](https://github.com/puppetlabs/r10k/pull/1130)
 - Add optional standard module and environment specification interface: name, type, source, version. These options can be used when specifying environments and/or modules in a yaml/exec source, as well as when specifying modules in a Puppetfile. Providing the standard interface simplifies integrations with external services [#1131](https://github.com/puppetlabs/r10k/pull/1131)
--- a/docker/README.md
+++ b/docker/README.md
@@ -10,7 +10,7 @@ The following environment variables are supported:
 
   Set to 'true' to enable Google Analytics metrics. Defaults to 'false'.
 
-If you want to use your own script to config the r10k container, you can mount your script to the directory `/docker-custom-entrypoint.d` in the container:
+If you want to use your own script to configure the r10k container, you can mount your script to the directory `/docker-custom-entrypoint.d` in the container:
 ```
 docker run --name r10k -v ./r10k-custom:/docker-custom-entrypoint.d puppet/r10k
 ```
--- a/docker/README.md
+++ b/docker/README.md
@@ -10,6 +10,11 @@ The following environment variables are supported:
 
   Set to 'true' to enable Google Analytics metrics. Defaults to 'false'.
 
+If you want to use your own script to config the r10k container, you can mount your script to the directory `/docker-custom-entrypoint.d` in the container:
+```
+docker run --name r10k -v ./r10k-custom:/docker-custom-entrypoint.d puppet/r10k
+```
+
 ## Analytics Data Collection
 
 The r10k container collects usage data. This is disabled by default. You can enable it by passing `--env PUPPERWARE_ANALYTICS_ENABLED=true`
--- a/docker/r10k/Dockerfile
+++ b/docker/r10k/Dockerfile
@@ -67,6 +67,7 @@ RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
     mv "$supercronic" "/usr/local/bin/${supercronic}" && \
     ln -s "/usr/local/bin/${supercronic}" /usr/local/bin/supercronic
 
+# hadolint ignore=DL3000
 WORKDIR "/home/${PP_USER}"
 
 COPY docker/r10k/Dockerfile /
--- a/docker/r10k/Dockerfile
+++ b/docker/r10k/Dockerfile
@@ -52,11 +52,10 @@ COPY --from=build /workspace/r10k.gem /
 SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 # ignore apk and gem pinning
 # hadolint ignore=DL3018,DL3028
-RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
+RUN chmod -R a+x /adduser.sh /docker-entrypoint.sh /docker-entrypoint.d && \
 # Add a puppet user to run r10k as for consistency with puppetserver
     /adduser.sh && \
     chown -R "${PP_USER}:" /docker-entrypoint.d /docker-entrypoint.sh && \
-    chmod -R +x            /docker-entrypoint.d /docker-entrypoint.sh && \
     mkdir -p /docker-custom-entrypoint.d && \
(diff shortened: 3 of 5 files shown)


docker/r10k/Dockerfile
```diff
@@ -35,6 +35,9 @@ LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
 COPY docker/r10k/adduser.sh docker/r10k/docker-entrypoint.sh /
 COPY docker/r10k/docker-entrypoint.d /docker-entrypoint.d
 
+ARG PP_USER
```
dhollinger: This should probably have a default value of `ARG PP_USER=puppet` so that a default user is defined.

docker/r10k/release.Dockerfile
```diff
@@ -25,10 +25,13 @@ LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
 COPY adduser.sh docker-entrypoint.sh /
 COPY docker-entrypoint.d /docker-entrypoint.d
 
+ARG PP_USER
```
dhollinger: Default value needed. Format `ARG PP_USER=puppet`

dhollinger (changes requested):
Looks good to me, but the Dockerfiles should have a default for their ARG value

dvaerum:
> Looks good to me, but the Dockerfiles should have a default for their ARG value

That has now been done 😉

dhollinger (dismissed):
LGTM

docker/r10k/Dockerfile
```diff
@@ -62,7 +67,6 @@ RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
     mv "$supercronic" "/usr/local/bin/${supercronic}" && \
     ln -s "/usr/local/bin/${supercronic}" /usr/local/bin/supercronic
 
-USER puppet
-WORKDIR /home/puppet
```
dhollinger, line 66: Add comment above this with content `# hadolint ignore=DL3000` to prevent a travis failure on Dockerfile linting.

docker/r10k/release.Dockerfile
```diff
@@ -49,7 +54,6 @@ RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
     mv "$supercronic" "/usr/local/bin/${supercronic}" && \
     ln -s "/usr/local/bin/${supercronic}" /usr/local/bin/supercronic
 
-USER puppet
-WORKDIR /home/puppet
```
dhollinger, line 53: Add comment above this with content # hadolint ignore=DL3000 to prevent a travis failure on Dockerfile linting.

dhollinger (changes requested):
Probably should also have the travis build lint all the Dockerfiles and not just the main dockerfile.

docker/r10k/release.Dockerfile
dvaerum: Done, and it seems to have worked :D

docker/r10k/Dockerfile
dvaerum: Done

dvaerum:
@dhollinger 
Will you merge this pull request? 🥺  If there is nothing else I should change/improve 😃

mwaggett:
@puppetlabs/pupperware could y'all take a look at this?

CHANGELOG.mkd
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
+- Added the option to mount your on custom script into the r10k (Docker) container, just like you can with [the puppetserver container](https://github.com/puppetlabs/puppetserver/blob/3b6d9181bc79671220ad938a193d938e95f0a70e/docker/puppetserver/docker-entrypoint.sh#L13-L19) [#1091](https://github.com/puppetlabs/r10k/pull/1091)
```
```suggestion
- Added the option to mount your own custom script into the r10k (Docker) container, just like you can with [the puppetserver container](https://github.com/puppetlabs/puppetserver/blob/3b6d9181bc79671220ad938a193d938e95f0a70e/docker/puppetserver/docker-entrypoint.sh#L13-L19) [#1091](https://github.com/puppetlabs/r10k/pull/1091)
```

docker/README.md
```diff
@@ -10,6 +10,11 @@ The following environment variables are supported:
 
   Set to 'true' to enable Google Analytics metrics. Defaults to 'false'.
 
+If you want to use your own script to config the r10k container, you can mount your script to the directory `/docker-custom-entrypoint.d` in the container:
```
mwaggett:
```suggestion
If you want to use your own script to configure the r10k container, you can mount your script to the directory `/docker-custom-entrypoint.d` in the container:
```

docker/r10k/Dockerfile
```diff
@@ -52,7 +55,9 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
 # Add a puppet user to run r10k as for consistency with puppetserver
     /adduser.sh && \
-    chown -R puppet: /docker-entrypoint.d /docker-entrypoint.sh && \
+    chown -R "${PP_USER}:" /docker-entrypoint.d /docker-entrypoint.sh && \
+    chmod -R +x            /docker-entrypoint.d /docker-entrypoint.sh && \
```
mwaggett: looks like there is a `chmod` call as the first line of this command, this should be consolidated with that.

docker/r10k/Dockerfile
```diff
@@ -52,7 +55,9 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
...
+    mkdir -p /docker-custom-entrypoint.d && \
```
mwaggett, line 59: Does this directory also need to be `chmod`ed and `chown`ed?

docker/r10k/Dockerfile
```diff
@@ -62,7 +67,7 @@ RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
     mv "$supercronic" "/usr/local/bin/${supercronic}" && \
     ln -s "/usr/local/bin/${supercronic}" /usr/local/bin/supercronic
 
-USER puppet
-WORKDIR /home/puppet
+# hadolint ignore=DL3000
```
mwaggett, line 71: Would you mind adding a comment explaining that this is an expected exclusion when using environment variables in `WORKDIR`, per https://github.com/hadolint/hadolint/wiki/DL3000 ?

docker/r10k/docker-entrypoint.sh
```diff
@@ -1,11 +1,15 @@
 #! /bin/sh
 
-set -e
+set -ue
 
 for f in /docker-entrypoint.d/*.sh; do
   # Don't print out any messages here since this is a CLI container
-  chmod +x "$f"
-  "$f"
+  su "${PP_USER}" -- "${f}"
```
mwaggett, line 7: I don't think you need to `su` to run this? since we run `chmod a+x` on the docker-entrypoint dir.

docker/r10k/docker-entrypoint.sh
```diff
@@ -1,11 +1,15 @@
...
 done
 
-exec /usr/bin/r10k "$@"
+for f in /docker-custom-entrypoint.d/*.sh; do
+  # Don't print out any messages here since this is a CLI container
+  [ -x "${f}" ] && "${f}"
+done
+
+exec su "${PP_USER}" -c "/usr/bin/r10k $*"
```
mwaggett, line 15: Why did this change from `exec /usr/bin/r10k "$@"`?

docker/r10k/release.Dockerfile
```diff
@@ -40,7 +43,9 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 # hadolint ignore=DL3018,DL3028
 RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
```
mwaggett: Comments I made on the other `Dockerfile` also apply to this one.

docker/r10k/Dockerfile
dvaerum: Okay, I have done this in both Dockerfile & release.Dockerfile

docker/r10k/Dockerfile
dvaerum: I have added the description for the exclusion 😃 
```
# Ignore the check for absolute WORKDIR path, this is need when using environment variables in WORKDIR.
# https://github.com/hadolint/hadolint/wiki/DL3000
```

docker/r10k/release.Dockerfile
dvaerum: Changes are applied to both files 👍

docker/r10k/docker-entrypoint.sh
dvaerum: This is done because I have removed
```Dockerfile
USER puppet
```
from Dockerfile & release.Dockerfile

To make sure that the scripts are executed by the `${PP_USER}` (aka puppet) user I had to use `su` because when `USER` in Dockerfile is not specified `root` is used. The reason for being the `root` user is, so that the scripts in `/docker-custom-entrypoint.d` is run as root because that is something I need.

docker/r10k/docker-entrypoint.sh
dvaerum: This is done because I have removed
```Dockerfile
USER puppet
```
from Dockerfile & release.Dockerfile

To make sure that `r10k` is executed by the `${PP_USER}` (aka puppet) user I am using `su` because when `USER` in Dockerfile is not specified `root` is used. The reason for being the `root` user is, so that the scripts in `/docker-custom-entrypoint.d` is run as root because that is something I need.

docker/r10k/Dockerfile
dvaerum: No, because the folder was meant to be a mount-point and is therefore empty, so I don't really see the reason for that.
The general idea was to allow people to mount a folder on the host with scripts there would be executed in the r10k container

dvaerum:
@mwaggett
I have made a new commit there should comply all your request, and answered all your question (hopefully satisfying).
If there is anything else please ask/comment

mwaggett:
@Iristyle any chance you could look at this soon?

Iristyle:
Thanks for the contribution!

I think the idea of executing scripts is fine, but changing the `USER` behavior inside the container is not necessary / orthogonal to the functionality of running user scripts. Furthermore, the use of binaries like `su` and `chmod` won't work properly in Kubernetes with appropriate security contexts configured.

So I think we need a slightly different approach to the problem that doesn't rely on permissions / user changes. Puppetserver doesn't have that behavior FWIW - https://github.com/puppetlabs/puppetserver/blob/6.x/docker/puppetserver/docker-entrypoint.sh#L11-L17

docker/r10k/Dockerfile
```diff
@@ -49,10 +52,11 @@ COPY --from=build /workspace/r10k.gem /
 SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 # ignore apk and gem pinning
 # hadolint ignore=DL3018,DL3028
-RUN chmod a+x /adduser.sh /docker-entrypoint.sh && \
+RUN chmod -R a+x /adduser.sh /docker-entrypoint.sh /docker-entrypoint.d && \
 # Add a puppet user to run r10k as for consistency with puppetserver
     /adduser.sh && \
-    chown -R puppet: /docker-entrypoint.d /docker-entrypoint.sh && \
+    chown -R "${PP_USER}:" /docker-entrypoint.d /docker-entrypoint.sh && \
```
Iristyle, line 58: Note that permissions here are being modified at *build time*... but then later consumed at runtime in the entrypoint script. So if a user ever set `PP_USER` to something other than `puppet`, you wouldn't get the behavior you're aiming for.

PR #1098, opened 2020-11-02
(maint) Update test regexes to work with Puppet 6 and 7

Magisus:
Puppet 7 stopped squashing errors from pluginsync while doing an agent
run. So the runs in these tests that error because of a missing
environment now error in a different spot with a different but related
message. This commit updates the test regexes to work in both cases.

```
```

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -3,6 +3,7 @@ CHANGELOG
 
 Unreleased
 ----------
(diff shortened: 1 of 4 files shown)


Magisus:
I ran this in both main and master, and saw both pass.

mwaggett (approved):

donoghuc (approved):

PR #1107, opened 2020-12-02
Implement less restrictive env module merge

reidmv:
By default, allow a module defined in an environment to override a copy
defined in a Puppetfile, rather than hard-failing the deploy.

This allows for a more frictionless path to using environment modules.

This commit changes default behavior of an experimental feature.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -5,6 +5,7 @@ Unreleased
 ----------
 
 - Fetch modules from the repo's default branch, rather than defaulting to 'master' [#1096](https://github.com/puppetlabs/r10k/issues/1096)
+- Enable configurable behavior of environment module / Puppetfile module conflicts (experimental features)
 
 3.7.0
 -----
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -5,7 +5,7 @@ Unreleased
 ----------
 
 - Fetch modules from the repo's default branch, rather than defaulting to 'master' [#1096](https://github.com/puppetlabs/r10k/issues/1096)
-- Enable configurable behavior of environment module / Puppetfile module conflicts (experimental features)
+- Experimental feature change: conflicts between environment-defined modules and Puppetfile-defined modules now default to logging a warning and deploying the environment module version, overriding the Puppetfile. Previously, conflicts would result in an error. The behavior is now configurable via the `module_conflicts` environment setting [#1107](https://github.com/puppetlabs/r10k/pull/1107)
 
 3.7.0
 -----
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -679,8 +679,8 @@ When a module is defined in an environment and also in a Puppetfile, the default
 
 Available `module_conflicts` options:
 
-* `override_puppetfile_and_warn` (default): the version of the module defined by the enviornment will be used, and the version defined in the Puppetfile will be ignored. A warning will be printed.
-* `override_puppetfile`: the version of the module defined by the enviornment will be used, and the version defined in the Puppetfile will be ignored.
+* `override_puppetfile_and_warn` (default): the version of the module defined by the environment will be used, and the version defined in the Puppetfile will be ignored. A warning will be printed.
+* `override_puppetfile`: the version of the module defined by the environment will be used, and the version defined in the Puppetfile will be ignored.
 * `error`: an error will be raised alerting the user to the conflict. The environment will not be deployed.
 
 ```yaml
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -673,6 +673,31 @@ modules:
     ref: 62d07f2
 ```
 
+#### Puppetfile module conflicts
+
+When a module is defined in an environment and also in a Puppetfile, the default behavior is for the environment definition of the module to take precedence, a warning to be logged, and the Puppetfile definition to be ignored. The behavior is configurable to optionally skip the warning, or allow a hard failure instead. Use the `module_conflicts` option in an environment definition to control this.
+
+Available `module_conflicts` options:
+
+* `override_puppetfile_and_warn` (default): the version of the module defined by the enviornment will be used, and the version defined in the Puppetfile will be ignored. A warning will be printed.
+* `override_puppetfile`: the version of the module defined by the enviornment will be used, and the version defined in the Puppetfile will be ignored.
+* `error`: an error will be raised alerting the user to the conflict. The environment will not be deployed.
+
+```yaml
+# production.yaml
+---
+type: git
+remote: git@github.com:puppetlabs/control-repo.git
+ref: 8820892
+module_conflicts: override_puppetfile_and_warn
+modules:
+  puppetlabs-stdlib: 6.0.0
+  puppetlabs-concat: 6.1.0
+  reidmv-xampl:
+    git: https://github.com/reidmv/reidmv-xampl.git
+    ref: 62d07f2
+```
+
 ### Bare Environment Type
 
 A "control repository" typically contains a hiera.yaml, an environment.conf, a manifests/site.pp file, and a few other things. However, none of these are strictly necessary for an environment to be functional if modules can be deployed to it.
--- a/lib/r10k/environment/with_modules.rb
+++ b/lib/r10k/environment/with_modules.rb
@@ -103,20 +103,21 @@ class R10K::Environment::WithModules < R10K::Environment::Base
                 .select { |_, v| v.size > 1 }
                 .map(&:first)
     unless conflicts.empty?
+      conflicts_str = conflicts.join(', ')
+      log_msg = _('Environment and Puppetfile both define the following modules, Puppetfile ' \
+                  'definition will be ignored: %{mods}' % { mods: conflicts_str })
       case conflict_opt = @options[:module_conflicts]
       when 'override_puppetfile'
-        logger.debug _('Environment and Puppetfile both define the following modules, Puppetfile ' \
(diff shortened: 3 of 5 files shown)


dhollinger (dismissed):
Good work. I like it!

doc/dynamic-environments/configuration.mkd
```diff
@@ -673,6 +673,31 @@ modules:
     ref: 62d07f2
 ```
 
+#### Puppetfile module conflicts
+
+When a module is defined in an environment and also in a Puppetfile, the default behavior is for the environment definition of the module to take precedence, a warning to be logged, and the Puppetfile definition to be ignored. The behavior is configurable to optionally skip the warning, or allow a hard failure instead. Use the `module_conflicts` option in an environment definition to control this.
+
+Available `module_conflicts` options:
+
+* `override_puppetfile_and_warn` (default): the version of the module defined by the enviornment will be used, and the version defined in the Puppetfile will be ignored. A warning will be printed.
```
mwaggett:
```suggestion
* `override_puppetfile_and_warn` (default): the version of the module defined by the environment will be used, and the version defined in the Puppetfile will be ignored. A warning will be printed.
```

doc/dynamic-environments/configuration.mkd
```diff
@@ -673,6 +673,31 @@ modules:
...
+* `override_puppetfile`: the version of the module defined by the enviornment will be used, and the version defined in the Puppetfile will be ignored.
```
mwaggett:
```suggestion
* `override_puppetfile`: the version of the module defined by the environment will be used, and the version defined in the Puppetfile will be ignored.
```

lib/r10k/environment/with_modules.rb
```diff
@@ -101,10 +103,21 @@ def validate_no_module_conflicts
...
+      when 'override_puppetfile_and_warn', nil
+        logger.warn  _('Environment and Puppetfile both define the following modules, Puppetfile ' \
+                       'definition will be ignored: %{mods}' % { mods: conflicts.join(', ') })
+      when 'error'
+        raise R10K::Error, _('Puppetfile cannot contain module names defined by environment ' \
+                             '%{name}; Remove the conflicting definitions of the following modules: ' \
+                             '%{mods}' % { name: self.name, mods: conflicts.join(', ') })
+      else
+        raise R10K::Error, _('Unexpected value for `module_conflicts` setting in %{env} ' \
+                             'environment: %{val}' % {env: 'name', val: conflict_opt})
```
mwaggett: Should this be `self.name` rather than 'name'?

Also, in the previous messages, we use `name` rather than `env` to denote the name of the environment. I think we should be consistent.

lib/r10k/environment/with_modules.rb
```diff
@@ -101,10 +103,21 @@ def validate_no_module_conflicts
...
                 .map(&:first)
     unless conflicts.empty?
-      msg = _('Puppetfile cannot contain module names defined by environment %{name}') % {name: self.name}
-      msg += ' '
-      msg += _("Remove the conflicting definitions of the following modules: %{conflicts}" % { conflicts: conflicts.join(' ') })
-      raise R10K::Error.new(msg)
+      case conflict_opt = @options[:module_conflicts]
+      when 'override_puppetfile'
+        logger.debug _('Environment and Puppetfile both define the following modules, Puppetfile ' \
+                       'definition will be ignored: %{mods}' % { mods: conflicts.join(', ') })
```
mwaggett: Should we pull this message into a variable, since we use it twice?
and/or pull out `conflicts.join(', ')`, since we use it three times?

spec/unit/environment/with_modules_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        mod = double('duplicate-stdlib', :name => 'stdlib')
+        expect(subject.puppetfile).to receive(:load)
+        expect(subject.puppetfile).to receive(:modules).and_return [mod]
+        expect { subject.validate_no_module_conflicts }.not_to raise_error
+      end
+    end
+
+    context "with module conflicts and 'error' behavior" do
+      let(:subject_params) {{ :module_conflicts => 'error' }}
+      it "does not raise an error" do
```
mwaggett:
```suggestion
      it "raises an error" do
```

spec/unit/environment/with_modules_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        expect { subject.validate_no_module_conflicts }.not_to raise_error
+      end
+    end
+
+    context "with module conflicts and default behavior" do
+      it "does not raise an error" do
+        mod = double('duplicate-stdlib', :name => 'stdlib')
+        expect(subject.puppetfile).to receive(:load)
+        expect(subject.puppetfile).to receive(:modules).and_return [mod]
+        expect { subject.validate_no_module_conflicts }.not_to raise_error
```
mwaggett, line 40: Should we also test the logging, like we do for the 'override_puppetfile' test below?

spec/unit/environment/with_modules_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        mod = double('module', :name => 'nonconflict')
+        expect(subject.puppetfile).to receive(:load)
+        expect(subject.puppetfile).to receive(:modules).and_return [mod]
+        expect { subject.validate_no_module_conflicts }.not_to raise_error
+      end
+    end
+
+    context "with module conflicts and default behavior" do
+      it "does not raise an error" do
+        mod = double('duplicate-stdlib', :name => 'stdlib')
```
mwaggett, line 36: Is there some magic that strips off the `puppetlabs-` part of the name? Otherwise, I'm not sure how this would be a duplicate of the environment version.

CHANGELOG.mkd
```diff
@@ -5,6 +5,7 @@ Unreleased
 ----------
 
 - Fetch modules from the repo's default branch, rather than defaulting to 'master' [#1096](https://github.com/puppetlabs/r10k/issues/1096)
+- Enable configurable behavior of environment module / Puppetfile module conflicts (experimental features)
```
mwaggett: I think the changelog entry should mention the change in default behavior and the setting where the behavior can be configured.

reidmv:
```suggestion
- Experimental feature change: conflicts between environment-defined modules and Puppetfile-defined modules now default to logging a warning and deploying the environment module version, overriding the Puppetfile. Previously, conflicts would result in an error. The behavior is now configurable via the `module_conflicts` environment setting [#1107](https://github.com/puppetlabs/r10k/pull/1107)
```

mwaggett: 💯  thank you!

spec/unit/environment/with_modules_spec.rb
reidmv: Yeah, there's some magic. `name` != what the user put in the Puppetfile as the title.
https://github.com/puppetlabs/r10k/blob/master/lib/r10k/module/base.rb#L51

mwaggett: 👍🏼  thanks

spec/unit/environment/with_modules_spec.rb
reidmv, line 40:
```suggestion
        expect(subject.logger).to receive(:warn).with(/Puppetfile.*both define.*ignored/i)
        expect { subject.validate_no_module_conflicts }.not_to raise_error
```

lib/r10k/environment/with_modules.rb
reidmv: Ah, yeah, that'd be a good maintainability improvement.

lib/r10k/environment/with_modules.rb
reidmv: Good catch. Standardizing on %{env}.

spec/unit/environment/with_modules_spec.rb
reidmv: Probably. I'll add it in.

reidmv:
@mwaggett Thanks so much for the detailed review! I think I've implemented all the suggestions. Ready for re-review.

lib/r10k/environment/with_modules.rb
```diff
@@ -101,10 +103,22 @@ def validate_no_module_conflicts
...
-      msg += ' '
-      msg += _("Remove the conflicting definitions of the following modules: %{conflicts}" % { conflicts: conflicts.join(' ') })
-      raise R10K::Error.new(msg)
+      conflicts_str = conflicts.join(', ')
+      log_msg = _('Environment and Puppetfile both define the following modules, Puppetfile ' \
+                  'definition will be ignored: %{mods}' % { mods: conflicts_str })
+      case conflict_opt = @options[:module_conflicts]
+      when 'override_puppetfile'
+        logger.debug log_msg
+      when 'override_puppetfile_and_warn', nil
```
mwaggett, line 112: I didn't notice this before, but what's the `nil` here for?

mwaggett (commented):
Looks great! Just one quick question.

lib/r10k/environment/with_modules.rb
reidmv: If the setting isn't specified, the option is nil. Default behavior. I don't _think_ r10k has a nice way of using schemas or other systems for this kind of thing? Could be wrong. Any suggestions?

mwaggett: ahhhh, gotcha. I think it's fine, I'm just a little out of practice at reading ruby haha 😅 
I had skimmed and thought that the first case was the default, so wasn't sure what this was doing here in the middle, but I see now.

mwaggett (approved):

mwaggett:
@dhollinger Would you mind taking another look, since there have been changes since your review?

dhollinger (approved):
LGTM

PR #1108, opened 2020-12-03
(maint) dockerfile mkdir -p

sbeaulie:
without this if the directory already exists, docker build will fail

```
```

### DIFF
--- a/docker/r10k/Dockerfile
+++ b/docker/r10k/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:${alpine_version} as build
 
 # hadolint ignore=DL3018
 RUN apk add --no-cache ruby git && \
-    mkdir /workspace
+    mkdir -p /workspace
 WORKDIR /workspace
 COPY . /workspace
 RUN gem build r10k.gemspec && \


docker/r10k/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM alpine:${alpine_version} as build
 
 # hadolint ignore=DL3018
 RUN apk add --no-cache ruby git && \
-    mkdir /workspace
+    mkdir -p /workspace
 WORKDIR /workspace
 COPY . /workspace
```
nwolfe, line 8: The change itself is fine, but wondering if it'd just move the problem elsewhere.

Since if `/workspace` already exists, what happens when we `COPY . /workspace`? 

Will we be left in the correct state? Will existing things in `/workspace` be removed or overwritten? What do we expect/want to happen regarding that 🤷

dhollinger (approved):

dhollinger (dismissed):

docker/r10k/Dockerfile
Iristyle: Really peculiar to require a `-p` for a directory only one level deep, isn't it?

sbeaulie: Yes this was part of my spike on kaniko and it would fail building because `mkdir` returns non-0 exit code if the directory already exists. `-p` is indeed for creating missing subdirectories, but in this case @Iristyle it's effectively to ignore existing directories. Otherwise the RUN command is unsuccessful and the build aborts.
I have opened an issue with kaniko https://github.com/GoogleContainerTools/kaniko/issues/1508
With regards to this PR, I think it is safe enough. I've had scripts where we would do `mkdir -p /tmp` to cover cases where the `/tmp` might or might not exist already in the base container. Alternatively, we could use a directory named something else.

mwaggett:
@sbeaulie Are you still hoping to get this merged? /cc @Iristyle @nwolfe

sbeaulie:
yes

docker/r10k/Dockerfile
Iristyle: I see the kaniko maintainers still haven't responded. This whole situation strikes me as an odd one.

I don't think we actually intend to use Kaniko, do we?

mwaggett: Would it make sense to just `rm -rf /workspace` before attempting to create the directory, if we already expect to start in a state where there is no `/workspace` directory?

sbeaulie: I was thining if the upstream image you start with in the FROM section already has a /workspace, even empty, this RUN command would fail. The safest is to do a `mkdir -p` instead of a delete then create. This is just something I found while using this file, and fixed by adding the `-p` flag and wanted not to forget about it, so I pushed it. Feel free to close, since it's not really useful.

Issue #1110, opened 2020-12-11
Require a ref, branch, tag, or commit when specifying a module from a git repo in the Puppetfile 


npwalker:
## Use Case
When a user does not specify a ref, branch, tag, or commit on a module coming from a git repo in the Puppetfile it automatically deploys from the master branch.  

There was a PR up to change that to automatically deploy from the default branch of the git server.  https://github.com/puppetlabs/r10k/pull/1097

Both of these are somewhat magical behaviors that IMO are more likely to cause unexpected results.  

## Describe the Solution You Would Like
Instead the user should be required to specify a ref, branch, tag, or commit and if one is not provided then parsing of the Puppetfile should fail and nothing deployed.  

## Describe Alternatives You've Considered
I've considered the PR above and do not prefer it because it makes it easy for users to get a result that may not be expected.  When the default branch of a git repo changes it can change the catalog applied to agents.  Users should knowingly opt-in to that behavior.  

## Additional Context
The primary reason the above PR was brought up was due to changing from master branch to main across the industry.  Users relying on this default behavior should knowingly opt-in to it by setting `ref => HEAD` in their Puppetfile instead of it just happening when no ref or branch is specified.  

Fwiw, the PR cannot be brought in without a major version because it is a breaking change in behavior.  If we are going to make a breaking change I propose we make a change that improves clarity.

trevor-vaughan:
Honestly, if I'm deploying from a git repo, I expect to get the default branch by default because that's how git works.

If you do this, then I would make a keyword like `__default__` to pull the default branch whatever it happens to be named at the moment.

Also, you could make it opt-in to make it non-breaking.

logicminds:
I would add a feature to Rak10e and other tooling to check for this behavior and warn the user that a default branch is NOT being used.  This won't work in all cases since many don't even know they can use tooling to validate Puppetfile.  

Additionally, if you change r10k you also have to change .fixture.yml too.  We don't want users asking "why do I do this here but not there.

I don't think this feature is worth the time and effort.  R10k has been working like this for 5+ years and the majority of us have never complained.

Now if r10k is hard coded to 'master' that would need to be changed.

dhollinger:
Why not take an extra step and add an option in the r10k.yaml file that can be used to configure an option to use the default branch if none is given? 

This prevents the magic from happening unless a user has specifically configured it to do so in the r10k configuration file.

mwaggett:
[...]

PR #1111, opened 2020-12-11
Revert "Issue 1096: use Default-Branch"

mwaggett:
This reverts commit 57329b6.
This constituted a breaking change, which we are not prepared to make right now.
In the future, when we are ready for a new major version, we hope to require
users to specify a ref in their Puppetfile, rather than defaulting to something.
See #1110 for more info.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,7 +4,6 @@ CHANGELOG
 Unreleased
 ----------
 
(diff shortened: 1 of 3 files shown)


mwaggett:
Hi @cocker-cc !! Thanks so much for your contribution, but unfortunately we have to revert, as it constitutes a breaking change. We recommend setting the ref (branch, SHA, etc.) explicitly in your Puppetfile, rather than relying on this default behavior.

npwalker (approved):

Magisus (approved):

dhollinger (approved):

Issue #1109, opened 2020-12-04
allow r10k to deploy with an arbitrary ssh key


tvpartytonight:
## Use Case
Currently, ssh credentials are configured in the r10k.yaml and use for authentication; it would also be useful to specify a key per deploy, so that r10k doesn't require having the key written to disk.
 
## Describe the Solution You Would Like
R10k accepts a deploy arg/env-var for the ssh key used.

This is similar to #1106.

Magisus:
If we decide to implement this for the shellgit provider, here's a post describing how you can pass SSH keys to `git clone` without messing with the system SSH config: https://stackoverflow.com/questions/6688655/select-private-key-to-use-with-git/41947805#41947805

Magisus:
If we decide to implement this for the shellgit provider, here is a post about how SSH keys can be passed directly to git operations without editing the system SSH config: https://stackoverflow.com/questions/6688655/select-private-key-to-use-with-git/41947805#41947805

mwaggett:
resolved by https://github.com/puppetlabs/r10k/pull/1116

Issue #1106, opened 2020-12-02
add support for using oauth tokens for authentication


tvpartytonight:
## Use Case
R10k supports both sshkeys and username/passwords embedded in a URL; however, there is no explicit support for using oauth tokens for authentication, as described in [here](https://github.blog/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/). R10k should take a command line arg for an oauth token and use it for authentication; it should not be read from disk or be leaked in a process list.

## Describe the Solution You Would Like
The rugged support is simple enough:

```

# code snippet to generate repo with oauth token
--
creds = Rugged::Credentials::UserPassword.new password: "super-cool-oauth-token", username: 'x-oauth-token'
Rugged::Repository.clone_at('https://github.com/my-cool-org/my-private-repo', Dir.mktmpdir, credentials: creds)
```

## Describe Alternatives You've Considered
Puppet itself is largely interested in rugged support, but shellgit shouldn't be too hard as well, though I haven't really tested it out...

Magisus:
I did some research, and to avoid leaking the secret, I think we should implement this as a flag that takes a file argument. We can then have that read `/dev/stdin` (and implement the `-` convention if we think that'd be good UX), and pass the token on stdin.

I think it might be nice if you could use the same flag for passing either tokens or SSH keys this way (see https://github.com/puppetlabs/r10k/issues/1109). The type of credential required to be passed can be inferred from the form of the URL, so we can give good input validation and nice error messages, without an explosion of new CLI flags.

Doing this as a file arg also gives users the option to use an actual file (rather than stdin) if they want, though for our usecase in PE, we want to avoid writing anything to disk.

Magisus:
Update, we're actually going to focus on implementing this for shellgit first, which is the default here in FOSS r10k.

Magisus:
Change of plans, it turns out it is nontrivial to support OAuth tokens in shellgit, because it expects them to be passed on stdin, which isn't possible with the way we currently shell out to git. Because this work is time-sensitive, we are going to do Rugged support first after all.

mwaggett:
resolved by https://github.com/puppetlabs/r10k/pull/1116

PR #1113, opened 2020-12-18
[#1106, #1109] Add CLI flags for sshkey and token paths

Magisus:
This PR adds two CLI flags to the deploy command, `sshkey-path`
and `token-path`. These will be used to pass credentials to the git
provider for use when cloning the repo, overriding values from the
config for the source.

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -189,6 +189,8 @@ module R10K
                       'generate-types': :self,
                       'puppet-path': :self,
                       'puppet-conf': :self,
+                      'sshkey-path': :self,
+                      'token-path': :self,
                       'default-branch-override': :self)
         end
       end
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -5,6 +5,7 @@ require 'r10k/action/visitor'
 require 'r10k/action/base'
 require 'r10k/action/deploy/deploy_helpers'
 require 'json'
+require 'tempfile'
 
 module R10K
   module Action
@@ -32,7 +33,8 @@ module R10K
           @visit_ok = true
 
           expect_config!
-          deployment = R10K::Deployment.new(@settings)
+          credentials = extract_credentials!
+          deployment = R10K::Deployment.new(@settings, credentials)
           check_write_lock!(@settings)
 
           deployment.accept(self)
@@ -42,6 +44,29 @@ module R10K
         include R10K::Action::Visitor
 
         private
+
+        def extract_credentials!
+          if @sshkey_path && @token_path
+            raise R10K::Error, "Cannot specify both an SSH key and a token to use with this deploy."
+          end
+
+          if @sshkey_path
+            if File.exists?(@sshkey_path)
+              { sshkey_path: @sshkey_path }
+            else
+              raise R10K::Error, _("{%path} does not exist, cannot load SSH key") % { path: @sshkey_path }
+            end
+          elsif @token_path
+            if @token_path == '-'
+              token = $stdin.read
+            elsif File.exists?(@token_path)
+              token = File.read(@token_path)
+            else
+              raise R10K::Error, _("{%path} does not exist, cannot load OAuth token") % { path: @token_path }
+            end
+            { token: token }
+          end
+        end
 
         def visit_deployment(deployment)
           # Ensure that everything can be preloaded. If we cannot preload all
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -5,7 +5,6 @@ require 'r10k/action/visitor'
 require 'r10k/action/base'
 require 'r10k/action/deploy/deploy_helpers'
 require 'json'
-require 'tempfile'
 
 module R10K
   module Action
@@ -60,12 +59,23 @@ module R10K
             if @token_path == '-'
               token = $stdin.read
             elsif File.exists?(@token_path)
-              token = File.read(@token_path)
+              token = File.read(@token_path).strip
             else
               raise R10K::Error, _("{%path} does not exist, cannot load OAuth token") % { path: @token_path }
             end
+
+            unless valid_token?(token)
+              raise R10K::Error, _("Supplied token contains invalid characters.")
+            end
+
             { token: token }
           end
+        end
+
+        # This regex is the only real requirement for OAuth token format,
+        # per https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/
+        def valid_token?(token)
+          return token =~ /^[\w\-\.~\+\/]+$/
         end
 
         def visit_deployment(deployment)
--- a/spec/unit/action/deploy/environment_spec.rb
+++ b/spec/unit/action/deploy/environment_spec.rb
@@ -168,49 +168,66 @@ describe R10K::Action::Deploy::Environment do
           end
         end
       end
+    end
 
-      describe 'extracting credentials' do
-        let(:deployment) do
-          R10K::Deployment.new(mock_config)
-        end
+    describe 'extracting credentials' do
+      let(:deployment) do
+        R10K::Deployment.new(mock_config)
+      end
 
-        it 'errors if both token and key paths are passed' do
-          action = described_class.new({ config: '/some/nonexistent/path',
-                                        'token-path': '/nonexistent',
-                                        'sshkey-path': '/also/fake' }, [])
-          expect{ action.call }.to raise_error(R10K::Error, /Cannot specify both/)
-        end
+      it 'errors if both token and key paths are passed' do
+        action = described_class.new({ config: '/some/nonexistent/path',
+                                      'token-path': '/nonexistent',
+                                      'sshkey-path': '/also/fake' }, [])
+        expect{ action.call }.to raise_error(R10K::Error, /Cannot specify both/)
+      end
 
-        it 'errors if sshkey file does not exist' do
-          action = described_class.new({ config: '/some/nonexistent/path',
-                                        'sshkey-path': '/also/fake' }, [])
-          expect{ action.call }.to raise_error(R10K::Error, /cannot load SSH key/)
+      it 'errors if sshkey file does not exist' do
+        action = described_class.new({ config: '/some/nonexistent/path',
+                                      'sshkey-path': '/also/fake' }, [])
+        expect{ action.call }.to raise_error(R10K::Error, /cannot load SSH key/)
 
-        end
+      end
 
-        it 'errors if token file does not exist' do
-          action = described_class.new({ config: '/some/nonexistent/path',
-                                        'token-path': '/also/fake' }, [])
-          expect{ action.call }.to raise_error(R10K::Error, /cannot load OAuth token/)
-        end
+      it 'errors if token file does not exist' do
+        action = described_class.new({ config: '/some/nonexistent/path',
+                                      'token-path': '/also/fake' }, [])
+        expect{ action.call }.to raise_error(R10K::Error, /cannot load OAuth token/)
+      end
 
-        it 'passes token to deployment from file' do
-          token_file = Tempfile.new('token')
-          token_file.write('my_token')
-          token_file.close
-          action = described_class.new({ config: '/some/nonexistent/path',
-                                         'token-path': token_file.path }, [])
-          expect(R10K::Deployment).to receive(:new).with({}, { token: "my_token" }).and_return(deployment)
-          action.call
-        end
+      it 'passes token to deployment from file' do
+        token_file = Tempfile.new('token')
+        token_file.write('my_token')
+        token_file.close
+        action = described_class.new({ config: '/some/nonexistent/path',
+                                       'token-path': token_file.path }, [])
+        expect(R10K::Deployment).to receive(:new).with({}, { token: "my_token" }).and_return(deployment)
+        action.call
+      end
 
-        it 'passes token to deployment from stdin' do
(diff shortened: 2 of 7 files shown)


lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +44,30 @@ def call
...
+              # TODO if this is reading from stdin, write the key to a (password
+              # protected?) tempfile instead and save the path for passing down
+              @settings[:sshkey] = File.read(path)
+            end
+          elsif @token_path
+            path = convert_dash_to_stdin(@sshkey_path)
+            if File.exists?(path)
+              @settings[:token] = File.read(path)
+            end
+          end
```
Magisus, line 72: Do we want to do validation of the file contents at this level?

Magisus:
I'll add more tests for the errors once I know this is generally what we want.

lib/r10k/action/deploy/environment.rb
```diff
@@ -32,6 +32,7 @@ def call
           @visit_ok = true
 
           expect_config!
+          extract_credentials
```
justinstoller: I think this should have an exclamation point since it raises.

lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +44,34 @@ def call
...
+            if File.exists?(path)
+              @settings[:token] = File.read(path)
+            else
+              raise R10K::Error, _("{%path} does not exist, cannot load OAuth token") % { path: path }
+            end
+          end
+        end
+
+        def convert_dash_to_stdin(path)
+          path == '-' ? '/dev/stdin' : path
```
justinstoller: Do we need to support Windows?? I don't think that `/dev/stdin` works on Windows, however I do think that reading the IO object `STDIN`/`$stdin` would be a portable way to handle that.

lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +44,34 @@ def call
 
         private
 
+        def extract_credentials
+          if @sshkey_path && @token_path
```
justinstoller, line 48: It's _soooo_ cool that these instance variables are defined and managed in a different file....

Magisus: In a super obtuse way too. Took me a while to figure out how I thought it worked, and even then I wasn't confident till I wrote the tests >.>

lib/r10k/action/deploy/environment.rb
Magisus: I think we probably do.

Magisus:
Turns out the `settings` hash is actually frozen, so I'll be making this less gross 🙂

lib/r10k/action/deploy/environment.rb
mwaggett: What sort of validation were you thinking? There's a certain point at which only actually attempting to clone a repo can confirm that the credentials are good, but catching other stuff sooner could be good.

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -340,6 +411,24 @@
         expect(subject.instance_variable_get(:@puppet_conf)).to eq('/nonexistent')
       end
     end
+
+    describe 'with sshkey_path' do
```
mwaggett: (nit) should this be `sshkey-path`?

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -340,6 +411,24 @@
...
+
+      subject { described_class.new({ config: '/some/nonexistent/path', 'sshkey-path': '/nonexistent' }, []) }
+
+      it 'sets puppet_path' do
```
mwaggett: should this be `sshkey_path`?

spec/unit/action/deploy/module_spec.rb
```diff
@@ -149,4 +157,22 @@
       expect(subject.instance_variable_get(:@cachedir)).to eq('/nonexistent')
     end
   end
+
+  describe 'with sshkey-path' do
+
+    subject { described_class.new({ config: '/some/nonexistent/path', 'sshkey-path': '/nonexistent' }, []) }
+
+    it 'sets puppet_path' do
```
mwaggett:
```suggestion
    it 'sets sshkey_path' do
```

lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +45,41 @@ def call
...
+            end
+          elsif @token_path
+            if @token_path == '-'
+              token = $stdin.read
+            elsif File.exists?(@token_path)
+              token = File.read(@token_path)
+            else
+              raise R10K::Error, _("{%path} does not exist, cannot load OAuth token") % { path: @token_path }
+            end
+            { token: token }
```
mwaggett, line 71: I assume the choice to send the token itself here, whereas we send the file path for the sshkey, is related to how we'll ultimately use them with rugged?

Magisus: Yes. Rugged expects the token as a string, and a file path for the SSH key.

lib/r10k/action/deploy/environment.rb
Magisus: I'll see if there is a spec for OAuth token format that we can rely on.

mwaggett: For reference, can I see where you found what you did?

mwaggett (dismissed):

lib/r10k/action/deploy/environment.rb
Magisus: https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/ Apparently it's super open-ended besides those character restrictions.

mwaggett: thanks!

Magisus:
This stuff still needs to be added to module deploy, but I think that warrants a refactor as part of it, so we're not just copy-pasting pretty much everything in this PR. I think that should be a separate piece of work.

mwaggett:
👍🏼  I would find it easier to review as a separate PR

mwaggett (dismissed):

lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +45,50 @@ def call
...
+            if @token_path == '-'
+              token = $stdin.read
+            elsif File.exists?(@token_path)
+              token = File.read(@token_path).strip
+            else
+              raise R10K::Error, _("{%path} does not exist, cannot load OAuth token") % { path: @token_path }
+            end
+
+            unless valid_token?(token)
+              raise R10K::Error, _("Supplied token is not a valid OAuth token.")
```
tvpartytonight: Nit: this is a little bit vague, because it could mean the the token is expired, too long, too short, etc. I think it should probably say something invalid characters found in the token, not that the token itself is invalid.

Magisus: Updated, lemme know what you think.

mwaggett (dismissed):

lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +45,50 @@ def call
...
+        def extract_credentials!
+          if @sshkey_path && @token_path
+            raise R10K::Error, "Cannot specify both an SSH key and a token to use with this deploy."
+          end
+
+          if @sshkey_path
+            if @sshkey_path == '-'
+              sshkey = $stdin.read
+              keypath = write_key_to_tmpfile(sshkey).path
+              { sshkey_file: keypath }
```
tvpartytonight: nit: is there a reason to switch from `@sshkey_path` to `sshkey_file`? I would just assume it should be called `sshkey_path` for consistency, but there could be reasons I'm not thinking of? I guess `sshkey_path.path` is a little hard on the 👀 ...

Magisus: Sure I can change it. I didn't have a strong reason.

lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +45,50 @@ def call
...
+          end
+        end
+
+        def write_key_to_tmpfile(sshkey)
+          keypath = Tempfile.new('sshkey')
+          File.chmod(0600, keypath)
+          keypath.write(sshkey)
+          keypath.close
+          keypath
+        end
```
tvpartytonight: I am vaguely worried about leaving behind keys in the Tmpfile dir, and there are some recommendations from ruby-doc about making this [more secure](https://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html#method-i-unlink-label-Unlink-before-close). 

Posting the relevant quote here:

> On POSIX systems, it’s possible to unlink a file right after creating it, and before closing it. This removes the filesystem entry without closing the file handle, so it ensures that only the processes that already had the file handle open can access the file’s contents. It’s strongly recommended that you do this if you do not want any other processes to be able to read from or write to the Tempfile, and you do not need to know the Tempfile’s filename either.

I guess the catch is Windows isn't super great about this either...but, we don't have PE on windows...yet...

Magisus: I think we do need to know the name though. The thing Rugged saves and loads is a file name...

I think the ultimate solution here is to use a regular file that is encrypted with a password and has locked down perms, and then clean it up by hand when we know we're done.

Magisus: I'm not sure if there's an interim thing we could be doing that might be more secure though. Thoughts, @justinstoller @pcarlisle ?

justinstoller: The file should be created in a directory that is locked down and owned by the process (e.g. `~/.ssh`). I assume somewhere in `/opt/puppetlabs` will work for PE and mostlikely is the process user's homedir (if it exists) and/or the configured datadir.

I'm not sure about unlinking before closing, we will be the same process as rugged right? But not if we provide support for shellgit at some point?

As a real handwaving thing, I would look to what `puppet` uses as that code base has learned (the hard way) how to secure tmp files. [FileSystem impl](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/file_system/file_impl.rb#L168-L183) & [FileSystem::UniqueFile](https://github.com/puppetlabs/puppet/blob/cd7e16542c919e516dda44dbf6bac0ff64dacf0a/lib/puppet/file_system/uniquefile.rb#L10).

justinstoller: Also.... If we're writing files now, we can do that in code-manager where we know the datadir and just give the path to r10k. Right?

Magisus: Yeah, that will work for our usecase, but we probably want to make sure we're still doing something sensible for FOSS users, I think? Does `~/.ssh` seem like the right call for the general case?

Magisus: Or wait. Are you saying we should just only accept a filepath (and not stdin) for this? Cause that works for me too, and that puts the onus on the user to secure the key properly, which makes sense to me.

tvpartytonight: I'm a little reticent to put items in `.ssh` that are ephemeral by nature and could get left behind. Do we need our own cool `.r10k-ssh` dir? 🤷

Magisus: I'm also totally fine changing this to only accept a file path (since that's what we need anyway) and not messing with this at all.

justinstoller: I'm thinking, the "-" stdin bit was good when our explicit goal was to not write anything to disk.

But, iiuc, we've landed on writing to disk is unavoidable.

Then for our use case we should write the secrets to disk where it makes the most sense for us. When writing to secrets to disk we should write them to a secure directory owned by the process (a directory locked down like ~/.ssh normally is). For us, I think that is our datadir in code-manager (and if not we can make a sub-directory and manage its perms).

Figuring out that where-to-write-secrets bit sounds like it would be hard from r10k (haven't done a lot of looking though, maybe I'm wrong?) and seems like we might as well write the secret to disk in the application calling r10k since it is explicitly configured with the process owned datadir.

Magisus: That works for me. I'll update this. I'll leave `-` as an option for the token though, since there we really don't ever want to write it out.

Magisus:
Updated to remove stdin support for SSH keys.

mwaggett (dismissed):

tvpartytonight (dismissed):

mwaggett (dismissed):

lib/r10k/action/deploy/environment.rb
```diff
@@ -43,6 +45,38 @@ def call
...
+            unless valid_token?(token)
+              raise R10K::Error, _("Supplied token contains invalid characters.")
+            end
+
+            { token: token }
+          end
+        end
+
+        def valid_token?(token)
+          return token =~ /^[\w\-\.~\+\/]+$/
```
justinstoller, line 78: It would be valuable to add a comment pointing to your link (https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/) about why we're using this bizarre set of characters.

lib/r10k/action/deploy/environment.rb
```diff
@@ -5,6 +5,7 @@
 require 'r10k/action/base'
 require 'r10k/action/deploy/deploy_helpers'
 require 'json'
+require 'tempfile'
```
justinstoller: This is unused outside of tests now, right?

Magisus: Good catch.

Magisus:
Updated

mwaggett (approved):

justinstoller (approved):

PR #1115, opened 2020-12-23
Plumb CLI credentials through to rugged provider

Magisus:

### DIFF
--- a/lib/r10k/git/rugged/credentials.rb
+++ b/lib/r10k/git/rugged/credentials.rb
@@ -10,9 +10,10 @@ class R10K::Git::Rugged::Credentials
   include R10K::Logging
 
   # @param repository [R10K::Git::Rugged::BaseRepository]
-  def initialize(repository)
+  def initialize(repository, credentials_from_cli = {})
     @repository = repository
     @called = 0
+    @credentials_from_cli = credentials_from_cli
   end
 
   def call(url, username_from_url, allowed_types)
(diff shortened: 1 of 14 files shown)


Magisus:
I successfully used this to deploy an environment from a private control repo on Github using an OAuth token.

lib/r10k/git/rugged/credentials.rb
```diff
@@ -43,7 +44,10 @@ def get_ssh_key_credentials(url, username_from_url)
 
     global_private_key = R10K::Git.settings[:private_key]
 
-    if per_repo_private_key
+    if @credentials_from_cli[:sshkey_file]
+      private_key = @credentials_from_cli[:sshkey_file]
+      logger.debug2 _("Using per-deploy private key %{key} for URL %{url}") % {key: private_key, url: url.inspect}
+    elsif per_repo_private_key
```
justinstoller, line 50: So... we already allow folks to specify the the ssh key path in their configuration? Can we treat `--sshkey_path` as overriding whatever option was given in that file. Like how in puppet you can set any setting in the puppet.conf but then override it on the command line with `--<setting_name>`? Then overwrite the value with the cli supplied value in whatever `R10K::Git.get_repo_settings(url)` is looking up?

Magisus: Just to be clear, this is an alternative plumbing approach (using the settings subsystem instead of treating it as an option on the repo), right, not a desire for a functionality change?

Just asking because the per-repo stuff does not behave like this does. CLI settings apply to all repos, so it's more like an override for the global private key that takes precedence over everything in the config. And I'm not sure we have a choice about that, because sources are not exposed as an abstraction at the CLI level (but they are in the config).

Magisus: So I would say, yes, it's probably doable, and has the advantage of touching fewer places and not changing the API as much. We can probably add something on that same settings object for the token too, though we'd want to make sure there isn't any way to print all the settings or something like that.

Magisus: I can plumb it through the settings architecture, which I think is better (fewer API changes), but I think I'll have to make a new setting for the private key to get the precedence to work right. I don't have access to the per-repo sections where I want to add this. Still seems like an improvement overall.

Magisus: See https://github.com/puppetlabs/r10k/pull/1116 for work on this approach

Magisus:
Closing this in favor of the settings-based appraoch.

PR #1116, opened 2021-01-05
[#1106, #1109] Plumb git creds from CLI to rugged

Magisus:
This is an alternative approach to https://github.com/puppetlabs/r10k/pull/1115 that avoids changing method signatures throughout the project. 

In this implementation, the `oauth_token_file` setting is available both on the CLI and and as a global config option, with the CLI version taking precedence. If we want to add a per-repo token setting to the config, we will have the same question as for SSH keys, described below.

When writing this, I assumed we wanted the CLI flag for the private key to take precedence over both the global and per-repo settings, but that made the precedence weird and I ended up having to add a new setting for the private key from the CLI. Does that assumption make sense? Or should we have it just override the global value, and let the per-repo still be the highest precedence? If so, we can do away with the weird extra setting. In this scenario, someone wanting to specify an SSH key file on the CLI should not configure it in the per-repo settings.

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -32,8 +32,7 @@ module R10K
           @visit_ok = true
 
           expect_config!
-          credentials = extract_credentials!
-          deployment = R10K::Deployment.new(@settings, credentials)
+          deployment = R10K::Deployment.new(@settings)
           check_write_lock!(@settings)
 
           deployment.accept(self)
@@ -43,40 +42,6 @@ module R10K
         include R10K::Action::Visitor
 
         private
-
-        def extract_credentials!
-          if @sshkey_path && @token_path
-            raise R10K::Error, "Cannot specify both an SSH key and a token to use with this deploy."
-          end
-
-          if @sshkey_path
-            if File.exists?(@sshkey_path)
-              { sshkey_path: @sshkey_path }
-            else
-              raise R10K::Error, _("{%path} does not exist, cannot load SSH key") % { path: @sshkey_path }
-            end
-          elsif @token_path
-            if @token_path == '-'
-              token = $stdin.read
-            elsif File.exists?(@token_path)
-              token = File.read(@token_path).strip
-            else
-              raise R10K::Error, _("{%path} does not exist, cannot load OAuth token") % { path: @token_path }
-            end
-
-            unless valid_token?(token)
-              raise R10K::Error, _("Supplied token contains invalid characters.")
-            end
-
-            { token: token }
-          end
-        end
-
-        # This regex is the only real requirement for OAuth token format,
-        # per https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/
-        def valid_token?(token)
-          return token =~ /^[\w\-\.~\+\/]+$/
-        end
 
         def visit_deployment(deployment)
           # Ensure that everything can be preloaded. If we cannot preload all
@@ -224,8 +189,8 @@ module R10K
                       'generate-types': :self,
                       'puppet-path': :self,
                       'puppet-conf': :self,
-                      'sshkey-path': :self,
-                      'token-path': :self,
+                      'private-key': :self,
+                      'oauth-token': :self,
                       'default-branch-override': :self)
         end
       end
--- a/lib/r10k/action/runner.rb
+++ b/lib/r10k/action/runner.rb
@@ -55,6 +55,10 @@ module R10K
           newval
         end
 
+        # Credentials from the CLI override both the global and per-repo
+        # credentials from the config, and so need to be handled specially
+        with_overrides = add_credential_overrides(with_overrides)
+
         @settings = R10K::Settings.global_settings.evaluate(with_overrides)
 
         R10K::Initializers::GlobalInitializer.new(@settings).call
@@ -91,6 +95,35 @@ module R10K
         end
 
         results
+      end
+
+      def add_credential_overrides(overrides)
+        sshkey_path = @opts[:'private-key']
+        token_path = @opts[:'oauth-token']
+
+        if sshkey_path && token_path
+          raise R10K::Error, "Cannot specify both an SSH key and a token to use with this deploy."
+        end
+
+        if sshkey_path
+          overrides[:git] ||= {}
+          overrides[:git][:private_key] = sshkey_path
+          if repo_settings = overrides[:git][:repositories]
+            repo_settings.each do |repo|
+              repo[:private_key] = sshkey_path
+            end
+          end
+        elsif token_path
+          overrides[:git] ||= {}
+          overrides[:git][:oauth_token] = token_path
+          if repo_settings = overrides[:git][:repositories]
+            repo_settings.each do |repo|
+              repo[:oauth_token] = token_path
+            end
+          end
+        end
+
+        overrides
       end
     end
   end
--- a/spec/unit/action/deploy/environment_spec.rb
+++ b/spec/unit/action/deploy/environment_spec.rb
@@ -27,8 +27,6 @@ describe R10K::Action::Deploy::Environment do
       described_class.new({:'no-force' => true}, [])
     end
 
-    it "normalizes environment names in the arg vector"
-
     it 'can accept a generate-types option' do
       described_class.new({ 'generate-types': true }, [])
     end
@@ -37,12 +35,12 @@ describe R10K::Action::Deploy::Environment do
       described_class.new({ 'puppet-path': '/nonexistent' }, [])
     end
 
-    it 'can accept an sshkey-path option' do
-      described_class.new({ 'sshkey-path': '/nonexistent' }, [])
+    it 'can accept a private-key option' do
+      described_class.new({ 'private-key': '/nonexistent' }, [])
     end
 
     it 'can accept a token option' do
-      described_class.new({ 'token-path': '/nonexistent' }, [])
+      described_class.new({ 'oauth-token': '/nonexistent' }, [])
     end
   end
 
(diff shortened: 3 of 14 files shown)


Magisus:
If we want to go this route, should we rename the flags to match the settings they're overriding, as Puppet does? Also happy for any feedback on the names themselves.

justinstoller:
This makes sense to me. Do you think it's a better, worse, or just different route than in #1115 ? 

> should we rename the flags to match the settings they're overriding, as Puppet does?

I would say yes.

> I assumed we wanted the CLI flag for the private key to take precedence over both the global and per-repo settings, but that made the precedence weird and I ended up having to add a new setting for the private key from the CLI.

I think that's correct, the current behavior of specifying a `--private-key` on the command line and that taking effect _only if you haven't specified a per-repo key in your config_ seems weird to me. I would expect the key with that behavior to be `--default-private-key` or similar. If we can't change that, I would probably call the flag `--private-key-override` or something. wrt changing the behavior of the `--private-key` flag, but I don't know if it's used at all or if the current behavior could be considered a bug.

Magisus:
Hold on, so the current behavior (as written in this PR) is that the private key specified on the CLI _does_ take precedence over everything in the config. But the only way I could think of to get it to work right was to add a "dummy" global setting, that has the highest precedence, because otherwise there's no way to tell whether the top-level `private_key` setting was overridden or not (and should take precedence over the per-repo `private_keys`). That felt a little gross, just because technically you can also use that setting `in` the config, but if we don't really document it that seems okay.

Also the `--private-key` flag is brand new as of my last PR (and hasn't been released) so we can do whatever we want with its behavior.

Magisus:
Is there a `private-key` flag currently? I don't think there is, but maybe I missed something?

Once we rename `--sshkey-path` to `--private-key`, we can give it the behavior of overriding everything (like it has in this PR currently), but the only way I could think of to do that was to create a new "dummy" setting so we can tell if we need to give it precedence, vs just overriding the global `private_key` setting, which would in turn be overridden by any per-repo key. The dummy setting is weird becuase it can _technically_ be used in the config file too. But we can probably avoid that just by not documenting it. But maybe there's another way?

spec/unit/action/runner_spec.rb
```diff
@@ -170,6 +171,37 @@ def call
...
+      token_file = Tempfile.new('token')
+      token_file.write('my_token')
+      token_file.close
+      runner = described_class.new({ 'token-path': token_file.path }, %w[args yes], action_class)
+      runner.call
+      expect(runner.instance.settings[:git][:oauth_token_file]).to eq(token_file.path)
+    end
+
+    it 'saves the token from stdin' do
+      allow($stdin).to receive(:read).and_return("my_token")
```
Magisus: I guess this isn't needed anymore, since we do the reading in the credentials class instead.

spec/unit/git/rugged/credentials_spec.rb
```diff
@@ -78,6 +90,49 @@
...
+      R10K::Git.settings[:repositories] = [{remote: "https://tessier-ashpool.freeside/repo.git"}]
+      expect {
+        subject.get_plaintext_credentials("https://tessier-ashpool.freeside/repo.git", nil)
+      }.to raise_error(R10K::Git::GitError, /invalid characters/)
+    end
+
+    it 'uses the token as a password' do
+      token_file = Tempfile.new('token')
+      token_file.write('my_token')
+      token_file.close
```
Magisus, line 115: Should add a test for the stdin case too.

spec/unit/action/runner_spec.rb
```diff
@@ -170,6 +171,37 @@ def call
...
   end
 
+  describe "configuring git credentials" do
+    it 'errors if both token and key paths are passed' do
+      runner = described_class.new({ 'token-path': '/nonexistent',
+                                     'sshkey-path': '/also/fake' }, %w[args yes], action_class)
+      expect{ runner.call }.to raise_error(R10K::Error, /Cannot specify both/)
+    end
+
+    it 'passes sshkey path to deployment' do
```
mwaggett: (nit) should this say something like 'saves the sshkey path in settings hash' or something? since I don't think we're passing to deployment anymore?

spec/unit/git/rugged/credentials_spec.rb
```diff
@@ -37,8 +37,20 @@
   describe "generating ssh key credentials" do
     after(:each) { R10K::Git.settings.reset! }
 
+    it "prefers a key from CLI" do
+      allow(File).to receive(:readable?).with("/etc/puppetlabs/r10k/ssh/clikey-id_rsa").and_return true
```
mwaggett: Should we also test the case where the file is not readable? I know we already test that below, but that specifically tests the `private_key` setting, rather than `private_key_from_cli` (or the per-repo private key, for that matter). I know it all turns out the same, since we check for readability after choosing which key to use anyway, but that feels like an implementation detail?

justinstoller:
> ...but the only way I could think of to do that was to create a new "dummy" setting so we can tell if we need to give it precedence, vs just overriding the global private_key setting, which would in turn be overridden by any per-repo key.

Gotcha, I was misunderstanding the issue.

I think what you're doing makes sense and though weird I think simply not documenting it seems fine. I think the only other way would be if we could merge this value into all the per-repo settings when we parse the command line flags. And I'm not sure that's feasible.

mwaggett (commented):
This looks reasonable to me. Do you want me to look at the other PR for comparison or do you feel pretty confident that this is the direction we'll go?

fwiw I agree that the CLI flags should override everything else.

Magisus:
I can look into the merging, but I think it happens in a different-enough spot that it might not be any cleaner to plumb this through as a separate value till we get there. Thanks for the input!

spec/unit/git/rugged/credentials_spec.rb
Magisus: Sure I can.

Magisus:
Yeah I think this is better. A lot less invasive, and gives the option for people to use tokens without a CLI flag, which seems nice.

Magisus:
Sorry that last commit wound up being a bit of a mess, but I got it all working, with an OAuth token config setting that mirrors what we have for `private-key`, global and per-repo, and no dummy settings anywhere. I'll squash this and clean up commit messages once we agree this is what we want.

lib/r10k/action/deploy/environment.rb
```diff
@@ -224,8 +189,8 @@ def allowed_initialize_opts
                       'generate-types': :self,
                       'puppet-path': :self,
                       'puppet-conf': :self,
-                      'sshkey-path': :self,
-                      'token-path': :self,
+                      'private-key': :self,
+                      'oauth-token': :self,
```
mwaggett, line 193: ugh I wish these could still be `path`s, but I understand why we're doing this 😞

lib/r10k/action/runner.rb
```diff
@@ -55,6 +55,10 @@ def setup_settings
           newval
         end
 
+        # Credentials from the CLI override both the global and per-repo
+        # credentials from the config, and so need to be handled speically
```
mwaggett:
```suggestion
        # credentials from the config, and so need to be handled specially
```

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -37,12 +35,12 @@
       described_class.new({ 'puppet-path': '/nonexistent' }, [])
     end
 
-    it 'can accept an sshkey-path option' do
-      described_class.new({ 'sshkey-path': '/nonexistent' }, [])
+    it 'can accept an private-key option' do
```
mwaggett:
```suggestion
    it 'can accept a private-key option' do
```

spec/unit/action/deploy/module_spec.rb
```diff
@@ -34,12 +34,12 @@
       described_class.new({ cachedir: '/nonexistent' }, [])
     end
 
-    it 'can accept an sshkey-path option' do
-      described_class.new({ 'sshkey-path': '/nonexistent' }, [])
+    it 'can accept an private-key option' do
```
mwaggett:
```suggestion
    it 'can accept a private-key option' do
```

spec/unit/action/runner_spec.rb
```diff
@@ -170,6 +171,58 @@ def call
...
+      runner = described_class.new({ 'oauth-token': '/my/token/path' }, %w[args yes], action_class)
+      runner.call
+      expect(runner.instance.settings[:git][:oauth_token]).to eq('/my/token/path')
+    end
+
+    it 'saves the stdin placeholder to the settings hash' do
+      runner = described_class.new({ 'oauth-token': '-' }, %w[args yes], action_class)
+      runner.call
+      expect(runner.instance.settings[:git][:oauth_token]).to eq('-')
+    end
```
mwaggett: Is this a meaningful test? It looks like it's identical to the regular token path test, just with a different value for the path.

mwaggett (commented):
Whenever you squash, would you mind fixing up the commit message(s)?

spec/unit/action/runner_spec.rb
Magisus: Probably not. I was sorta thinking to leave it for "tests as documentation", but I don't feel strongly.

lib/r10k/action/runner.rb
```diff
@@ -92,6 +96,36 @@ def settings_from_config(override_path)
...
+          overrides[:git] ||= {}
+          overrides[:git][:private_key] = sshkey_path
+          if repo_settings = overrides[:git][:repositories]
+            repo_settings.each do |repo|
+              repo[:private_key] = sshkey_path
+            end
+          end
+        elsif token_path
+          overrides[:git] ||= {}
+          overrides[:git][:private_key] = sshkey_path
```
justinstoller: why are we setting this to nil here?

Magisus: I feel like this might be a typo. It will ignore whichever one isn't needed for the kind of URL associated with each remote, so it's not an issue to have both set. Thanks for catching this.

mwaggett (approved):

mwaggett:
not merging since @tvpartytonight said he was gonna take another look in the morning

tvpartytonight (approved):

PR #1119, opened 2021-01-18
Very first test to check git settings

tuxmea:
check :ref, :commit, :branch and :git config settings and verify that they start with single quotes, except when setting :branch to :control_branch

reason: using a symbol causes git to throw errors

this is my first idea on how to check.
any feedback or recommendation is welcome.

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -85,6 +85,7 @@ class Puppetfile
     dsl.instance_eval(puppetfile_contents, @puppetfile_path)
     
     validate_no_duplicate_names(@modules)
+    validate_ref_names()
     @loaded = true
   rescue SyntaxError, LoadError, ArgumentError, NameError => e
     raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile_path})
@@ -105,6 +106,20 @@ class Puppetfile
       msg += ' '
       msg += _("Remove the duplicates of the following modules: %{dupes}" % { dupes: dupes.join(' ') })
       raise R10K::Error.new(msg)
+    end
+  end
+
+  def validate_ref_names()
+    # check that all elements with name :git, :ref, :branch, :commit are of type string
+    # unless :branch is set to :control_repo
+    [/:git/, /:ref/, /:branch/, /:commit/].each do |obj|
+      puppetfile_contents.split("\n").grep(obj).each do |line|
+        result = line.split('=>')[1].strip.tr(',', '')
+        next if obj == /:branch/ and result == ':control_branch'
+        if ! result.start_with? "'" or ! result.end_with? "'"
+          raise R10K::Error.new("#{obj.to_s} must be of type string")
+        end
+      end
     end
   end
 


tuxmea:
note: r10k puppetfile install will not fail, but r10k deploy environment fails.

justinstoller:
I think you could add this validation to the `mod` method in `DSL`. I think in that case, you wouldn't be grepping the file contents, but checking to see if `args` passed is a hash, and if so if it contains these keys, and if so, those values are a `String`. like:
```rb
if args.is_a?(Hash)
  if !args.
    select {|k,v| %i{ref commit branch git}.include?(k) }.
    all? {|k,v| v.is_a? String }
      raise R10K::Error.new(...)
  end
end
```

tuxmea:
@justinstoller I was already checking DSL, but I did not find a solution onhow to access the args element.
when querying `modules` instance variable I receive an Array:

```
=> [#<R10K::Module::Git:0x00007fac181f7838
  @args={:git=>"https://github.com/example42/psick-hieradata", :branch=>:control_branch, :default_branch=>"production"},
  @default_ref="production",
  @desired_ref=:control_branch,
  @dirname="/Users/mea/Documents/github/puppet/r10k/modules",
  @environment=nil,
  @name="hieradata",
  @origin="Puppetfile",
  @owner="example42",
  @path=#<Pathname:/Users/mea/Documents/github/puppet/r10k/modules/hieradata>,
  @remote="https://github.com/example42/psick-hieradata",
...
```

Using .each method does not provide the desired results.
I am stuck here trying to access the args element.

justinstoller:
Sorry, before we create the module object but are evaluating the puppetfile as ruby code (rather than as a string) is what I meant. I started walking through the module creation code to see where it validation might best live, I hope you don't mind if I include my code safari below:

So, when we [eval the puppetfile](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/puppetfile.rb#L85) each module is created when the [`mod`](https://github.com/puppetlabs/r10k/blob/master/lib/r10k/puppetfile.rb#L289) method is invoked that would be one place to add it if we want this validation to happen for all modules. It immediately calls into `add_module` that is implemented in a couple of providers. Mainly [here](https://github.com/puppetlabs/r10k/blob/1e5a637804f0fd09dcc9f023f95430b06fe1b060/lib/r10k/puppetfile.rb#L127), where we already are trying to validate the install path (maybe this is similar?). But also in this experimental provider [with_modules](https://github.com/puppetlabs/r10k/blob/1e5a637804f0fd09dcc9f023f95430b06fe1b060/lib/r10k/environment/with_modules.rb#L76). Finally, and perhaps, the best place that we could put this info, is where already validate the args to the [git provider](https://github.com/puppetlabs/r10k/blob/1e5a637804f0fd09dcc9f023f95430b06fe1b060/lib/r10k/module/git.rb#L34)

It looks like there's a little bit of validation everywhere, each piece validating some of the bits it cares most about. I have a a feeling that the git provider is the best for this.

binford2k:
Agree. I think it belongs directly in the git provider. Either where @justinstoller suggested, or maybe in the [`validate_ref`](https://github.com/puppetlabs/r10k/blob/5e774259276693b48fd08fe3a4aad9295deaff7a/lib/r10k/module/git.rb#L92) method.

@tuxmea, I'm going to close this PR for now but if you want to add the check there, feel free to PR that! Thanks! 🍻

PR #1120, opened 2021-01-27
Fix leaky env module deploys

reidmv:
This fixes a problem with "Implement less restrictive env module merge", merged at e6831cbfcf813a583cbbb1cfe6df76027eb97896.

Previously, when an environment should have overridden a Puppetfile-defined module, it did not. This is because the Puppetfile still contained a copy of the module, and when visited, it deployed that copy, overwriting the environment's version of the module.

### DIFF
--- a/lib/r10k/environment/with_modules.rb
+++ b/lib/r10k/environment/with_modules.rb
@@ -56,12 +56,13 @@ class R10K::Environment::WithModules < R10K::Environment::Base
 
   def accept(visitor)
     visitor.visit(:environment, self) do
+      validate_no_module_conflicts
+
       @modules.each do |mod|
         mod.accept(visitor)
       end
 
       puppetfile.accept(visitor)
-      validate_no_module_conflicts
     end
   end
 
@@ -90,7 +91,7 @@ class R10K::Environment::WithModules < R10K::Environment::Base
     @managed_content[install_path] = Array.new unless @managed_content.has_key?(install_path)
 
     mod = R10K::Module.new(name, install_path, args, self.name)
-    mod.origin = 'Environment'
+    mod.origin = :environment
 
     @managed_content[install_path] << mod.name
     @modules << mod
@@ -101,16 +102,18 @@ class R10K::Environment::WithModules < R10K::Environment::Base
     conflicts = (@modules + @puppetfile.modules)
                 .group_by { |mod| mod.name }
                 .select { |_, v| v.size > 1 }
-                .map(&:first)
     unless conflicts.empty?
-      conflicts_str = conflicts.join(', ')
+      conflicts_str = conflicts.keys.join(', ')
+      conflicting_mods = conflicts.values.flatten.reject { |mod| @modules.include?(mod) }
       log_msg = _('Environment and Puppetfile both define the following modules, Puppetfile ' \
                   'definition will be ignored: %{mods}' % { mods: conflicts_str })
       case conflict_opt = @options[:module_conflicts]
-      when 'override_puppetfile'
-        logger.debug log_msg
       when 'override_puppetfile_and_warn', nil
         logger.warn log_msg
+        conflicting_mods.each { |mod| @puppetfile.remove_module(mod) }
+      when 'override_puppetfile'
+        logger.debug log_msg
+        conflicting_mods.each { |mod| @puppetfile.remove_module(mod) }
       when 'error'
         raise R10K::Error, _('Puppetfile cannot contain module names defined by environment ' \
                              '%{env}; Remove the conflicting definitions of the following modules: ' \
--- a/lib/r10k/environment/with_modules.rb
+++ b/lib/r10k/environment/with_modules.rb
@@ -103,21 +103,23 @@ class R10K::Environment::WithModules < R10K::Environment::Base
                 .group_by { |mod| mod.name }
                 .select { |_, v| v.size > 1 }
     unless conflicts.empty?
-      conflicts_str = conflicts.keys.join(', ')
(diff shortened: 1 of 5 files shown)


lib/r10k/environment/with_modules.rb
```diff
@@ -90,27 +91,29 @@ def add_module(name, args)
...
-    @puppetfile.load unless @puppetfile.loaded?
-    conflicts = (@modules + @puppetfile.modules)
+    puppetfile.load unless puppetfile.loaded?
+    conflicts = (@modules + puppetfile.modules)
                 .group_by { |mod| mod.name }
                 .select { |_, v| v.size > 1 }
-                .map(&:first)
     unless conflicts.empty?
-      conflicts_str = conflicts.join(', ')
+      conflicts_str = conflicts.keys.join(', ')
```
mwaggett: (nit) can we give this a more descriptive name, like maybe `conflicting_module_names`?

spec/unit/puppetfile_spec.rb
```diff
@@ -302,12 +283,12 @@ def expect_wrapped_error(orig, pf_path, wrapped_error)
         block.call
       end
 
-      mod1 = spy('module')
+      mod1 = double('module', :cachedir => :none)
```
mwaggett: Forgive me for not knowing much about `spy`s, but why did all the `spy`s change to `double`s?

mwaggett (dismissed):
This seems reasonable to me, but I'd love for someone else to take a look as well.

spec/unit/puppetfile_spec.rb
reidmv: > Spies are [a] type of test double that support [...] allowing you to expect that a message has been received after the fact, using have_received.

Spies are the least specific kind of double, and allow any method to be called on them. However, no matter what method is called, a spy pretty much always just returns itself.

For this test, I'd like the doubles to respond to a specific method, `#cachedir`, with a common value. This is because in [`32af983`](https://github.com/puppetlabs/r10k/pull/1120/commits/32af9836ae1fbfa65df1b76bb06b2df44f198dcf#diff-1e18cef8c5fe2a6b1b2146ea4282b5bb0e40cf5f6fec5bdae9112d0830cf12fcR217) we implemented grouping of modules using `#group_by { |mod| mod.cachedir }`. So, I upgraded the spies to the next level of specificity, doubles, so that they will report a known value when interrogated for their `#cachedir`.

> When creating a double, you can allow messages (and set their return values) by passing a hash.

Thinking about it now I'm not really testing for how the queue has absorbed or grouped these modules so it kinda doesn't matter, I guess. However, I think in the commit to improve variable names I'll upgrade them to verifying doubles, which will at least have them alert if any methods are called on them that aren't valid module methods. As a non-spy verifying double it becomes required to let them expect that they're gonna have `#cachedir` called on them, and so makes sense to give them a value to return for it.

lib/r10k/environment/with_modules.rb
reidmv: Sure thing, I'll push an update.

spec/unit/puppetfile_spec.rb
mwaggett: gotcha, thanks so much for this explanation!

lib/r10k/environment/with_modules.rb
```diff
@@ -90,31 +91,33 @@ def add_module(name, args)
...
-    conflicts = (@modules + @puppetfile.modules)
+    puppetfile.load unless puppetfile.loaded?
+    conflicts = (@modules + puppetfile.modules)
                 .group_by { |mod| mod.name }
                 .select { |_, v| v.size > 1 }
-                .map(&:first)
     unless conflicts.empty?
-      conflicts_str = conflicts.join(', ')
+      conflicting_module_names = conflicts.keys.join(', ')
+      conflicting_puppetfile_modules = conflicts.values.flatten.reject { |mod| @modules.include?(mod) }
```
mwaggett: I think it would be clearer if this `select`ed the conflicts in `puppetfile.modules` instead of `reject`ing those from `@modules`. I had to read this a few times to understand the intention.

spec/unit/puppetfile_spec.rb
```diff
@@ -323,12 +304,11 @@ def expect_wrapped_error(orig, pf_path, wrapped_error)
         block.call
       end
 
-      mod1 = spy('module')
+      mod1 = instance_double('R10K::Module::Base', :cachedir => :none)
+      mod2 = instance_double('R10K::Module::Base', :cachedir => :none)
```
mwaggett, line 308: hmm okay I understand the `spy`->`double` change now, but why did you now only change these to `instance_double`s? Is there something particular about this test that makes it relevant here but not for the other tests you updated?

mwaggett (commented):
okay just a couple more small things! 😅 

I think it would maybe be good for @justinstoller to look at this, since he's the one who originally did the grouping-modules-by-vcs-cachedir work.

spec/unit/puppetfile_spec.rb
reidmv: Nope! Just didn't know about verifying doubles until our conversation. Pretty much all of these doubles could be upgraded to verifying doubles. I learned something new doing some research about spies. 🙂

lib/r10k/environment/with_modules.rb
reidmv: I'll flip it around

reidmv:
@mwaggett alright, latest suggestions applied. 👍  / 👎  or would you like @justinstoller to take a quick look at the `#group_by` part?

justinstoller (approved):

mwaggett (approved):

PR #1122, opened 2021-02-01
(CODEMGMT-1350) Fall back to default branch even when overridden

Magisus:
Currently, if a default branch override for a module is specified, if
that branch doesn't exist, r10k will error without checking the module's
normal default branch. This is not desirable, so this commit updates the
fallback logic so that the default branch ref will also be checked, if
the branch override fails to resolve. This makes it more likely that we
will successfully resolve a module, rather than erroring.

```
```

### DIFF
--- a/doc/dynamic-environments/usage.mkd
+++ b/doc/dynamic-environments/usage.mkd
@@ -66,11 +66,12 @@ Update a single environment and specify a default branch override:
 
     r10k deploy environment my_working_environment --puppetfile --default-branch-override default_branch_override
 
-This will update the given environment and update all contained modules, overrideing 
-the :default_branch entry in the Puppetfile of each module. This is used primarily to allow
+This will update the given environment and update all contained modules, overriding
+the :default_branch entry in the Puppetfile of each module. If the specified override branch is not
+found, it will fall back to the normal default branch and attempt to use that. This is used primarily to allow
 automated r10k solutions using the control_branch pattern with a temporary branch deployment to 
-ensure the deployment is pushed to the correct module repository branch. Note that the :default_branch
-is only ever utilized if the desired ref cannot be located.
+ensure the deployment is pushed to the correct module repository branch. Note that the :default_branch and its
+override are only ever used if the specific desired ref cannot be located.
 
 ### Deploying modules
 
--- a/lib/r10k/module/git.rb
+++ b/lib/r10k/module/git.rb
@@ -28,6 +28,11 @@ class R10K::Module::Git < R10K::Module::Base
   #   @return [String]
   attr_reader :default_ref
 
+  # @!attribute [r] default_override_ref
+  #   @api private
+  #   @return [String]
+  attr_reader :default_override_ref
+
   def initialize(title, dirname, args, environment=nil)
     super
 
@@ -37,7 +42,7 @@ class R10K::Module::Git < R10K::Module::Base
   end
 
   def version
-    validate_ref(@desired_ref, @default_ref)
+    validate_ref(@desired_ref, @default_ref, @default_override_ref)
   end
 
   def properties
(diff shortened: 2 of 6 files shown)


Magisus:
This is an improvement requested in service of CD4PE, so I've used an internal ticket number. If this will mess with any community workflows, please let me know, and we can look into options for backwards compatibility.

lib/r10k/module/git.rb
```diff
@@ -81,6 +88,11 @@ def validate_ref(desired, default)
...
       end
 
+      if default_override
+        msg << "or resolve the default branch override '%{default_override}',"
+        vars[:default_override] = default_override
+      end
+
       if default
         msg << "or resolve default ref '%{default}'"
         vars[:default] = default
```
mwaggett, line 98: I won't let me comment on the actual line, but do we think the "no default provided" message below will make sense if there's a default override but no actual default?

spec/unit/module/git_spec.rb
```diff
@@ -268,6 +268,48 @@ def test_module(extra_opts, env=nil)
...
+              expect(mock_repo).to receive(:resolve).with('default_override').and_return('5566aabb')
+              mod = test_module({branch: :control_branch,
+                                 default_branch: 'default',
+                                 default_branch_override: 'default_override'},
+                                 mock_env)
+              expect(mod.properties).to include(expected: 'default_override')
+            end
+          end
+
+          context "when default branch override is not resolvable" do
```
mwaggett: (nit, just to match the other context above)
```suggestion
          context "and the default branch override is not resolvable" do
```

spec/unit/module/git_spec.rb
```diff
@@ -268,6 +268,48 @@ def test_module(extra_opts, env=nil)
...
+            context "and default branch is provided" do
```
mwaggett, line 289: Should we test the case where the default branch is provided, but also not resolvable? That feels different from just not providing a default branch.

mwaggett (commented):
Do we need to talk to Nick about whether this constitutes a breaking change or not?

lib/r10k/module/git.rb
Magisus: The docs make clear that there is a "normal default" and a "default override". This is also the language the puppetfile uses for the options themselves.

Magisus: The whole message becomes: "Could not resolve control repo branch or resolve the default branch override 'override', and no default provided." I think it sounds okay.

spec/unit/module/git_spec.rb
Magisus: We already test default suceeding (proving that we fall back), and elsewhere we test the default failing, demonstrating what that looks like. It feels like it's covered to me. But I can add a case if you think that's not sufficient.

lib/r10k/module/git.rb
mwaggett: okey!

spec/unit/module/git_spec.rb
mwaggett: I'm not familiar with the tests in this repo, so if you think it's covered then I trust you.
I was just thinking that the default branch being unresolvable has different consequences depending on whether you use a default_override or not. When using a default_override, the default being unresolvable only matters if the default_override is also unresolvable.

Magisus: I added something for that last case. If the default is not resolvable and the override is, that doesn't matter because we never try it.

mwaggett (dismissed):

npwalker (dismissed):
LGTM.

doc/dynamic-environments/usage.mkd
```diff
@@ -67,10 +67,11 @@ Update a single environment and specify a default branch override:
     r10k deploy environment my_working_environment --puppetfile --default-branch-override default_branch_override
 
 This will update the given environment and update all contained modules, overrideing
```
justinstoller: nit: can you change "overrideing" here since you're right next to it.

mwaggett (approved):

justinstoller (approved):

PR #1124, opened 2021-02-06
Bitbucket OAuth enablement

Magisus:
PR to fix bugs found while testing OAuth tokens with bitbucket.

### DIFF
--- a/lib/r10k/git/rugged/credentials.rb
+++ b/lib/r10k/git/rugged/credentials.rb
@@ -92,8 +92,9 @@ class R10K::Git::Rugged::Credentials
 
   # This regex is the only real requirement for OAuth token format,
   # per https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/
+  # Bitbucket's tokens also can include an underscore, so that is added here.
   def valid_token?(token)
-    return token =~ /^[\w\-\.~\+\/]+$/
+    return token =~ /^[\w\-\.~_\+\/]+$/
   end
 
   def get_default_credentials(url, username_from_url)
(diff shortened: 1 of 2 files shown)


justinstoller:
What's up with this?

Magisus:
I could go either way on it. It's fine to go in, but I think there are other issues with Bitbucket OAuth besides this. Let me update the commit to reference the ticket at least.

Magisus:
I had one more commit to fix a general bug related to auth retries. Once that is fixed, the underlying issue with Bitbucket is that it expects a different username along with the token. I think how we fix that could be a bigger question, to avoid special-casing in this code.

lib/r10k/git/rugged/credentials.rb
```diff
@@ -67,9 +67,14 @@ def get_plaintext_credentials(url, username_from_url)
     end
 
     if token_path = per_repo_oauth_token || R10K::Git.settings[:oauth_token]
-      if token_path == '-'
+      if @oauth_token
+        token = @oauth_token
```
justinstoller: Is it just me or is setting a token to the value of an instance variable in one branch and in the next branch setting the same instance variable to the value of the same token confusing?

IMO, it'd help readability if we moved this conditional and the token validation into a separate function (say `foo_fn`) and conditionally called it like
```rb
@oauth_token ||= foo_fn()
```
... and then set `password` to `@oauth_token`.

Magisus: It's definitely terrible, I never really meant to put this up (and didn't look at it when I pushed it just now), I just wanted to get something working. I'll see about refactoring it.

Magisus: Updated, with a test, lemme know what you think.

justinstoller (approved):

mwaggett (approved):

PR #1125, opened 2021-02-11
(maint) Fix problem introduced in #1120

reidmv:
The changes in #1120 caused the Puppetfile to be loaded too early,
breaking deploys when performed with a default-branch-override. This
commit refactors how Puppetfile module conflicts are handled to avoid
needing to load the Puppetfile early.

### DIFF
--- a/lib/r10k/environment/with_modules.rb
+++ b/lib/r10k/environment/with_modules.rb
@@ -49,15 +49,34 @@ class R10K::Environment::WithModules < R10K::Environment::Base
     return @modules if puppetfile.nil?
 
     puppetfile.load unless puppetfile.loaded?
+    @modules + puppetfile.modules
+  end
 
-    env_mod_names = @modules.map(&:name)
-    @modules + puppetfile.modules.select { |mod| !env_mod_names.include?(mod.name) }
+  def module_conflicts?(mod_b)
+    conflict = @modules.any? { |mod_a| mod_a.name == mod_b.name }
+    return false unless conflict
+
+    msg_vars = {src: mod_b.origin, name: mod_b.name}
+    msg_error = _('Environment and %{src} both define the "%{name}" module' % msg_vars)
+    msg_continue = _("#{msg_error}. The %{src} definition will be ignored" % msg_vars)
+
+    case conflict_opt = @options[:module_conflicts]
+    when 'override_and_warn', nil
+      logger.warn msg_continue
+    when 'override'
+      logger.debug msg_continue
+    when 'error'
+      raise R10K::Error, msg_error
+    else
+      raise R10K::Error, _('Unexpected value for `module_conflicts` setting in %{env} ' \
+                           'environment: %{val}' % {env: self.name, val: conflict_opt})
+    end
+
+    true
   end
 
   def accept(visitor)
     visitor.visit(:environment, self) do
-      validate_no_module_conflicts
-
       @modules.each do |mod|
         mod.accept(visitor)
       end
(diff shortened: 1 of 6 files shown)


lib/r10k/environment/with_modules.rb
```diff
@@ -49,15 +49,34 @@ def modules
...
-    @modules + puppetfile.modules.select { |mod| !env_mod_names.include?(mod.name) }
+  def module_conflicts?(mod_b)
+    conflict = @modules.any? { |mod_a| mod_a.name == mod_b.name }
+    return false unless conflict
+
+    msg_vars = {src: mod_b.origin, name: mod_b.name}
+    msg_error = _('Environment and %{src} both define the "%{name}" module' % msg_vars)
+    msg_continue = _("#{msg_error}. The %{src} definition will be ignored" % msg_vars)
+
+    case conf = @options[:module_conflicts]
```
mwaggett: (nit) I liked calling this `conflict_opt`

Magisus:
I confirmed that this fixes the issue in the code-manager tests. It looks like it still should also do what you intended, so 👍 from me.

Magisus (approved):

spec/unit/puppetfile_spec.rb
```diff
@@ -127,6 +127,18 @@
...
+    it "should disable and not add modules that conflict with the environment" do
+      env = instance_double('R10K::Environment::Base')
+      mod = instance_double('R10K::Module::Base', name: 'conflict', origin: :puppetfile)
+      allow(mod).to receive(:origin=).and_return(nil)
+      allow(subject).to receive(:environment).and_return(env)
+      allow(env).to receive(:'module_conflicts?').with(mod).and_return(true)
+
+      allow(R10K::Module).to receive(:new).with('test', anything, anything, anything).and_return(mod)
+      expect { subject.add_module('test', {}) }.not_to change { subject.modules }
+      expect(subject.disabled_modules).to eq [mod]
```
mwaggett: It looks like this is the only place `disabled_modules` is used. Could we maybe check that the `mod` _isn't_ included in `subject.modules`, rather than adding a whole new attr just for testing?

spec/unit/environment/with_modules_spec.rb
```diff
@@ -21,90 +21,54 @@
...
-  describe "dealing with Puppetfile module conflicts" do
+  describe "dealing with module conflicts" do
     context "with no module conflicts" do
       it "validates when there are no conflicts" do
-        mod = instance_double('R10K::Module::Base', :name => 'nonconflict')
-        expect(subject.puppetfile).to receive(:load)
-        expect(subject.puppetfile).to receive(:modules).and_return [mod]
-        expect { subject.validate_no_module_conflicts }.not_to raise_error
+        mod = instance_double('R10K::Module::Base', name: 'nonconflict', origin: :puppetfile)
+        expect { subject.module_conflicts?(mod) }.not_to raise_error
```
mwaggett: Can we check that this returns `false` rather than just not erroring? I think that would be more accurate. This would still pass if it caught conflicts but overrode them, right?

spec/unit/environment/with_modules_spec.rb
```diff
@@ -21,90 +21,54 @@
...
-
-        visitor = spy('visitor')
-        expect(visitor).to receive(:visit).with(:environment, subject) { |&block| block.call }
-        expect(visitor).to receive(:visit).with(:puppetfile, pf) { |&block| block.call }
-
-        expect(mod1).to receive(:accept)
-        expect(mod2).not_to receive(:accept)
-
-        subject.accept(visitor)
+        expect { subject.module_conflicts?(mod) }.not_to raise_error
```
mwaggett: similarly, can we expect this to return `true`?

spec/unit/environment/with_modules_spec.rb
```diff
@@ -21,90 +21,54 @@
...
-        mod = instance_double('R10K::Module::Base', :name => 'stdlib')
-        expect(subject.puppetfile).to receive(:load)
-        expect(subject.puppetfile).to receive(:modules).at_least(:once).and_return([mod])
-        expect { subject.validate_no_module_conflicts }.to raise_error(R10K::Error, /Puppetfile.*defined.*conflict/i)
+        mod = instance_double('R10K::Module::Base', name: 'stdlib', origin: :puppetfile)
+        expect { subject.module_conflicts?(mod) }.to raise_error(R10K::Error, /Puppetfile.*both define.*/i)
       end
     end
 
     context "with module conflicts and 'override_puppetfile' behavior" do
```
mwaggett:
```suggestion
    context "with module conflicts and 'override' behavior" do
```

spec/unit/environment/with_modules_spec.rb
```diff
@@ -21,90 +21,54 @@
...
-      let(:subject_params) {{ :module_conflicts => 'override_puppetfile' }}
+      let(:subject_params) {{ :module_conflicts => 'override' }}
       it "does not raise an error" do
-        mod = instance_double('R10K::Module::Base', :name => 'stdlib')
-        expect(subject.puppetfile).to receive(:load)
-        expect(subject.puppetfile).to receive(:modules).at_least(:once).and_return([mod])
+        mod = instance_double('R10K::Module::Base', name: 'stdlib', origin: :puppetfile)
         expect(subject.logger).to receive(:debug).with(/Puppetfile.*both define.*ignored/i)
-        expect { subject.validate_no_module_conflicts }.not_to raise_error
+        expect { subject.module_conflicts?(mod) }.not_to raise_error
```
mwaggett: same here

spec/unit/puppetfile_spec.rb
reidmv: The line above this should do that, yes?

I originally added `@disabled_modules` because I thought it might be needed to log or debug something somewhere. However, I never ended up doing that. We could fully remove it right now with no effect. Should I just do that? Seems like probably...

lib/r10k/environment/with_modules.rb
reidmv: I'll change it back 🙂

spec/unit/puppetfile_spec.rb
mwaggett: I think so, yeah. There's already enough confusing stuff in this repo, so less is more imo.

I guess the first line kind of does that? so I don't feel too strongly about changing that part. it feels a _little_ more implementation-aware than seems ideal, but I think that's a larger problem in this repo 😅

mwaggett (approved):
Looks good to me! Wanna squash your commits?

reidmv:
@mwaggett I'm good with leaving this as two commits, with the `@disabled_modules` thing left in the commit history in case there's a near-future need for better logging or debugging related to it, I think it'll serve as a memory jogger for one way to do that. If you'd prefer to squash on merge though I won't object 🙂. It's good to go from my end

PR #1128, opened 2021-02-24
Implement strip_component source feature

reidmv:
This feature allows r10k to be configured to strip a leading component string from a source's environment names. This is useful because it allows e.g. Git branches to include organizational prefixes such as "environment/production", "environment/development", but r10k/Puppet to deploy and use environments named "production", "development" from these branches.

### DIFF
--- a/lib/r10k/environment/name.rb
+++ b/lib/r10k/environment/name.rb
@@ -72,9 +72,15 @@ module R10K
       private
 
       def derive_name(name, strip_component)
-        if strip_component == nil
-          name
-        elsif %r{^/.*/$}.match(strip_component)
+        return name unless strip_component
+
+        unless strip_component.is_a?(String)
+          raise _('Improper configuration value given for strip_component setting in %{src} source. ' \
+                  'Value must be a string, a /regex/, false, or omitted. Got "%{val}" (%{type})' \
+                  % {src: @source, val: strip_component, type: strip_component.class})
+        end
+
+        if %r{^/.*/$}.match(strip_component)
           regex = Regexp.new(strip_component[1..-2])
           name.gsub(regex, '')
         elsif name.start_with?(strip_component)
@@ -82,10 +88,6 @@ module R10K
         else
           name
         end
-      rescue TypeError, NoMethodError
-        raise _('Improper configuration value given for strip_component setting in %{src} source. ' \
-                'Value must be a string, a /regex/, or omitted. Got "%{val}" (%{type})' \
-                % {src: @source, val: strip_component, type: strip_component.class})
       end
 
       def derive_prefix(source,prefix)
--- a/lib/r10k/environment/name.rb
+++ b/lib/r10k/environment/name.rb
@@ -12,12 +12,12 @@ module R10K
       INVALID_CHARACTERS = %r[\W]
 
       def initialize(name, opts)
-        @name   = name
-        @opts   = opts
-
         @source  = opts[:source]
         @prefix  = opts[:prefix]
         @invalid = opts[:invalid]
+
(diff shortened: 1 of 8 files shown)


mwaggett:
@reidmv What prompted this? My main concern is mostly just around adding yet _more_ configuration for a project that already isn't well-understood and isn't getting a ton of love.

reidmv:
@mwaggett This was prompted by an [internal Slack conversation](https://puppet.slack.com/archives/CF3DYNJ3A/p1614090002025000) relating to defining best practices setting customers up for success in the context of CD4PE.

Gist of it: trying to define a best practice for Git branch naming in the context of CD4PE, needing to find a way to balance

1. CD4PE’s today-imposed requirement that deployment state is exposed in Git branches
2. Our desire to give normalized Git workflow for users, with deployment state info intuitively corralled and user-ignorable
3. The literal Puppet environment names also being as intuitive as possible, since they are exposed to users in a number of GUIs and APIs

However, that conversation is ongoing and needs input from the CD4PE product owner because r10k supporting this doesn't actually mean CD4PE supports it. So, totally fine to put this PR "on hold" for awhile while we figure all of that out.

mwaggett:
okay cool, thanks for the context! After reading that thread, I would agree that we need product input before deciding if this should go in. cc @ccaum @npwalker

ccaum:
Thanks, Reid. My initial take is that this is very useful for separating the CD for PE managed branches in git vs the human readable Puppet environments and I'd like to see it implemented. However, it's not a great user setup experience and only something I'd expect professional services to setup for a user. Hopefully in the future we can remove the need for this setting altogether.

reidmv:
@ccaum totally agree on wanting to make it unnecessary—in r10k land, the `exec` source type is more oriented towards that "big" step forward, while this feature if implemented would be enabling more of a small, iterative polish kind of thing.

Notably, while the motivation for this came from CD4PE, it seems like it could improve OSP workflows too, and may be desirable for that reason even if CD4PE takes the big step forward and doesn't ever need something like this.

Thoughts on this as a stand-alone r10k OSP-oriented capability?

lib/r10k/environment/name.rb
```diff
@@ -71,8 +71,24 @@ def dirname
 
       private
 
-      def derive_prefix(source,prefix)
+      def derive_name(name, strip_component)
+        if strip_component == nil
```
justinstoller: Is `false` a value you could see users put in here? Should we instead handle "falsey" values?

lib/r10k/environment/name.rb
```diff
@@ -71,8 +71,24 @@ def dirname
...
+          name
+        elsif %r{^/.*/$}.match(strip_component)
+          regex = Regexp.new(strip_component[1..-2])
+          name.gsub(regex, '')
+        elsif name.start_with?(strip_component)
+          name[strip_component.size..-1]
+        else
+          name
+        end
+      rescue TypeError, NoMethodError
```
justinstoller: IMO, it's simpler to test the class of the input then doing a rescue afterwards.

reidmv: Good call. Funny how code structure sometimes reveals developer (my) chronological thought process, and highlights afterthoughts 😅 . Pushing improvement.

lib/r10k/environment/name.rb
reidmv: Yeah, that would make sense. Updated to handle "falsey" instead of nil.

npwalker (approved):
+1 Carl likes it so I'm good with it

justinstoller (approved):

PR #1130, opened 2021-03-04
Make setopts warn on conflicting args

reidmv:
r10k may have several different ways to specify basically the same thing. E.g., in the git module type, r10k accepts `:branch`, `:commit`, and `:ref`, all of which just set the `@desired_ref` instance variable.

When the user intent is not clear because the user has used multiple equivalent parameters, log a warning, at least if `R10K::Util::Setopts` is being used.

This should more ideally raise an exception, but that would be a behavior change.


justinstoller (approved):

mwaggett:
didn't get a chance to look at the PR, but this should've had a changelog entry with it

PR #1131, opened 2021-03-04
Standard module options

reidmv:
This PR is primarily designed to sanitize the schema used by the Hash source type to define environments and modules, adopting a standard way of describing content (environments, modules) which is not implementation-dependent, the way today's modules especially are.

The idea is that when specifying content in a data hash, it should be possible to specify it using these standard keys.

```yaml
---
object-name:
  type: <string> # E.g. "git", "forge", "svn"
  source: <string> # E.g. "git://server.tld/project/repo.git", "https://forge.puppet.com", etc.
  version: <string> # E.g. "main", "06a32bc", "6.1.0", etc.
```

This PR assumes #1130 exists to improve the visibility of what happens if users use conflicting ways of specifying values like version, such as by using both `version` and an implementation-specific key like `git`, but does not depend on that improvement.

This PR homogenizes how options are parsed for many objects by standardizing on the use of `R10K::Util::Setopts`.

See the [rich diff of the doc/dynamic-environments/configuration.mkd](https://github.com/puppetlabs/r10k/pull/1131/files?short_path=5a41a0b#diff-5a41a0b389c98bce0313a7b6e74d1dcb2a7dfb8e09d5c7f983c33b9fb46c9ec6) page for the best visual of what this change accomplishes for consumers.

This improvement takes a UX step in the direction of r10k being able to support other types of environment sources and module sources, such as e.g. tarballs from Artifactory or S3.

This PR should be backwards compatible, without behavior changes, with one exception. Some object types did not validate schema in any way for keys passed into them. `R10K::Util::Setopts` does validate schema, and so garbage keys that would have been ignored previously may now cause an error.

### DIFF
--- a/doc/puppetfile.mkd
+++ b/doc/puppetfile.mkd
@@ -109,11 +109,19 @@ latest version available.
 
     mod 'puppetlabs/apache', :latest
 
+An explicit type and/or version can be specified using the standard interface,
+`:type` and `:version`. The `:source` parameter is not supported for individual
+forge modules and will be ignored.
+
+    mod 'puppetlabs/apache',
+      type:    'forge',
+      version: '6.0.0'
+
 ### Git
 
 Git repositories that contain a Puppet module can be cloned and used as modules.
 When Git is used, the module version can be specified by using `:ref`, `:tag`,
-`:commit`, and `:branch`.
+`:commit`, `:branch`, or the standard interface parameter `:version`.
 
 When a module is installed using `:ref`, r10k uses some simple heuristics to
 determine the type of Git object that should be checked out. This can be used
@@ -153,6 +161,13 @@ mod 'apache',
 mod 'apache',
   :git    => 'https://github.com/puppetlabs/puppetlabs-apache',
   :branch => 'docs_experiment'
+
+# Install puppetlabs/apache and use standard interface parameters pinned to the
+# '2098a17' commit.
+mod 'puppetlabs-apache',
+  type:    'git',
+  source:  'https://github.com/puppetlabs/puppetlabs-apache',
+  version: '2098a17'
 ```
 
 #### Control Repo Branch Tracking
@@ -195,8 +210,8 @@ the latest version available in the main SVN repository.
     mod 'apache',
       :svn => 'https://github.com/puppetlabs/puppetlabs-apache/trunk'
 
-If an SVN revision number is specified with `:rev` (or `:revision`), that
-SVN revision will be kept checked out.
+If an SVN revision number is specified with `:rev`, `:revision`, or `:version`,
+that SVN revision will be kept checked out.
 
     mod 'apache',
       :svn => 'https://github.com/puppetlabs/puppetlabs-apache/trunk',
@@ -205,6 +220,11 @@ SVN revision will be kept checked out.
     mod 'apache',
       :svn      => 'https://github.com/puppetlabs/puppetlabs-apache/trunk',
       :revision => '154'
+
+    mod 'apache',
+      type:    'svn',
+      source:  'https://github.com/puppetlabs/puppetlabs-apache/trunk',
+      version: '154'
 
 If the SVN repository requires credentials, you can supply the `:username` and
 `:password` options.
--- a/lib/r10k/environment/base.rb
+++ b/lib/r10k/environment/base.rb
@@ -43,7 +43,7 @@ class R10K::Environment::Base
     @basedir = basedir
     @dirname = dirname
     @options = options
-    @puppetfile_name = options[:puppetfile_name]
+    @puppetfile_name = options.delete(:puppetfile_name)
 
     @full_path = File.join(@basedir, @dirname)
     @path = Pathname.new(File.join(@basedir, @dirname))
--- a/lib/r10k/environment/git.rb
+++ b/lib/r10k/environment/git.rb
@@ -27,6 +27,8 @@ class R10K::Environment::Git < R10K::Environment::WithModules
   #   @return [R10K::Git::StatefulRepository] The git repo backing this environment
   attr_reader :repo
 
+  include R10K::Util::Setopts
+
   # Initialize the given Git environment.
   #
   # @param name [String] The unique name describing this environment.
@@ -38,8 +40,21 @@ class R10K::Environment::Git < R10K::Environment::WithModules
   # @param options [String] :ref The git reference to use for this environment
   def initialize(name, basedir, dirname, options = {})
     super
-    @remote = options[:remote]
-    @ref    = options[:ref]
+    setopts(options, {
+      # Standard option interface
+      :version => :ref,
+      :source  => :remote,
+      :type    => ::R10K::Util::Setopts::Ignore,
+
+      # Type-specific options
+      :ref     => :self,
+      :remote  => :self,
+
+    }, raise_on_unhandled: false)
+    # TODO: in r10k 4.0.0, a major version bump, stop allowing garbage options.
+    # We only allow them now, here, on this object, because prior to adopting
+    # setopts in the constructor, this object type didn't do any validation
+    # checking of options passed, and would permit garbage parameters.
 
     @repo = R10K::Git::StatefulRepository.new(@remote, @basedir, @dirname)
   end
--- a/lib/r10k/environment/svn.rb
+++ b/lib/r10k/environment/svn.rb
@@ -44,8 +44,17 @@ class R10K::Environment::SVN < R10K::Environment::Base
   # @option options [String] :password The SVN password
   def initialize(name, basedir, dirname, options = {})
     super
+    setopts(options, {
+      # Standard option interface
+      :source  => :remote,
+      :version => :expected_revision,
+      :type    => ::R10K::Util::Setopts::Ignore,
 
-    setopts(options, {:remote => :self, :username => :self, :password => :self, :puppetfile_name => :self })
+      # Type-specific options
+      :remote => :self,
+      :username => :self,
+      :password => :self,
+    })
 
     @working_dir = R10K::SVN::WorkingDir.new(Pathname.new(@full_path), :username => @username, :password => @password)
   end
@@ -61,7 +70,7 @@ class R10K::Environment::SVN < R10K::Environment::Base
     if @working_dir.is_svn?
       @working_dir.update
     else
-      @working_dir.checkout(@remote)
+      @working_dir.checkout(@remote, @expected_revision)
     end
     @synced = true
   end
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -13,7 +13,7 @@ class R10K::Module::Forge < R10K::Module::Base
   R10K::Module.register(self)
 
   def self.implement?(name, args)
-    !!(name.match %r[\w+[/-]\w+]) && valid_version?(args)
+    (args.is_a?(Hash) && args[:type].to_s == 'forge') || (!!(name.match %r[\w+[/-]\w+]) && valid_version?(args))
   end
 
   def self.valid_version?(expected_version)
@@ -32,13 +32,24 @@ class R10K::Module::Forge < R10K::Module::Base
 
   include R10K::Logging
 
-  def initialize(title, dirname, expected_version, environment=nil)
+  include R10K::Util::Setopts
+
+  def initialize(title, dirname, opts, environment=nil)
     super
+
+    if opts.is_a?(Hash)
+      setopts(opts, {
+        # Standard option interface
+        :version => :expected_version,
+        :source  => ::R10K::Util::Setopts::Ignore,
+        :type    => ::R10K::Util::Setopts::Ignore,
+      })
+    else
+      @expected_version = opts || current_version || :latest
+    end
 
     @metadata_file = R10K::Module::MetadataFile.new(path + 'metadata.json')
     @metadata = @metadata_file.read
-
-    @expected_version = expected_version || current_version || :latest
     @v3_module = PuppetForge::V3::Module.new(:slug => @title)
   end
 
--- a/lib/r10k/module/svn.rb
+++ b/lib/r10k/module/svn.rb
@@ -7,7 +7,7 @@ class R10K::Module::SVN < R10K::Module::Base
   R10K::Module.register(self)
 
   def self.implement?(name, args)
-    args.is_a? Hash and args.has_key? :svn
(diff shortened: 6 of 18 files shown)


lib/r10k/module/svn.rb
```diff
@@ -47,7 +52,7 @@ def self.implement?(name, args)
   def initialize(name, dirname, opts, environment=nil)
     super
 
-    setopts(opts, INITIALIZE_OPTS)
+    setopts(@args, INITIALIZE_OPTS)
```
justinstoller: IMO, stylistically it's better to just use the parameter name you're passed into, even it is effectively the same as an instance variable created elsewhere. I don't know if that's actually

justinstoller:
Overall, I really like it.

However, I'm also really concerned about there being garbage in folks repository definitions. I haven't looked far enough into it to know if that concern is valid though.

lib/r10k/module/svn.rb
reidmv: Yeah. The main reason this goes to `@args` is the sanitization performed on the instance variable [here](https://github.com/puppetlabs/r10k/pull/1131/files#diff-db1c390dda7d3b2f14abcee49a787a9a31ffa9a41d2a0cbea53e35fb65b86773R55), to remove a common key "type" so that individual `setopts()` calls don't have to care if it's present.

One alternative option could be to provide `setopts()` with the ability to ignore certain keys. If we did that, we'd have to ignore it everywhere, but also wouldn't have to sanitize `@args`. E.g.,

```
setopts(options, {
  # Standard option interface
  :source  => :remote,
  :version => :expected_revision,
  :type => :ignore

  # Type-specific options
  :remote => :self,
  :username => :self,
  :password => :self,
  :puppetfile_name => :self,
})
```

Think that would be better?

reidmv: Pushed a new commit with this change. Seems more straightforward, and therefore better.

reidmv:
@binford2k can you take a look at this and @justinstoller's comment, let me know what you think?

For now I'll whip up a commit permitting and warning on garbage parameters.

binford2k:
@reidmv is there a pressing reason to disallow garbage parameters? Warn/permit seems like a reasonable non-breaking solution, with the expectation that next major release would make it invalid.

reidmv:
@binford2k @justinstoller I've added some commits that permit R10K::Util::Setopts to warn on unhandled options instead of raising, and R10K::Environment::Git now makes use of that. It looks to me like that was the only object, actually, that wasn't already validating input options somehow. In any case though, it's now possible to permit unhandled options when necessary to preserve behavior.

justinstoller:
@mwaggett had mentioned in the pre-req PR that we should be updating the changelog as we go - we should do that here. Did you want to also review this PR @mwaggett ?

mwaggett:
yes, this should also include a changelog update (and add one for the other PR while you're at it, please 😊 ).

I was planning to review this eventually, but don't wait on me if you think it's already received sufficient review.

reidmv:
New commit pushed with additional docs updates, changelog additions

doc/puppetfile.mkd
```diff
@@ -109,11 +109,20 @@ latest version available.
...
     mod 'puppetlabs/apache', :latest
 
+An explicit type and/or version can be specified using the standard interface,
+`:type` and `:version`. 
+
+    mod 'puppetlabs/apache',
+      type:    'forge',
+      version: '6.0.0'
+
+Note that at present the `:source` parameter is permitted, but not implemented.
```
mwaggett: I'm not sure what you mean by this.

lib/r10k/environment/base.rb
```diff
@@ -43,7 +43,7 @@ def initialize(name, basedir, dirname, options = {})
     @basedir = basedir
     @dirname = dirname
     @options = options
-    @puppetfile_name = options[:puppetfile_name]
+    @puppetfile_name = @options.delete(:puppetfile_name)
```
mwaggett: Would it make more sense to do `options.delete` instead of deleting from the global `@options`? It probably doesn't really matter, but I guess I would expect `@options` to contain _all_ the options, while `options` is only used in the scope of this method, so who cares if we delete from it.

lib/r10k/environment/git.rb
```diff
@@ -38,8 +40,16 @@ class R10K::Environment::Git < R10K::Environment::WithModules
   # @param options [String] :ref The git reference to use for this environment
   def initialize(name, basedir, dirname, options = {})
     super
-    @remote = options[:remote]
-    @ref    = options[:ref]
+    setopts(options, {
+      # Standard option interface
+      :version => :ref,
+      :source  => :remote,
+      :type    => ::R10K::Util::Setopts::Ignore,
```
mwaggett, line 47: To make sure I understand this (since I don't have a ton of r10k experience) - is `:type` just used to determine which of these subclasses to use, and that's why we're ignoring it here?

lib/r10k/environment/git.rb
```diff
@@ -38,8 +40,16 @@ class R10K::Environment::Git < R10K::Environment::WithModules
...
+
+      # Type-specific options
+      :ref     => :self,
+      :remote  => :self,
+    }, raise_on_unhandled: false)
```
mwaggett, line 53: Why don't we want to raise on unhandled options here?

lib/r10k/module/forge.rb
```diff
@@ -32,13 +32,24 @@ def self.valid_version?(expected_version)
 
   include R10K::Logging
 
-  def initialize(title, dirname, expected_version, environment=nil)
+  include R10K::Util::Setopts
+
+  def initialize(title, dirname, opts, environment=nil)
```
mwaggett, line 37: I don't love that the third arg can now be two different things, but I don't know if I have a better idea right now. and this repo is probably full of that kind of thing, huh..

lib/r10k/module/svn.rb
```diff
@@ -36,18 +36,21 @@ def self.implement?(name, args)
...
-    :password => :self
-  }
-
   def initialize(name, dirname, opts, environment=nil)
     super
-
-    setopts(opts, INITIALIZE_OPTS)
+    setopts(opts, {
+      # Standard option interface
+      :source   => :url,
```
mwaggett, line 43: gah why do we use `:url` here and `:remote` in the environment version 😭 
(I know it was already like that, but still 😭)

spec/unit/module_spec.rb
```diff
@@ -3,20 +3,36 @@
 
 describe R10K::Module do
   describe 'delegating to R10K::Module::Git' do
-    it "accepts args {:git => 'git url}" do
-      obj = R10K::Module.new('foo', '/modulepath', :git => 'git url')
-      expect(obj).to be_a_kind_of(R10K::Module::Git)
+    [
+      ['foo', {git: 'git url'}],
+      ['foo', {type: 'git', source: 'git url'}],
+    ].each do |scenario|
```
mwaggett: If you made this an array of maps, you could refer to the name and args separately, instead of using array indices.

spec/unit/module_spec.rb
```diff
@@ -3,20 +3,36 @@
...
+      ['bar/quux', {type: 'forge', version: '8.0.0'}],
     ].each do |scenario|
       it "accepts a name matching #{scenario[0]} and args #{scenario[1].inspect}" do
-        expect(R10K::Module.new(scenario[0], '/modulepath', scenario[1])).to be_a_kind_of(R10K::Module::Forge)
+        obj = R10K::Module.new(scenario[0], '/modulepath', scenario[1])
+        expect(obj).to be_a_kind_of(R10K::Module::Forge)
+        expect(obj.send(:instance_variable_get, :'@expected_version')).to eq('8.0.0')
       end
     end
   end
```
mwaggett, line 47: Should there be svn tests? I notice there aren't any in this file and you didn't add any in `spec/unit/module/svn_spec.rb`.

spec/unit/module/forge_spec.rb
```diff
@@ -23,6 +23,12 @@
     end
   end
 
+  describe "implementing the generic hash spec" do
+    it "should implement {type: forge}" do
```
mwaggett, line 27: Should this test include all the generic hash keys?

spec/unit/module/forge_spec.rb
```diff
@@ -23,6 +23,12 @@
     end
   end
 
+  describe "implementing the generic hash spec" do
```
mwaggett: Do we need to test this for other module types too?

doc/puppetfile.mkd
reidmv: I think based on that, I'm going to remove it, and just say that a `:source` parameter isn't supported for individual forge modules. At present, r10k will only let you specify a Forge source once, globally. We could add individual forge sources, but there isn't a use case reason to right now.

Edit: oh wait, I remember now why it was there. Because the whole point of this update is to make it possible to use a uniform schema for these objects. So, because `:source` is part of that schema, even though we ignore it for this type of module, we should allow a value to be passed for it.

lib/r10k/environment/base.rb
reidmv: Ooh, yes, and drawing attention to this caught another problem, 'cause the derived classes all use `options` now instead of `@options` ... will add a fix for this.

lib/r10k/environment/git.rb
reidmv: It used to be for some objects, and after this PR is across the board, yeah. The r10k user interface is hella inconsistent. The big objective of this commit is to at least have a uniform common denominator interface, even though we have to stay backwards compatible with the old inconsistent stuff.

lib/r10k/environment/git.rb
reidmv: Because previously, git-based environments would let you pass garbage options. So for backwards compatibility, we still let you pass garbage options, here, for this one object type. 🤦

doc/puppetfile.mkd
mwaggett: ah right, I forgot to circle back to this after reading the rest of the PR and understood things better.
yeah, I think it would be clearer to say something like
> Note: The `:source` parameter can be specified for Forge modules, but is currently ignored.

lib/r10k/module/forge.rb
reidmv: Yuuuup.

I also don't love that what _kind_ of arguments you gave `#mod` in a Puppetfile or a data hash, previously, was actually the main thing used to determine what _type_ of object you were specifying. Yay! But we need to keep it for backwards compatibility.

spec/unit/module_spec.rb
reidmv: May as well. Adding.

spec/unit/module_spec.rb
reidmv: True. Adding a minor re-work of these tests.

lib/r10k/environment/git.rb
mwaggett: oof. Want to add a comment to that effect? Then ideally we could remove that if we ever release a new major version.

lib/r10k/module/forge.rb
mwaggett: Yay! 🙃

spec/unit/module/forge_spec.rb
reidmv: Not sure. I kind of trust `#setopts` to do its thing, at which point tests mostly sanity check that the `#setopts` definition of options is correct. A meaningful test would basically be a copy of that logic.

spec/unit/module/forge_spec.rb
reidmv: Might as well. Adding them.

lib/r10k/environment/git.rb
reidmv: Will do

spec/unit/module/forge_spec.rb
mwaggett: That's fair. Would it make sense for this to go in the `base_spec` then? Or would that not work?

doc/puppetfile.mkd
```diff
@@ -109,11 +109,19 @@ latest version available.
 
     mod 'puppetlabs/apache', :latest
 
+An explicit type and/or version can be specified using the standard interface,
+`:type` and `:version`. The `:source` parameter is not supported for individual
+forge modules.
```
mwaggett:
```suggestion
forge modules and will be ignored.
```

lib/r10k/environment/svn.rb
```diff
@@ -44,8 +44,21 @@ class R10K::Environment::SVN < R10K::Environment::Base
...
+      :source  => :remote,
+      :version => :expected_revision,
+      :type    => ::R10K::Util::Setopts::Ignore,
 
-    setopts(options, {:remote => :self, :username => :self, :password => :self, :puppetfile_name => :self })
+      # Type-specific options
+      :remote => :self,
+      :username => :self,
+      :password => :self,
+      :puppetfile_name => :self,
```
mwaggett: I take it this should be removed now, if it's being handled in the base class.

spec/unit/module/forge_spec.rb
mwaggett: Not sure if you wanna update this description also

mwaggett:
Anything you wanna squash @reidmv ? I take it this probably shouldn't all be a single commit, but there are pieces that could be squashed.

reidmv:
Yeah, lemme clean it up a bit. Just a moment.

spec/unit/module/forge_spec.rb
reidmv: I think I put this test here specifically because the forge type was slightly more complicated than just calling setopts. The module_spec.rb already does basic initialization validation for several types.

mwaggett: ah okay!

reidmv:
Rebased into something resembling a clean history. There is one final cleanup change mixed in here as well. It relates to how `:puppetfile_name` is handled. Some tests added to help verify that passing options managed in base classes (such as `:puppetfile_name`) all works correctly.

mwaggett (approved):
Not sure if you wanted to squash in the various pieces of that last improvements commit, but otherwise looks good!

@justinstoller did you want to take another look?

reidmv:
Alright, totally squashed, ended with four total commits: 1 for setopts enhancements, 1 for module interface additions, 1 for environment interface additions, and a separate commit for docs updates.

PR #1132, opened 2021-03-09
(wip) don't deploy spec dirs

mwaggett:
```
```

### DIFF
--- a/lib/r10k/module/base.rb
+++ b/lib/r10k/module/base.rb
@@ -1,5 +1,6 @@
 require 'r10k/module'
 require 'puppet_forge'
+require 'r10k/logging'
 
 # This class defines a common interface for module implementations.
 class R10K::Module::Base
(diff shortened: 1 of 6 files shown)


lib/r10k/module/base.rb
```diff
@@ -63,7 +64,9 @@ def full_path
   # Synchronize this module with the indicated state.
   # @abstract
   def sync(opts={})
-    raise NotImplementedError
+    spec_dir = File.join(path, 'spec')
```
jarretlavallee, line 67: This may change the scope of this PR, but would it be possible to make this a configuration option that defaults to `spec`? It would be nice to have a module blocklist that can allow users to specify which directories will be deleted. I could see adding additional items such as `vendor`, or `.bundle` to this list.

mwaggett: yeah, that's definitely an idea that's been floating around in my head. A community member also brought up wanting to exclude other files. You're right though, it does change the scope of this work though, so I'll need to talk to Product about this. /cc @npwalker @puppetlabs/puppetserver-maintainers

PR #1134, opened 2021-03-17
(RK-308) Use alternate forge from Puppetfile

cdenneen:
- R10K setting to allow_override to use forge setting
 - Tests updated for using alternate forge
 - Rebased off master

- Summary of changes.  Rebased against master - Replaces old PR #779

### DIFF
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -134,6 +134,28 @@ class Puppetfile
     @loaded_content[:modules]
   end
 
+  # @param [String] forge
+  def set_forge(forge)
+    unless @forge.nil?
+      raise R10K::Error.new(_('You cannot declare multiple \'forge\' settings in the same Puppetfile'))
+    end
+    @forge = forge
+  end
+
+  # @param [Array<String>] modules
+  def validate_no_duplicate_names(modules)
+    dupes = modules
+            .group_by { |mod| mod.name }
+            .select { |_, v| v.size > 1 }
+            .map(&:first)
+    unless dupes.empty?
+      msg = _('Puppetfiles cannot contain duplicate module names.')
+      msg += ' '
+      msg += _("Remove the duplicates of the following modules: %{dupes}" % { dupes: dupes.join(' ') })
+      raise R10K::Error.new(msg)
+    end
+  end
+
   # @see R10K::ModuleLoader::Puppetfile#add_module for upcoming signature changes
   def add_module(name, args)
     @loader.add_module(name, args)
--- a/lib/r10k/settings.rb
+++ b/lib/r10k/settings.rb
@@ -128,7 +128,12 @@ module R10K
         }),
         Definition.new(:authorization_token, {
           :desc => "The token for Puppet Forge authorization. Leave blank for unauthorized or license-based connections."
-        })
+        }),
+        Definition.new(:allow_override, {
+          :desc => "Whether or not to allow the Forge URL to be overridden by a 'forge' directive inside an environment Puppetfile.",
+          :default => lambda { false },
+          :normalize => lambda { |v| v.to_s.strip == 'true' },
+        }),
       ])
     end
 
--- a/r10k.yaml.example
+++ b/r10k.yaml.example
@@ -110,3 +110,8 @@ forge:
   # The 'baseurl' setting indicates where Forge modules should be installed
   # from. This defaults to 'https://forgeapi.puppetlabs.com'
   #baseurl: 'https://forgemirror.example.com'
+
+  # The 'allow_override' setting controls whether or not to allow the Forge
+  # 'baseurl' to be overridden by a 'forge' directive inside a Puppetfile.
+  # This defaults to 'false'.
+  #allow_override: true
--- a/spec/fixtures/unit/puppetfile/alternate-forge/Puppetfile
+++ b/spec/fixtures/unit/puppetfile/alternate-forge/Puppetfile
@@ -1,0 +1,3 @@
+forge 'http://forge.example.com/'
+
+mod 'puppetlabs/stdlib'
--- a/spec/unit/puppetfile_spec.rb
+++ b/spec/unit/puppetfile_spec.rb
@@ -228,6 +228,14 @@ describe R10K::Puppetfile do
       loaded_module = subject.modules.first
       expect(loaded_module.version).to eq('foo')
     end
+
+    it "accepts alternate forge" do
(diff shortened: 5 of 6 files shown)


lib/r10k/puppetfile.rb
```diff
@@ -88,6 +87,14 @@ def loaded?
     @loaded
   end
 
+  # @param [String] forge
+  def set_forge(forge)
```
Magisus, line 138: Looks like this same method is also declared down below on line 113, which will override this. I think you'd want to add to that definition instead of duplicating it here.

lib/r10k/puppetfile.rb
```diff
@@ -88,6 +87,14 @@ def loaded?
...
+    unless @forge.nil?
+      raise R10K::Error.new(_('You cannot declare multiple \'forge\' settings in the same Puppetfile'))
+    end
+    @forge = forge
```
Magisus, line 142: There was a comment in the previous PR about needing to make sure this is set to the default from the forge module if `allow_overrides` is false. Is that actually necessary? It kind of looks like the logic to just ignore this field in the install action would take care of that case? That is, when overriding is not allowed, this field is entirely ignored. Is that accurate?

lib/r10k/puppetfile.rb
cdenneen: @Magisus looks like with the install.rb and settings.rb changes this might not be necessary anymore as you pointed out.

lib/r10k/puppetfile.rb
cdenneen: @Magisus if the part of this commit for this file isn't necessary then yes I believe this wouldn't be needed anymore either.

npwalker:
@cdenneen what is the high level use case you are solving with this change?

cdenneen:
@npwalker utiizing the forge url from the top of 99% of the Puppetfiles in the wild. While you can globally update repository at the server level it’s useful to do as per env override and even with an r10k run without having to configure baseurl in the r10k configuration. 

As far as opt-in/out currently it’s not “read” so if defaulting to the default forgeapi if missing and with the 99% of the ones out in the wild it should still appear as an “opt-in” anyway. For those running local forges for compliance, security, testing reasons should be allowed to leverage this existing configuration option that currently does nothing.

npwalker:
@cdenneen can you provide a specific example?  I'm familiar with users having an internal artifactory that they serve modules from instead of the forge but that's served by the current setting.  

Is this multiple teams wanting to use different forges?  or is this a testing workflow vs what you'd use to deploy to a puppet server in production?

cdenneen:
Currently when you deploy and Environment the "forge" string from a Puppetfile isn't actually used. You can configure the module repository for the Puppet Server to use which would pull all "forge modules" from whatever forge is configured. This would allow ENV_A to use the default forgeapi.puppetlabs.com and ENV_B to use a specified internal forge. You could also have a workflow that some sort of "dev" or pre-prod might publish to internal forge which needs to pull from before going to production which might publish to the public forge.

Currently you must update the r10k setting globally to use a particular forge. When you use `spec_prep` the `.fixtures.yml` allows you to specify a flag to use specific module_repository but there the `forge` setting inside Puppetfile isn't actually used.

Magisus:
Fwiw, I think those use cases make sense. However, I think I would like to see a slightly more specific name for the setting, maybe `use_puppetfile_url` or something.

Unfortunately the changes in here are now quite out of date with all the refactors we've been doing around Puppetfile loading. And there are more coming shortly in https://github.com/puppetlabs/r10k/pull/1200/. I'm alright with this approach generally, but I think we'll probably want to rework it once that PR has landed to use the `ModuleLoader::Puppetfile` instead. Happy to help with that if you like.

cdenneen:
@Magisus yes please. The current PR should have write access to you so you can update accordingly when ready. There was already a set_forge parameter so just utilizing that vs creating a new one. Also from what I can tell onceover actually do a `r10k puppetfile install` so reading the forge line from the top of the Puppetfile would definitely be useful.

My only thing that has come up recently and obviously it would expand past r10k Puppetfile `forge` line is authenticated private forge endpoints. We just actually moved from a private internal JFrog Artifactory to a JFrog Cloud SaaS so the `anonymous` access to that had to go away and we are using a generated Access Token to connect to the forge from `module_repository`, `.fixtures.yml` or the `forge: baseurl` in r10k configuration. It would be useful to determine if there was better way to handle auth via user/pass (Basic Authentication) or Bearer Token? adding https://USER:TOKEN@endpoint is a bit sloppy and lacking in security. As I said it stems outside this current issue but possibly it is something that could gain some internal discussion there and possible affect how you implement the `use_puppetfile_url` change you are proposing?

Thanks

Magisus:
Great, thanks for the context! We recently started supporting Bearer tokens for Forge access (see https://github.com/puppetlabs/r10k/pull/1181). Obviously, that setting is global and so couldn't be used per-environment to go along with the URLs in the Puppetfile, but the underlying machinery is all there if we can find a good way to specify it (and maybe this helps you in the meantime?).

Some ideas:
* Allow a bearer token to be specified on the command line. This could follow the same semantics as git OAuth tokens use in the CLI (read from a file or from `stdin`). The downside here is it probably wouldn't work right for the "deploy all" scenario.
* Add another arg to the forge declaration in the Puppetfile, that contains a path pointing to a token file on the machine that will be running r10k, which we will read from. While this fits the use case the best, I think people are fairly resistant to adding more magic to the Puppetfile...

But I agree that would probably be best done as a separate issue/PR.

I'll circle back around to updating this once #1200 goes in.

cdenneen:
@Magisus ok sounds good. I also tried my best to rebase from current main. Feel free to do the same after #1200 merges

mwaggett:
okay https://github.com/puppetlabs/r10k/pull/1200 is in, so this can be updated to use the `R10K::ModuleLoader::Puppetfile` class instead of `R10K::Puppetfile`.

cdenneen:
> okay #1200 is in, so this can be updated to use the `R10K::ModuleLoader::Puppetfile` class instead of `R10K::Puppetfile`.

I have rebased but haven't made the class change as this is larger than just my PR. Feel free to update my PR if you'd like it to encompass these changes as well.

Magisus:
Okay, I reworked this here: https://github.com/puppetlabs/r10k/pull/1214. Going to close this, thanks so much for getting the start on this @cdenneen ! Let me know if you have any thoughts about my refactored version, and if you want to test it out and make sure it fulfills your needs, that would be excellent.

PR #1142, opened 2021-04-07
(maint) Use docker-compose style testing

Iristyle:
- All other suites use docker-compose for testing, so do the same for
   consistency across all Puppet containers

PR #1143, opened 2021-04-07
(maint) Use docker-compose style testing

Iristyle:
New version of #1142 since I couldn't repush my own branch for some reason.

 - All other suites use docker-compose for testing, so do the same for
   consistency across all Puppet containers

 - Remove repetitive spec code


jpartlow (approved):
Looks equivalent.

PR #1145, opened 2021-04-13
(CODEMGMT-1415) SPIKE assume modules unchanged locally

justinstoller:
```
```

### DIFF
--- a/lib/r10k/mock_module.rb
+++ b/lib/r10k/mock_module.rb
@@ -1,0 +1,65 @@
+require 'puppet_forge'
+
+class R10K::MockModule
+
+  # @!attribute [r] title
+  #   @return [String] The forward slash separated owner and name of the module
+  attr_reader :title
+
+  # @!attribute [r] name
+  #   @return [String] The name of the module
+  attr_reader :name
+
+  # @param [r] dirname
+  #   @return [String] The name of the directory containing this module
+  attr_reader :dirname
+
+  # @!attribute [r] owner
+  #   @return [String, nil] The owner of the module if one is specified
+  attr_reader :owner
+
+  # @!attribute [r] path
+  #   @return [Pathname] The full path of the module
+  attr_reader :path
+
+  # @!attribute [r] version
+  #   @return [Pathname] The version, if it can be statically determined from args
+  attr_reader :version
+
+  # @param title [String]
+  # @param dirname [String]
+  # @param args [Array]
+  def initialize(title, dirname, args, environment=nil)
+    @title   = PuppetForge::V3.normalize_name(title)
+    @dirname = dirname
+    @owner, @name = parse_title(@title)
+    @path = Pathname.new(File.join(@dirname, @name))
+    @version = find_version(args)
+  end
+
+  private
+
+  def find_version(args)
+    if args.is_a?(String)
+      args
+    elsif args.is_a?(Hash)
+      if args[:type] == 'forge'
+        args[:version]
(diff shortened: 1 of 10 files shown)


lib/r10k/mock_module.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  # @param title [String]
+  # @param dirname [String]
+  # @param args [Array]
+  def initialize(title, dirname, args, environment=nil)
+    @title   = PuppetForge::V3.normalize_name(title)
+    @dirname = dirname
+    @owner, @name = parse_title(@title)
+    @path = Pathname.new(File.join(@dirname, @name))
+    @version = find_version(args)
```
Magisus, line 37: Do we need all these fields? Is there some hidden contract that requires them?

justinstoller: We need the version and name in the current implementation (which requires the title). I thought we might need the path but I don't think so. This file was just a copy and pared down version of the module base class.

justinstoller:
This is still a spike, but I'm opening it up as "ready for review" so that those subscribing can learn about it. Especially, regarding the efficacy of this change and the UX around `--assume-unchanged`. /cc @reidmv

lib/r10k/puppetfile.rb
```diff
@@ -144,7 +162,9 @@ def add_module(name, args)
     @managed_content[install_path] = Array.new unless @managed_content.has_key?(install_path)
     @managed_content[install_path] << mod.name
 
-    @modules << mod
+    # We want to manage the content (and not purge it) even if we do not want
+    # to sync modules that we assume unchanged.
+    no_change ? @modules : @modules << mod
```
reidmv, line 167: I worry about focusing effort on the Puppetfile, when I see the future direction being more "modules attached to an environment", with the Puppetfile being kind of a legacy way to attach them.

Is it possible to take something like this approach instead?

1. Add something like `#insync?` and `#insync=` methods to module objects. Also `#insync` to get the detailed value.
2. In the Puppetfile, set `mod.insync = true/false/:unknown` as appropriate. When `:unknown`, `#insync?` returns false.
3. Instead of adding logic to mask the existence of the module here in the Puppetfile code, figure out where the module is actually synced/verified, and put an `... unless mod.insync?` there. I think that might be in [the environment deploy action](https://github.com/puppetlabs/r10k/blob/aab476ab05432c2a7e5361cd68524e19a53b9f74/lib/r10k/action/deploy/environment.rb#L142-L145) (to pass e.g. an `@verify` flag) and/or in the [module sync method](https://github.com/puppetlabs/r10k/blob/aab476ab05432c2a7e5361cd68524e19a53b9f74/lib/r10k/module/base.rb#L63-L67).

That approach would also preserve the ability to "see" all modules attached to an environment, regardless of what you're planning on doing to/with them.

Again, mostly I'm just worried about embedding important logic in the Puppetfile specifically, given that the Puppetfile is only one way to define environment content. At some point a general encapsulation/cleanup of how environments and modules relate would be nice, but in the meantime I think not implementing new stuff as Puppetfile-specific would be a good idea.

justinstoller:
One thing I've noticed with this is that the environment info that is written will no longer include the full module list like it currently does. (when investigating this I also opened this PR to clean up writing that info a little: https://github.com/puppetlabs/r10k/pull/1148)

Magisus:
Not stale 🙂

Magisus:
I think this has been superseded by https://github.com/puppetlabs/r10k/pull/1200/, going to close it. Please reopen if you still need it for something.

PR #1148, opened 2021-04-20
(maint) Do not raise NoMethodError on normal env info writing

justinstoller:
Previously when calling write_environment_info we would call
Module#repo to collect the sha, if it existed. However, only git backed
modules have this method defined. So the write_environment_info method
would cause an exception which it caught and ignored when inspecting any
non-git-backed modules.

This moves to use the Module#properties method which is well defined for
every type of module and has an `:actual` key which matches the value
that the write_environment_info used to gather. We do not collect
`:actual` key values for any module besides those of type `:git` to
match the behavior preivously.

This also adds a test for the failure case of the write_environment_info
method that was previously untested. I have a feeling that this is no
longer needed now that we do not call Module#repo indiscriminantly,
however it is included to err on the safe side.

```
```

### DIFF
--- a/spec/unit/action/deploy/environment_spec.rb
+++ b/spec/unit/action/deploy/environment_spec.rb
@@ -432,9 +432,25 @@ describe R10K::Action::Deploy::Environment do
 
     let(:mock_stateful_repo_1) { instance_double("R10K::Git::StatefulRepository", :head => "123456") }
     let(:mock_stateful_repo_2) { instance_double("R10K::Git::StatefulRepository", :head => "654321") }
-    let(:mock_git_module_1) { instance_double("R10K::Module::Git", :name => "my_cool_module", :version => "1.0", :repo => mock_stateful_repo_1) }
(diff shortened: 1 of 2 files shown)


spec/unit/action/deploy/environment_spec.rb
```diff
@@ -432,9 +432,25 @@ def initialize(path, info)
...
+                      :name       => "my_cool_module",
+                      :properties => {
+                        :type     => :git,
+                        :expected => "1.0",
+                        :actual   => mock_stateful_repo_1.head
+                      })
+    end
+    let(:mock_git_module_2) do
+      instance_double("R10K::Module::Git",
+                      :name       => "my_lame_module",
```
mwaggett: can we use a different word than `lame`? it has an ableist history.

mwaggett:
This looks good to me, other than my one comment.

Magisus (approved):

spec/unit/action/deploy/environment_spec.rb
justinstoller: updated

mwaggett (approved):

PR #1152, opened 2021-04-23
(maint) Update action calling convention

justinstoller:
----
(maint) Use Ruby 2.3's Hash#dig & default arg values

R10K only supports Ruby >= 2.3 which means we can use Hash#dig instead
of Hash#fetch#fetch.

Also, it is never okay (outside of testing) to not pass a Hash for
the Action#new `settings` arg. Because of this, make the setting
argument mandatory.

for the settings, no need for a `nil` default that we later
conditionally override.

---

(maint) Give @argv a more meaningful name when used

Previously, some actions would use the argv values given to them but
never name them semantically. This patch names them "requested_modules"
or "requested_environments" where the content of argv should be one or
more modules or environments respectively.

### DIFF
--- a/lib/r10k/action/deploy/display.rb
+++ b/lib/r10k/action/deploy/display.rb
@@ -43,7 +43,7 @@ module R10K
           puts output.to_yaml
         end
 
-        def source_info(source, argv=[])
+        def source_info(source, requested_environments = [])
           source_info = {
             :name => source.name,
             :basedir => source.basedir,
@@ -52,7 +52,8 @@ module R10K
           source_info[:prefix] = source.prefix if source.prefix
           source_info[:remote] = source.remote if source.respond_to?(:remote)
 
-          env_list = source.environments.select { |env| argv.empty? || argv.include?(env.name) }
+          select_all_envs = requested_environments.empty?
+          env_list = source.environments.select { |env| select_all_envs || requested_environments.include?(env.name) }
           source_info[:environments] = env_list.map { |env| environment_info(env) }
 
           source_info
--- a/lib/r10k/action/base.rb
+++ b/lib/r10k/action/base.rb
@@ -10,7 +10,7 @@ module R10K
 
       attr_accessor :settings
 
-      def initialize(opts, argv, settings = {})
+      def initialize(opts, argv, settings)
         @opts = opts
         @argv = argv
         @settings = settings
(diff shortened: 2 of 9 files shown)


lib/r10k/action/deploy/display.rb
```diff
@@ -52,7 +52,8 @@ def source_info(source, argv=[])
           source_info[:prefix] = source.prefix if source.prefix
           source_info[:remote] = source.remote if source.respond_to?(:remote)
 
-          env_list = source.environments.select { |env| argv.empty? || argv.include?(env.name) }
+          all_envs = requested_environments.empty?
```
mwaggett: (nit) maybe call this `select_all_envs` or something? I want it to be clear that this is a boolean and not a collection of environments.

mwaggett (approved):

lib/r10k/action/deploy/display.rb
justinstoller: yeah, this is one place where I'd love to make a local variable with a `?`.

justinstoller: updated

lib/r10k/action/deploy/display.rb
```diff
@@ -52,7 +52,8 @@ def source_info(source, argv=[])
           source_info[:prefix] = source.prefix if source.prefix
           source_info[:remote] = source.remote if source.respond_to?(:remote)
 
-          env_list = source.environments.select { |env| argv.empty? || argv.include?(env.name) }
+          select_all_envs = requested_environments.empty?
+          env_list = source.environments.select { |env| select_all_envs || request_environments.include?(env.name) }
```
mwaggett:
```suggestion
          env_list = source.environments.select { |env| select_all_envs || requested_environments.include?(env.name) }
```

seems Not Great that none of the tests caught this.. Is there something we could add?

lib/r10k/action/deploy/display.rb
mwaggett: ugh same, it bums me out every time I remember that you can't do that in Ruby

lib/r10k/action/deploy/display.rb
justinstoller: _ouch_  I'll look

justinstoller: I've added a test to check that it can at least meaningfully collect source and environment info.

justinstoller:
rebased, should be good now

mwaggett (approved):

PR #1153, opened 2021-04-23
(maint) Extract Puppetfile accept logic to ContentSynchronizer

justinstoller:
The Puppetfile internally contains logic to sync modules either
concurrently or serially depending on pool size. This would be valuable
to other means of attaching modules to an environment. The first step in
doing so is to extract the sync code into a helper module called
R10K::ContentSynchronizer.

```
```

### DIFF
--- a/lib/r10k/content_synchronizer.rb
+++ b/lib/r10k/content_synchronizer.rb
@@ -1,0 +1,60 @@
+module R10K
+  module ContentSynchronizer
+
+    def self.serial_accept(modules, visitor, loader)
+      visitor.visit(:puppetfile, loader) do
+        modules.each do |mod|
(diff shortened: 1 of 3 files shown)


lib/r10k/content_synchronizer.rb
```diff
@@ -0,0 +1,60 @@
+module R10K
+  module ContentSynchronizer
+
+    def self.serial_accept(modules, visitor, loader, logger)
```
mwaggett: doesn't look like this needs a `logger`

justinstoller: updated

mwaggett (approved):

PR #1155, opened 2021-04-27
(maint) Use a "Cleaner" helper to purge Puppetfile content

justinstoller:
Prior to this Puppetfile content was managed via the Purgeable mixin.
As a long term direction it would be valuable to move the Puppetfile in
a direction of acting on, or contributing to, modules connected to an
environment. In the processing of "dumbing" down the Puppetfile class
this patch moves the responsibility of purging to a helper class
(`R10K::Util::Cleaner`), leaving the Puppetfile class only the
responsibility of determining the Cleaner's inputs.

```
```

### DIFF
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -1,4 +1,5 @@
 require 'r10k/util/setopts'
+require 'r10k/util/cleaner'
 require 'r10k/deployment'
 require 'r10k/logging'
 require 'r10k/action/visitor'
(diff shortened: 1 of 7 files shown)


Magisus:
Do we want to remove this https://github.com/puppetlabs/r10k/blob/main/lib/r10k/puppetfile.rb#L151?

justinstoller:
I don't think so. I kept it there for backwards compatibility in case somewhere else someone is caling `puppetfile.purge!` Not sure if I need to be that concerned about it though...

Magisus (approved):

PR #1156, opened 2021-04-30
(maint) Pass Action configuration through helper objects all the way to Modules

justinstoller:
The visitor pattern can be confusing for new comers and has a considerable maintenance burden. For some in progress features it would be ideal to embed the `visit_module` and `visit_puppetfile` functionality into the modules & environments themselves.

To do so we need the configuration that comes in from the command line and is up-to-now own by the Action classes. This patch seeks to enable the above goal by passing the configuration currently owned by Actions all the way through the helper classes that participate in the visitor pattern down to the modules. Because of the scope of this work, this PR is intentionally naive. It passes _all_ the configuration _all_ the way down to Modules in a very naively shaped data structure.

The hope being that once we have that data everywhere we may need it we can start unwinding some of the visitor usage starting at Modules and working our way up. In doing so, only some values will need to be passed down and the data structure can be tweaked to better represent the values that are really needed, how they are needed.

Two small, potentially user facing changes snuck in while doing this:
1. If calling `r10k deploy display --fetch` the fetched source & environment code will now also be validated.
2. The debug message in `r10k deploy module --environment foo` will display the environment name "foo" in a single element array rather than a simple string

These small changes normalize some trivial behavior between different Actions `visit_*` calls so they may use the same data structures.

It may be best to step through the PR by commit rather than all at once.

### DIFF
--- a/lib/r10k/action/deploy/display.rb
+++ b/lib/r10k/action/deploy/display.rb
@@ -16,7 +16,7 @@ module R10K
             overrides: {
               environments: {
                 preload_environment: @fetch,
-                requested_environments: @argv
+                requested_environments: @argv.map { |arg| arg.gsub(/\W/, '_') }
               },
               modules: {},
               output: {
--- a/lib/r10k/action/deploy/display.rb
+++ b/lib/r10k/action/deploy/display.rb
@@ -9,17 +9,38 @@ module R10K
 
         include R10K::Action::Deploy::DeployHelpers
 
+        def initialize(opts, argv, settings)
+          super
+
+          @settings = @settings.merge({
+            overrides: {
+              environments: {
+                preload_environment: @fetch,
+                requested_environments: @argv
+              },
+              modules: {},
+              output: {
+                format: @format,
+                trace: @trace,
+                detail: @detail
+              },
+              purging: {}
+            }
+          })
+        end
+
         def call
           expect_config!
           deployment = R10K::Deployment.new(@settings)
 
-          if @fetch
+          if @settings[:overrides][:environments][:preload_environments]
             deployment.preload!
+            deployment.validate!
           end
 
-          output = { :sources => deployment.sources.map { |source| source_info(source, @argv) } }
+          output = { :sources => deployment.sources.map { |source| source_info(source, @settings[:overrides][:environments][:requested_environments]) } }
 
-          case @format
+          case @settings[:overrides][:output][:format]
           when 'json' then json_format(output)
           else yaml_format(output)
           end
@@ -27,7 +48,7 @@ module R10K
           # exit 0
           true
         rescue => e
-          logger.error R10K::Errors::Formatting.format_exception(e, @trace)
+          logger.error R10K::Errors::Formatting.format_exception(e, @settings[:overrides][:output][:trace])
           false
         end
 
@@ -60,21 +81,22 @@ module R10K
         end
 
         def environment_info(env)
-          if !@modules && !@detail
+          modules = @settings[:overrides][:environments][:deploy_modules]
+          if !modules && !@settings[:overrides][:output][:detail]
             env.dirname
           else
             env_info = env.info.merge({
               :status => (env.status rescue nil),
             })
 
-            env_info[:modules] = env.modules.map { |mod| module_info(mod) } if @modules
+            env_info[:modules] = env.modules.map { |mod| module_info(mod) } if modules
 
             env_info
           end
         end
 
         def module_info(mod)
-          if @detail
+          if @settings[:overrides][:output][:detail]
             { :name => mod.title, :properties => mod.properties }
           else
             mod.title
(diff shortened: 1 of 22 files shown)


lib/r10k/action/deploy/module.rb
```diff
@@ -45,10 +60,11 @@ def visit_source(source)
         end
 
         def visit_environment(environment)
-          if @opts[:environment] && (@opts[:environment] != environment.dirname)
-            logger.debug1(_("Only updating modules in environment %{opt_env} skipping environment %{env_path}") % {opt_env: @opts[:environment], env_path: environment.path})
+          requested_envs = @settings[:overrides][:environments][:requested_environments]
+          if !requested_envs.empty? && requested_envs.none? { |env| env != environment.dirname }
```
Magisus: I find the second part of this logic really hard to parse. Do we know there can only ever be one thing in `requested_environments`? If so, can we simplify this to take advantage of that? Otherwise, I think just `include?` might work?

lib/r10k/action/deploy/display.rb
```diff
@@ -9,25 +9,46 @@ class Display < R10K::Action::Base
...
         include R10K::Action::Deploy::DeployHelpers
 
+        def initialize(opts, argv, settings)
+          super
+
+          @settings = @settings.merge({
+            overrides: {
+              environments: {
+                preload_environment: @fetch,
+                requested_environments: @argv
```
Magisus: I see that we didn't do any correction of this before, like we do in [deploy/environment](https://github.com/puppetlabs/r10k/blob/92df0c2e6fb762e7acc16e6f50653e63c7dcafa6/lib/r10k/action/deploy/environment.rb#L29). Is that intentional, or an oversight?

lib/r10k/environment/git.rb
```diff
@@ -50,7 +50,7 @@ def initialize(name, basedir, dirname, options = {})
       :ref     => :self,
       :remote  => :self,
 
-    }, raise_on_unhandled: false)
+    }, raise_on_unhandled: true)
```
Magisus: Is this related to the comment below? If so, are you thinking we ought to cut a branch (i.e. do a major) for this PR to go into? Or do you think the caution there was overblown?

lib/r10k/puppetfile.rb
```diff
@@ -43,17 +43,32 @@ class Puppetfile
...
+  attr_reader :overrides
+
   # @param [String] basedir
-  # @param [String] moduledir The directory to install the modules, default to #{basedir}/modules
-  # @param [String] puppetfile_path The path to the Puppetfile, default to #{basedir}/Puppetfile
-  # @param [String] puppetfile_name The name of the Puppetfile, default to 'Puppetfile'
-  # @param [Boolean] force Shall we overwrite locally made changes?
-  def initialize(basedir, moduledir = nil, puppetfile_path = nil, puppetfile_name = nil, force = nil )
+  # @param [Hash, String, nil] options_or_moduledir The directory to install the modules or a Hash of options.
+  #         Usage as moduledir is deprecated only use as options, defaults to nil
```
Magisus: Can we make this two sentences to make it clearer what's going on here? "...deprecated. Only use..."

spec/unit/action/puppetfile/install_spec.rb
```diff
@@ -2,8 +2,11 @@
 require 'r10k/action/puppetfile/install'
 
 describe R10K::Action::Puppetfile::Install do
-  let(:default_opts) { {root: "/some/nonexistent/path"} }
-  let(:puppetfile) { R10K::Puppetfile.new('/some/nonexistent/path', nil, nil) }
+  let(:default_opts) { { root: "/some/nonexistent/path" } }
+  let(:puppetfile) {
+    R10K::Puppetfile.new('/some/nonexistent/path',
+                         {:moduledir => nil, :puppetfile_path => nil, :force => nil})
```
Magisus, line 8: Do you need to specify these? Won't we fall through to the default whether you specify them as `nil` or don't specify them?

lib/r10k/action/deploy/module.rb
justinstoller: Very good point wrt `include?`! The flag for this action is a single `--environment` but I wanted to try and find any similarities between this and environment handling in `deploy/environment.rb` in case we can merge this functionality later.

lib/r10k/action/deploy/display.rb
justinstoller: I believe that's an oversight.

lib/r10k/action/deploy/module.rb
Magisus: Yeah makes sense to keep treating it as a real list.

lib/r10k/environment/git.rb
justinstoller: This still being here is an oversight. I was experimenting with changing the args passed here and wanted to use this to validate that wip.

spec/unit/action/puppetfile/install_spec.rb
justinstoller: we don't need to specify these to create a Puppetfile object, but the Install action always passes these options in, so I did so for consistency with how Puppetfile.new is called in the tested class.

justinstoller:
I think I've addressed the feedback so far

Magisus (approved):

PR #1158, opened 2021-05-05
(maint) Normalize deploy display override structure

justinstoller:
Previously, the structure in Action::Deploy::Display contained a key at
`overrides.environments.preload_environment` rather than the plural
`overrides.environments.preload_environments`, which is used elsewhere and
what the body of the action actually checked for. Without preloading the
environments changes to the list of environments since the last deploy
will not be shown.

This particularly affects code-manager which uses the display command to
retreive a list of environments when requested to do a "deploy all".

```
```


Magisus (approved):
Ahh, the tiny typos. Nice catch 🙂

Issue #567, opened 2016-02-08
Support for plain tarballs


dannytrigo:
I would like to support downloading Puppet module tarballs from plain HTTP URLs, not just from Puppet Forge. For example:
mod "mymodule", :tarball => "http://puppetrepo/me-mymodule-1.0.13.tar.gz"

I find it strange that this doesn't already exist, when a tarball is a standard packaging format for modules (puppet module build). Is there a reason for it not already existing, or it simply needs implementing?

Cinderhaze:
I would love for support like this, as well as potentially support for s3 based urls

rnelson0:
@dannytrigo I am not sure it is encouraged, but doubt it would be frowned upon as an option. If it is implemented well - including tests and docs - I am pretty sure it could be merged.

My concern as a user is how much of an impact that would have on deployment times and how it would verify authenticity (esp with a high likelyhood of self-signed certs with https).

scotje:
@dannytrigo How would you want to determine whether or not the module needed to be updated/redeployed? Theoretically it could just redeploy every time r10k deploys the environment, but that seems pretty inefficient. We could also just require a certain format to the tarball URL that includes a version number, or have a separate key to declare the desired version.

Cinderhaze:
I know that in our usecase, having an http mirror of forge tarballs, requiring the tar file to have the version in it is enough.

PR #1159, opened 2021-05-06
Tarball

reidmv:
This PR implements:

* An R10K::Tarball base class
* An R10K::Module::Tarball class, providing a "tarball" module type
* An R10K::Environment::Tarball class, providing a "tarball" environment type

Example usage:

```
---
production:
  type: tarball
  source: https://example.com/puppet-code-env-v1.0.0.tar.gz
  version: 99a906c99c2f144de43f2ae500509a7474ed11c583fb623efa8e5b377a3157f0 # sha256digest
  modules:
    puppetlabs-peadm:
      type: forge
      version: 2.5.0
    reidmv-xampl:
      type: git
      version: 62d07f2
    reidmv-mymodule:
      type: tarball
      source: https://example.com/reidmv-mymodule-1.0.0.tar.gz
      version: 6128ada158622cd90f8e1360fb7c2c3830a812d1ec26ddf0db7eb16d61b7293f # sha256digest
```

Later enhancements built on top of this might include Artifactory types, or S3 types, with options for real versions, depending on how the remote repository works.

Solves #567


lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+module R10K
+  class Tarball
+
+    include R10K::Settings::Mixin
+
+    def_setting_attr :proxy
+    def_setting_attr :cache_root, if R10K::Util::Platform.windows?
+                                    File.join(ENV['LOCALAPPDATA'], 'r10k', 'tarball')
+                                  else
+                                    File.expand_path(ENV['HOME'] ? '~/.r10k/tarball': '/root/.r10k/tarball')
```
mwaggett: Would it make sense to just do
```suggestion
                                    File.join(ENV['HOME'] || '/root', '.r10k', 'tarball')
```
?

mwaggett: Do we need to account for `ENV['LOCALAPPDATA']` being unset?

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+    #   @return [String] The tarball's name
+    attr_accessor :name
+
+    # @!attribute [r] source
+    #   @return [String] The tarball's source
+    attr_accessor :source
+
+    # @!attribute [r] sha256digest
+    #   @return [String] The tarball's expected sha256 digest
+    attr_accessor :sha256digest
```
mwaggett: Can we just call this `sha256`?

lib/r10k/tarball.rb
mwaggett: Also, is there a reason to have a different `cache_root` from other module types? I'm looking at https://github.com/puppetlabs/r10k/blob/main/lib/r10k/forge/module_release.rb#L16 and https://github.com/puppetlabs/r10k/blob/main/lib/r10k/forge/module_release.rb#L66 and wondering how much of that can be reused.

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+      ensure
+        reader.close
+      end
+    end
+
+    # @param target_dir [String] The directory to check if is insync with the
+    #        tarball content
+    #
+    # @return [Boolean]
+    def insync?(target_dir, purge: true)
```
mwaggett: Can you add a yardoc for the `purge` param? It's not super clear to me what it's being used for.

mwaggett:
```suggestion
    def in_sync?(target_dir, purge: true)
```
don't love smushing words together 😅

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+      target_dir_entries = Find.find(target_dir).map(&:to_s) - [target_dir]
```
mwaggett: Looking at https://ruby-doc.org/stdlib-2.6.3/libdoc/find/rdoc/Find.html, I don't think this is giving you a collection of entries. You might just want `Dir[File.join(target_dir, '**', '*')]`?

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+      each_tarball_entry do |entry|
+        found = target_dir_entries.delete(File.join(target_dir, entry.full_name.chomp('/')))
+        return false if found.nil?
+        next if entry.directory?
+        return false unless file_digest(found) == reader_digest(entry)
+      end
+      purge ? target_dir_entries.empty? : true
```
mwaggett: Not really sure what `purge` is being used for, but it might make sense to put the `target_dir_entries.empty?` check higher up so that we can return early and not loop through tarball entries if there's nothing in the target dir.

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+
+        temp_digest = case src_uri.scheme
+                      when 'file', nil
+                        copy_to_file(src_uri.path, tempfile)
+                      when %r{[a-z]} # Windows drive letter
+                        copy_to_file(src_uri.to_s, tempfile)
+                      when %r{https?}
+                        download_to_file(src_uri, tempfile)
+                      else
+                        raise "Unexpected source scheme #{src_uri.scheme}"
```
mwaggett, line 129: This should probably raise an `R10K::Error`. And it would probably be helpful to include the `source` in the error message.

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+    def download
+      Dir.mktmpdir do
+        tempfile = Tempfile.new(tarball_basename)
+        tempfile.binmode
+        src_uri = URI.parse(source)
+
+        temp_digest = case src_uri.scheme
+                      when 'file', nil
+                        copy_to_file(src_uri.path, tempfile)
+                      when %r{[a-z]} # Windows drive letter
```
mwaggett: What does 'Windows drive letter' mean?

Also, this should be a much more restrictive regex - right now this will match literally anything with a lowercase letter in it (including 'https').

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+                        copy_to_file(src_uri.to_s, tempfile)
+                      when %r{https?}
+                        download_to_file(src_uri, tempfile)
+                      else
+                        raise "Unexpected source scheme #{src_uri.scheme}"
+                      end
+
+        # Verify the download
+        unless (sha256digest == temp_digest) || sha256digest.nil?
+          raise 'Downloaded file does not match digest'
```
mwaggett: This should also probably raise an `R10K::Error` and ideally include more info - what downloaded file? what does it not match?

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+    end
+
+    # Download the tarball from @source to @cache_path
+    def download
+      Dir.mktmpdir do
+        tempfile = Tempfile.new(tarball_basename)
+        tempfile.binmode
+        src_uri = URI.parse(source)
+
+        temp_digest = case src_uri.scheme
```
mwaggett, line 121: Can we call this `temp_sha256` or something? `digest` is kinda general imo.

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+                      when %r{[a-z]} # Windows drive letter
+                        copy_to_file(src_uri.to_s, tempfile)
+                      when %r{https?}
+                        download_to_file(src_uri, tempfile)
+                      else
+                        raise "Unexpected source scheme #{src_uri.scheme}"
+                      end
+
+        # Verify the download
+        unless (sha256digest == temp_digest) || sha256digest.nil?
```
mwaggett: In what scenario would `sha256digest` be nil? That's something the user is supposed to define in their Puppetfile, right? Are we okay letting them not do that?

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+          raise 'Downloaded file does not match digest'
+        end
+
+        # Move the download to cache_path
+        FileUtils::mkdir_p(cache_root)
+        begin
+          FileUtils.mv(tempfile.path, cache_path)
+        rescue Errno::EACCES
+          FileUtils.cp(tempfile.path, cache_path)
```
mwaggett: When would we expect to encounter this error? Is there also a case where we'd get this error, or another error, when attempting to `cp` instead of `mv`?

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+        end
+
+        tempfile.close
+        tempfile.unlink
```
mwaggett: Do these need to be in an `ensure`?

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+    # Is the cached tarball valid? Checks the cached tarball's digest against
+    # the expected digest.
+    #
+    # @return [Boolean]
+    def cache_valid?
+      unless sha256digest.nil?
+        File.exist?(cache_path) && sha256digest == file_digest(cache_path)
+      else
+        File.exist?(cache_path)
+      end
```
mwaggett: I would switch these around and use an `if` instead of `unless`. Using `unless` with `else` is hard to read imo.

mwaggett: or you could just do
```suggestion
      return false unless File.exist?(cache_path)
      return false if sha256digest && sha256digest != file_digest(cache_path)
      return true
```

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+      File.open(path) do |file|
+        reader_digest(file)
+      end
+    end
+
+    # Return the sha256 digest of the readable data
+    #
+    # @param reader [String] An object that responds to #read
+    # @return [String] The read data's sha256 hex digest
+    def reader_digest(reader)
```
mwaggett: Does it make sense to put these digest methods into some utility class that can be shared? They don't necessarily seem specific to this tarball feature.

Also I think they should include `sha256` in the names, since `digest` is fairly general.

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+      digest = Digest::SHA256.new
+      while chunk = reader.read(CHUNK_SIZE)
+        digest.update(chunk)
+      end
+
+      digest.hexdigest
+    end
+
+    def sanitized_dirname(string)
```
mwaggett: Isn't this defined elsewhere? We should probably pull it into a place that can be shared.

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
+    #
+    # @return [Array] A list of file paths contained in the archive
+    def paths
+      names = Array.new
+      each_tarball_entry { |entry| names << entry.full_name.chomp('/') }
+      names
+    end
+
+    def cache_sha256digest
+      raise "Cache not present" unless File.exist?(cache_path)
```
mwaggett:
```suggestion
      raise R10K::Error, _("Cache not present at %{path}") % {path: cache_path} unless File.exist?(cache_path)
```

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,250 @@
...
+      end
+
+      digest.hexdigest
+    end
+
+    def sanitized_dirname(string)
+      string.gsub(/[^@\w\.-]/, '-')
+    end
+
+    def http_get(uri, method: Net::HTTP::Get, redirect_limit: 10, &block)
```
mwaggett, line 193: If you want this to be a generic http request method, I would change the method name to reflect that. Otherwise, if you want to keep it specific to GET, I would remove `method` as a param.

This also seems like something that should be shared, rather than specific to this class.

lib/r10k/tarball.rb
```diff
@@ -0,0 +1,257 @@
...
```
mwaggett: If you want this to be a generic http request method, I would change the method name to reflect that. Otherwise, if you want to keep it specific to GET, I would remove method as a param.

This also seems like something that should be shared, rather than specific to this class.

mwaggett:
ugh not sure how my review comments got submitted before I was done.. More to come.. 😅 

also @reidmv if/when you make updates, if you could make them in separate commits, that would be helpful for review (and then we can squash where relevant at the end).

lib/r10k/tarball.rb
reidmv: The user setting `cachedir` will dictate what is used here, just as it does [for git and forge cache roots](https://github.com/puppetlabs/r10k/blob/a722b71f49c8e8249f978ac443eb78afe877796f/lib/r10k/initializers.rb#L36-L38).

The value supplied is the default, and is [modeled after R10K::Git::Cache](https://github.com/puppetlabs/r10k/blob/a722b71f49c8e8249f978ac443eb78afe877796f/lib/r10k/git/cache.rb#L19-L29). That model does not account for `ENV['LOCALAPPDATA']` being unset.

It would be cleaner overall if there were a central way to define this default, but at present it doesn't seem like there is one. I think figuring out how to change that may best be done as a separate refactor PR.

lib/r10k/tarball.rb
reidmv: Honestly, I'm leaning towards `version` being the term that gets documented.

If we shorten it, I think `digest` (or the commonly used but slightly more technically ambiguous `checksum`) would be the preferred shortening. That's the part that conveys what this parameter is really about—the output of a hash function which can be used to validate the expected content.

I went with the very specific and technical `sha256digest` because I was trying to avoid the user needing to look up which algorithm to use for the `checksum`. However, we could probably expose that later by offering a `checksum_algorithm` if a use case arose for it.

I've put in a commit to change this to just `checksum` for now.

lib/r10k/tarball.rb
reidmv: Renamed the parameter to `ignore_untracked_files` and added a yarddoc.

reidmv: Yeah. I tend to agree, but `insync` as a term seems to be the r10k convention. As such, I think it's maybe best to leave it this way in service to the greater goal of consistency.

lib/r10k/tarball.rb
reidmv: I took a look into this, and while I'm not 100% sure why it works this way, it seems that `Find.find()` returns an iterator that will hit all the files it is supposed to, and if that iterator is passed to `#map`, it does what's expected. The end result is an array of strings.

The globbing behavior is actually not complete, as it misses dot-files. Given that `Find.find` is working, I think it's best to leave it. I suspect it's more reliable than globbing can be.

lib/r10k/tarball.rb
reidmv: Yep, that would be preferred. Pushed a commit to address this by switching to `Tempfile::open` with a block.

lib/r10k/tarball.rb
reidmv: Good call. Adding a commit to simplify.

lib/r10k/tarball.rb
reidmv: Created an `R10K::Util::Cacheable` module to share this method.

lib/r10k/tarball.rb
reidmv: Removed `method:` as a user opt, and hard-coded to `Net::HTTP::Get`.

Regarding sharing, at the moment I don't know of anything else in R10k that needs to download files this way. I've added a comment to the method which clarifies exactly what it is doing:

```
# Start a Net::HTTP::Get connection, then yield the Net::HTTPSuccess object
# to the caller's block. Follow redirects if Net::HTTPRedirection responses
# are encountered, and honor settings[:proxy].
```

I think this could potentially be generalized for sharing, but I'm hesitant to break it off unless/until there's another consumer in r10k that wants to use it, for fear of over engineering. As near as I can tell R10K currently offloads all download-esque activity to other libraries. I'm considering `Net::HTTP` to be such a library, though, it's probably possible to use Faraday instead... I think we already have that as a dependency, if it gets us something `Net::HTTP` doesn't.

lib/r10k/tarball.rb
reidmv: The names were originally left generic in order to be applicable in the event the digest algorithm ever changes, or is made selectable per-tarball. I've tried making that more evident by using a `@checksum_algorithm` instance variable, and references such as the following. 

```
# At this time, the only checksum type supported is sha256. In the future,
# we may decide to support other algorithms if a use case arises. TBD.
@checksum_algorithm = :SHA256
```

```
def reader_digest(reader)
  digest = Digest(@checksum_algorithm)
```

lib/r10k/tarball.rb
reidmv: Reviewing Faraday, it looks like that library can do everything we need it to and more succinctly than `Net::HTTP`. Given that it's already an r10k dependency, the right thing to do may be to reduce this method to using Faraday instead.

reidmv: I took a stab at using Faraday, but it looks like:

1. We want/need [streaming downloads](https://lostisland.github.io/faraday/usage/streaming), so I think we need a newer version of Faraday than r10k currently supports, due to the puppet-forge gem dependency constraints.
2. Using the newer version of Faraday, we run into a bit of a problem for testing; I can't figure out how to make Faraday's test stubs work with streaming downloads.

If these two problems didn't exist, it seems like we could fairly handily offload this to Faraday. I'm stopping that line of investigation for now though. Seems pointless given item 1.

If it did work, it _might_ look something like this.

```ruby
# @param uri [URI] The URI to download from
# @param output The file or path to save to
# @return [String] The downloaded file's sha256 hex digest
def download_to_file(uri, output)
  digest = DIGEST_CLASS.new

  File.open(output, 'wb') do |output_stream|
    faraday_connection.get(uri) do |req|
      req.options.on_data = Proc.new do |chunk, _|
        output_stream.write(chunk)
        digest.update(chunk)
      end
    end
  end

  digest.hexdigest
end

# Defined separately to enable test stubbing
def faraday_connection
  Faraday.new do |faraday|
    faraday.response(:follow_redirects)
    faraday.adapter(:net_http)
    faraday.proxy(settings[:proxy]) unless settings[:proxy].nil?
  end
end
```

lib/r10k/tarball.rb
reidmv: Note: stumbled across settings.rb, and it looks like there may be a way to set defaults there. Probably worth reviewing. If so, then the default can be set there, and these classes (`R10K::Tarball`, `R10K::Git::Cache`) shouldn't need to set a default themselves... 🤔

lib/r10k/tarball.rb
reidmv: Alright, last tweak for now: went ahead and moved all the file download/copy/checksum related methods out of `R10K::Tarball` and into a new utility module, `R10K::Util::Downloader`.

reidmv:
@mwaggett ready for re-review! I solemnly promise to leave this alone and stop fiddling with it in my spare time until then.* 😇

_* Unlikely to happen but if the mood strikes me I will not refrain from adding additional tests._

reidmv:
I noticed that after the first review, this PR expanded from being a fairly new-files-only change into something that modifies how some existing structures work. @mwaggett if it's helpful for review and approval, I could probably break the PR apart into smaller requests, and split the structure changing stuff from the new feature stuff. Let me know.

reidmv:
Closing in favor of #1244

PR #1160, opened 2021-05-07
(RK-364) Add StaleBot config

Magisus:
This will cause PRs and issues that have been inactive for 60 days to be
flagged, and the closed after 7 more days.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -5,7 +5,7 @@ daysUntilClose: 7
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. It will be closed in 7 days if no further activity occurs.
+  Thank you for your contributions.
 # Comment to post when closing a stale issue. Set to `false` to disable
 closeComment: false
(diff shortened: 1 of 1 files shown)


Magisus:
I added this repo to the list that the app has access to; I guess that's all that's needed?

.github/stale.yml
```diff
@@ -0,0 +1,11 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
```
mwaggett: Should we update this comment to mention the time frames we're using? Not sure if we should mention both or not, but something like
```suggestion
  This issue has been automatically marked as stale because it has not had
  recent activity. It will be closed in 7 days if no further activity occurs.
  Thank you for your contributions.
```

mwaggett (approved):
Did you want to squash or nah?

Magisus:
You can do it when you merge if you want 😉

mwaggett:
I always forget about that! I think because I usually want to squash my own stuff 😅 also I could be convinced either way that this particular PR should or should not be squashed 😝

PR #1161, opened 2021-05-07
(maint) Do not use visitors to synchronize modules

justinstoller:
In removing the visit_module functionality logic that was duplicated within actions has been consolidated. Notably the info message in each action before it called `module.sync` was one of:
```
"Deploying %{origin} content %{path}"
"Deploying module %{mod_path}"
"Updating module %{mod_path}"
```
And now is:
```
"Checking module %{path}"
```
(Open to to a better message, btw).

By moving the `control_branch` logic from the `puppetfile install` command into Module::Git users will see the warning about invalid control branches if applicable during regular deploys as well (eg, if they are deploying git modules with the control_branch pattern in an SVN or Yaml backed environment).

Giving the Module::Base have a method like `will_sync?`seems a bit unnecessary for the single use case of the `deploy module` action, however I think it will provide a useful start to the kind of behavior that Reid mentions in https://github.com/puppetlabs/r10k/pull/1145#discussion_r616076004.

The first commit is a separate refactor that changes the API of Module.new to work more like the (Source|Environment).from_hash that is very similar to. I'm torn on whether that level of breaking change should go in. Would love to hear comments about how much it might be used in the wild (vs being a private API to R10K). I'm also happy to remove that commit as it isn't directly related to the remaining changes.

```
```

### DIFF
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -28,7 +28,7 @@ module R10K
                 generate_types: @generate_types
               },
               modules: {
-                requested_modules: @argv,
+                requested_modules: @argv.map.to_a,
                 # force here is used to make it easier to reason about
                 force: !@no_force
               },
@@ -77,15 +77,12 @@ module R10K
         end
 
         def visit_module(mod)
-          requested_mods = @settings.dig(:overrides, :modules, :requested_modules)
-          if requested_mods.include?(mod.name)
-            mod.sync
-            if mod.environment && @settings.dig(:overrides, :environments, :generate_types)
-              logger.debug("Generating puppet types for environment '#{mod.environment.dirname}'...")
-              mod.environment.generate_types!
-            end
-          else
-            logger.debug1(_("Only updating modules %{modules}, skipping module %{mod_name}") % {modules: requested_mods.inspect, mod_name: mod.name})
+          mod.sync
+
+          requested_mods = @settings.dig(:overrides, :modules, :requested_modules) || []
+          if requested_mods.include?(mod.name) && mod.environment && @settings.dig(:overrides, :environments, :generate_types)
+            logger.debug("Generating puppet types for environment '#{mod.environment.dirname}'...")
+            mod.environment.generate_types!
           end
         end
 
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -40,12 +40,17 @@ module R10K
 
         def call
           @visit_ok = true
+          begin
+            expect_config!
+            deployment = R10K::Deployment.new(@settings)
+            check_write_lock!(@settings)
 
-          expect_config!
-          deployment = R10K::Deployment.new(@settings)
-          check_write_lock!(@settings)
+            deployment.accept(self)
+          rescue => e
+            @visit_ok = false
+            logger.error R10K::Errors::Formatting.format_exception(e, @trace)
+          end
 
-          deployment.accept(self)
           @visit_ok
         end
 
@@ -77,16 +82,11 @@ module R10K
               environment.generate_types!
             end
           end
-
         end
 
         def visit_puppetfile(puppetfile)
           puppetfile.load
           yield
-        end
-
-        def visit_module(mod)
-          mod.sync
         end
 
         def allowed_initialize_opts
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -79,7 +79,6 @@ module R10K
         def visit_module(mod)
           requested_mods = @settings.dig(:overrides, :modules, :requested_modules)
           if requested_mods.include?(mod.name)
-            logger.info _("Deploying module %{mod_path}") % {mod_path: mod.path}
             mod.sync
             if mod.environment && @settings.dig(:overrides, :environments, :generate_types)
               logger.debug("Generating puppet types for environment '#{mod.environment.dirname}'...")
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -15,8 +15,14 @@ module R10K
 
         attr_reader :settings
 
+        # @param opts [Hash] A hash of options defined in #allowed_initialized_opts
+        #   and managed by the SetOps mixin within the Action::Base class.
+        #   Corresponds to the CLI flags and options.
+        # @param argv [CRI::ArgumentList] A list-like collection of the remaining
+        #   arguments to the CLI invocation (after removing flags and options).
+        # @param settings [Hash] A hash of configuration loaded from the relevant
+        #   config (r10k.yaml).
         def initialize(opts, argv, settings)
-
           super
 
           requested_env = @opts[:environment] ? [@opts[:environment].gsub(/\W/, '_')] : []
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -80,7 +80,7 @@ module R10K
           requested_mods = @settings.dig(:overrides, :modules, :requested_modules)
           if requested_mods.include?(mod.name)
             logger.info _("Deploying module %{mod_path}") % {mod_path: mod.path}
-            mod.sync(force: @settings.dig(:overrides, :modules, :force))
+            mod.sync
             if mod.environment && @settings.dig(:overrides, :environments, :generate_types)
               logger.debug("Generating puppet types for environment '#{mod.environment.dirname}'...")
               mod.environment.generate_types!
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -67,8 +67,17 @@ module R10K
             logger.debug1(_("Only updating modules in environment(s) %{opt_env} skipping environment %{env_path}") % {opt_env: requested_envs.inspect, env_path: environment.path})
           else
             logger.debug1(_("Updating modules %{modules} in environment %{env_path}") % {modules: @settings.dig(:overrides, :modules, :requested_modules).inspect, env_path: environment.path})
+
             yield
+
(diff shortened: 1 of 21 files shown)


lib/r10k/environment/with_modules.rb
```diff
@@ -87,25 +87,27 @@ def accept(visitor)
...
+      if !args.is_a?(Hash)
+        args = { version: args }
+      end
+
       add_module(name, args)
     end
   end
 
   # @param [String] name
   # @param [*Object] args
```
Magisus: Should update the docs string if we require this to be a hash.

lib/r10k/environment/base.rb
```diff
@@ -44,14 +44,14 @@ def initialize(name, basedir, dirname, options = {})
...
     @options = options
     @puppetfile_name = options.delete(:puppetfile_name)
-    @overrides = options.delete(:overrides)
+    @overrides = options.delete(:overrides) || {}
 
     @full_path = File.join(@basedir, @dirname)
     @path = Pathname.new(File.join(@basedir, @dirname))
 
     @puppetfile  = R10K::Puppetfile.new(@full_path,
-                                        {overrides: @options[:overrides],
```
justinstoller, line 53: This was a bug that was uncaught earlier, but exposed with the additional testing that comes from attempting to actually use this data at the Module level. It accidentally got rolled into a refactor commit but I'm happy to break it out if that would be better.

lib/r10k/action/deploy/module.rb
```diff
@@ -28,7 +28,7 @@ def initialize(opts, argv, settings)
                 generate_types: @generate_types
               },
               modules: {
-                requested_modules: @argv,
+                requested_modules: @argv.map.to_a,
```
justinstoller, line 37: This was a bug in that was exposed with further functional testing. `@argv` here _acts_ like an array but is actually a CRI::ArgumentList that includes the Enumerable module. This causes problems with logging that calls `inspect` on the value. This change was accidentally rolled into another refactor, but I can break it out if that's best.

lib/r10k/module/base.rb
```diff
@@ -61,11 +65,22 @@ def full_path
   end
 
   # Synchronize this module with the indicated state.
-  # @abstract
+  # @param [Hash] opts Deprecated
   def sync(opts={})
     raise NotImplementedError
   end
 
+  def will_sync?
```
justinstoller: `should_sync?` Update instance variable to match method name (sans '?')?

lib/r10k/module/base.rb
```diff
@@ -62,9 +62,8 @@ def full_path
 
   # Synchronize this module with the indicated state.
   # @param [Hash] opts Deprecated
-  # @abstract
   def sync(opts={})
-    raise NotImplementedError
+    logger.info _("Checking module %{path}") % {path: path}
```
Magisus: Not sure if I like the word "checking" or not... So this gets logged any time we go to sync a module, right? And that is when we're deploying either a whole env or just this module, and the module may
1) not exist
2) exist but need updating
3) exist and already be at the right version

Is there additional logging elsewhere to indicate which of these states we were in?

justinstoller:
mmm... I was testing with Ruby 2.7 locally. It looks like I may have used something that is specific to that version (everything _lower_ than that version of Ruby failed in Travis).

lib/r10k/action/deploy/module.rb
Magisus: Might be good to add a docs comment to this method indicating the type as it comes in?

justinstoller:
Looks like Array#intersection wasn't added until 2.7. I've replaced it with the (IMO less readable) Array#&.

justinstoller:
I should do a bit more functional testing as well, but I wanted folks to be able to start reviewing asap.

Magisus:
Personally I'm pro consolidating APIs so they don't take multiple types, particularly a language with heavy type inference like Ruby. The variability adds complexity that usually feels unnecessary. However, I understand it could be a breaking change. _Are_ people using r10k as a library? Does Bolt, maybe?

lib/r10k/module/base.rb
justinstoller: Not for those higher level cases and not at info level. However, there are debug messages by lower level objects that can differentiate between those. I was going to use "Synchronizing" since internally it's called "sync" but I something stopped me from doing that - buy I don't remember why anymore.

justinstoller:
Would also love to get @reidmv 's feedback on this work.

lib/r10k/module/base.rb
Magisus: I think I kind of agree that `synchronizing` might not be good, because sometimes we don't do anything (e.g. if the module is already in the right state). In that sense, "checking" is fine. I guess my mind just says "Checking for what?"

justinstoller: I'd also be fine moving back to the term "Deploying" since that was used previously in two the three previous messages.

reidmv: For least-change, I'd also suggest going back to "Deploying". Note on the purpose of `%{origin}`, that helps indicate in the logging where a module was defined, which can be useful when, for example, a module is defined both in the Puppetfile and by the environment, and the environment's version is overriding the definition from the Puppetfile.

lib/r10k/action/deploy/module.rb
```diff
@@ -77,15 +77,12 @@ def visit_puppetfile(puppetfile)
...
-          requested_mods = @settings.dig(:overrides, :modules, :requested_modules)
-          if requested_mods.include?(mod.name)
-            mod.sync
-            if mod.environment && @settings.dig(:overrides, :environments, :generate_types)
-              logger.debug("Generating puppet types for environment '#{mod.environment.dirname}'...")
-              mod.environment.generate_types!
-            end
-          else
-            logger.debug1(_("Only updating modules %{modules}, skipping module %{mod_name}") % {modules: requested_mods.inspect, mod_name: mod.name})
+          mod.sync
```
reidmv: I'm going through this commit-by-commit, so maybe I'll come around to this as I move forward, but... at first blush, I feel like this change (1901284b05164f920c882a3281f062ecb6dafac2) actually makes the code _more_ confusing, rather than less. Admittedly, I've been working with r10k for a little while, so may be looking a little more favorably on visitor stuff than I would have when I first started trying to figure it out.

In general, I don't like the idea of putting decision-making code into module objects. I am in favor of keeping module objects dumb, being given only simple instructions and doing what they're told. I think we should be able to find a way to simplify the visitor stuff without losing that simplicity for module objects.

reidmv (commented):
I have mixed feelings about this.

I don't like calling sync() on all modules, along with settings, and having them self-decide what to do with that.

Would we be able to simplify this if we instead sort of "demoted" the Puppetfile?

Alternate algorith:

* _[Action]_
Visit deployment
    * _[Deployment]_
    Choose, then visit, environments
        * _[Environment]_
        Call `environment.puppetfile.load()`
        Choose, then concurrently visit, modules
            * _[Module]_
            Call `module.sync()`

Would that work? Would demoting the Puppetfile and removing it from the visitor pattern accomplish the reduction in complexity that's needed?

Side note: laying it out like this, it seems like we could also decide to visit chosen environments concurrently, if we wanted to. We'd probably just need to move our queuing/concurrency code into an `R10K::Action::Visitor#concurrently_visit` method, which accepted an array of `other` objects instead of just one.

justinstoller:
> Would we be able to simplify this if we instead sort of "demoted" the Puppetfile?

I don't quite follow what you mean by "demoting" the Puppetfile. But, if you'd like my idea of where this is going it's this:
 - Remove the Puppetfile and Modules from participating in the visitor pattern.
 - Break out the Puppetfile into its constituent behaviors:
     - Given some configuration it loads modules definitions
     - It synchronizes module content. (started in https://github.com/puppetlabs/r10k/pull/1153)
     - It tracks managed content for purging (started in https://github.com/puppetlabs/r10k/pull/1155)
 - Refactor Environments to load modules from configured module loaders (Puppetfile would be one source, the WithModule Env is close to how I'd see this end up being), then pass those modules to the content synchronizer, and clean up afterwards.

Is something like that what you're thinking of when you say "demote" the Puppetfile?

> I don't like calling sync() on all modules, along with settings, and having them self-decide what to do with that.

As far as the calling sync and letting them decide...

I don't really care. Having the module container make the call of what module gets sync called on it was my approach in https://github.com/puppetlabs/r10k/pull/1145. In that case the container (in that case a Puppetfile) will need to separately track all the modules loaded, the content to be managed (and not purged), and the modules to _actually_ sync. Having the module itself know if it should be updated is also fine by me and I believe closely tracks your suggestion in this comment: https://github.com/puppetlabs/r10k/pull/1145#discussion_r616076004. I don't think there's a huge difference between calling
```rb
  mod.sync
```
when mod#sync is defined by:
```rb
def sync
  if should_sync?
    do stuff...
end
```
vs
```rb
mod.sync if mod.should_sync?
```
and
```rb
def sync
  do stuff...
end
```
Assuming the `should_sync?` method isn't otherwise used by callers of `mod.sync`. (fwiw, I believe you used the more Puppet-y `insync?` in your comment).

I think it's a better separation of concerns for the caller of module.sync to decide but I also think it makes the caller more complicated for logic that is pretty trivially inlined into the module. So :shrug:. I have a feeling you were less concerned with which object had the responsibility and more to do with the fact that the particular object being updated was the Puppetfile class?

Regardless of the end state, I do think that having the module determine whether or not it syncs will be advantageous to the further refactors that I mentioned at the top of this comment (breaking up the Puppetfile class).

reidmv (approved):
Talked with @justinstoller about this, I'm 👍  to move it forward re: design.

There are a lot of good fixes in here. If we can move it forward, as well as get a solution for CODEMGMT-1415 in place, there will be room afterwards to start more gentle refactoring, if we want to and if it's important enough.

Small steps = good. 🙂

Magisus (approved):

PR #1162, opened 2021-05-10
(RK-364) Switch to `actions/stale`

Magisus:
StaleBot does not seem to be working. This commit removes its config in
favor of using `actions/stale` instead.

Issues and PRs will be marked as stale after 60 days, and closed if
there is no activity after 7 more days.


Magisus:
This is based on the version in the PDK repo https://github.com/puppetlabs/pdk/blob/main/.github/workflows/stale.yml, with a difference where we actually close issues (and they do not).

.github/workflows/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+  - cron: "30 1 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-stale: 90
```
mwaggett: 60 was the default that Stale Bot used, which we agreed would be a good starting place. personally, if something hadn't been touched for 60 days, I wouldn't expect anything to change after an additional 30 days. also, the longer this period is, the less likely it is that anyone will have retained context on the PR/issue in question in the event that we do want to move forward with it in some way.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+        days-before-close: 7
+        stale-issue-message: 'This issue has been marked stale because it has had no activity for 90 days. The Puppet Team is actively prioritizing existing bugs and new features, if this issue is still important to you please comment and we will add this to our backlog to complete. Otherwise, it will be closed in 7 days.'
+        stale-issue-label: 'no-issue-activity'
```
mwaggett: I know this was copied from PDK, but I'm wondering is we should just use something simple like 'stale', since this applies to both issues and PRs.

Magisus: I did create this label. Renamed it in response to this comment.

.github/workflows/stale.yml
Magisus: Weird, this was updated for me locally, and I pushed.. but it just... didn't update. Weird. Fixed now :)

mwaggett (approved):

PR #1164, opened 2021-05-12
(maint) Remove the visitor pattern from Puppetfile usage

justinstoller:
```
```

### DIFF
--- a/spec/unit/action/deploy/environment_spec.rb
+++ b/spec/unit/action/deploy/environment_spec.rb
@@ -83,13 +83,13 @@ describe R10K::Action::Deploy::Environment do
       end
 
       it "syncs the puppetfile when given the puppetfile flag" do
-        expect(puppetfile).to receive(:accept).and_return([])
+        expect(puppetfile).to receive(:sync)
         action = described_class.new({config: "/some/nonexistent/path", puppetfile: true}, [], {})
         action.call
       end
 
       it "syncs the puppetfile when given the modules flag" do
-        expect(puppetfile).to receive(:accept).and_return([])
+        expect(puppetfile).to receive(:sync)
         action = described_class.new({config: "/some/nonexistent/path", modules: true}, [], {})
         action.call
       end
@@ -206,21 +206,13 @@ describe R10K::Action::Deploy::Environment do
       end
     end
 
-    describe 'extracting credentials' do
-      let(:deployment) do
-        R10K::Deployment.new(mock_config)
-      end
-
-    end
-
     describe "Purging white/allowlist" do
 
       let(:settings) { { deploy: { purge_levels: [:environment], purge_allowlist: ['coolfile', 'coolfile2'] } } }
-
+      let(:overrides) { { environments: {}, modules: {}, purging: { purge_levels: [:environment], purge_allowlist: ['coolfile', 'coolfile2'] } } }
       let(:deployment) do
-        R10K::Deployment.new(mock_config.merge(settings))
+        R10K::Deployment.new(mock_config.merge(overrides))
       end
-
       before do
         expect(R10K::Deployment).to receive(:new).and_return(deployment)
       end
@@ -246,9 +238,22 @@ describe R10K::Action::Deploy::Environment do
 
     describe "purge_levels" do
       let(:settings) { { deploy: { purge_levels: purge_levels } } }
+      let(:overrides) do
+        {
+          environments: {
+            requested_environments: ['PREFIX_first']
+          },
+          modules: {
+            deploy_modules: true
+          },
+          purging: {
+            purge_levels: purge_levels
+          }
+        }
+      end
 
       let(:deployment) do
-        R10K::Deployment.new(mock_config.merge(settings))
+        R10K::Deployment.new(mock_config.merge({ overrides: overrides }))
       end
 
       before do
@@ -261,9 +266,13 @@ describe R10K::Action::Deploy::Environment do
         let(:purge_levels) { [:deployment] }
 
         it "only logs about purging deployment" do
+          expect(subject).to receive(:visit_environment).and_wrap_original do |original, env, &block|
+            expect(env.logger).to_not receive(:debug).with(/purging unmanaged puppetfile content/i)
+            original.call(env)
+          end.at_least(:once)
+
           expect(subject.logger).to receive(:debug).with(/purging unmanaged environments for deployment/i)
           expect(subject.logger).to_not receive(:debug).with(/purging unmanaged content for environment/i)
-          expect(subject.logger).to_not receive(:debug).with(/purging unmanaged puppetfile content/i)
 
           subject.call
         end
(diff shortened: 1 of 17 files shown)


spec/unit/action/deploy/environment_spec.rb
```diff
@@ -206,21 +206,13 @@
       end
     end
 
-    describe 'extracting credentials' do
```
justinstoller, line 209: Maybe I should have broken this out, I don't know what it was doing here.

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -273,16 +282,18 @@
         let(:purge_levels) { [:environment] }
```
justinstoller, line 282: Locally, these two tests regarding environment level purging are failing.

spec/unit/action/deploy/environment_spec.rb
Magisus: Looks like leftover maybe from the work Tony and I did around OAuth tokens. I think we moved the logic (and tests) elsewhere, and probably forgot this.

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -273,16 +282,18 @@
...
+
           expect(subject.logger).to receive(:debug).with(/purging unmanaged content for environment/i)
           expect(subject.logger).to_not receive(:debug).with(/purging unmanaged environments for deployment/i)
-          expect(subject.logger).to_not receive(:debug).with(/purging unmanaged puppetfile content/i)
 
           subject.call
         end
 
         it "logs that environment was not purged if deploy failed" do
-          expect(subject).to receive(:visit_puppetfile) { subject.instance_variable_set(:@visit_ok, false) }
```
justinstoller, line 284: While I'm not sure what is wrong with the first test of these two. This test, I think, is failing because it depends on the instance variable `@visit_ok` being set to false during the deploy. However, I attempted this with no success:
```rb
expect(subject).to receive(:visit_environment).and_wrap_original do |original, env, &block|
  expect(env).to receive(:deploy) { subject.instance_variable_set(:@visit_ok, false) }
  original.call(env)
end
```
without success. (Not sure why I switched the test content back)

I'm wondering if I'm missing something simple in the first test case and if figured out then this one would be fixable by the above code block (or similar)

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -275,13 +284,17 @@
         it "only logs about purging environment" do
           expect(subject.logger).to receive(:debug).with(/purging unmanaged content for environment/i)
           expect(subject.logger).to_not receive(:debug).with(/purging unmanaged environments for deployment/i)
-          expect(subject.logger).to_not receive(:debug).with(/purging unmanaged puppetfile content/i)
```
justinstoller, line 278: So whenever I tried to implement a similar "wrap_original" expectation on "visit_environment" like above to test for the logger output of "purging unmanaged puppetfile content" the remaining expectations wouldn't fire. I think that's a testing issue though I'm unclear on what the root cause is. I wanted to get other's opinion on it. This test is testing less, though I don't know if it's any worse.

Magisus: Weird that it works in one case but not the other... I feel like I ran into something weird like this recently with some work I was doing on the CA CLI tests, but I can't remember if/how I fixed it... I'll think about it.

spec/unit/action/deploy/environment_spec.rb
Magisus: Is this out of date now? All the tests seem to be passing at this point.

spec/unit/action/deploy/module_spec.rb
```diff
@@ -90,8 +90,8 @@
           else
             expect(environment).not_to receive(:generate_types!)
           end
-          @modules << mod
           expect(environment.puppetfile).to receive(:modules).and_return([mod]).twice
+          @modules << mod
```
Magisus: Why did this need to move?

Magisus (commented):
Definitely feel like this is an improvement, that will only get better if/when we remove the rest of the pattern.

spec/unit/action/deploy/environment_spec.rb
justinstoller: Yeah, this is out of date now.

spec/unit/action/deploy/module_spec.rb
justinstoller: :facepalm: It didn't. I was messing with this before block and put it back wrong.

spec/unit/action/deploy/environment_spec.rb
Magisus: So... it doesn't actually look like you need the `and_wrap_original` at all. From my testing, `env.logger` is the same as `subject.logger`, so you can just add both expectations to `subject.logger` directly. If I do that, the test passes. Works for both for me.

Magisus: I don't know why using `and_wrap_original` doesn't work though 🤷‍♀️

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -261,9 +266,13 @@
         let(:purge_levels) { [:deployment] }
 
         it "only logs about purging deployment" do
+          expect(subject).to receive(:visit_environment).and_wrap_original do |original, env, &block|
+            expect(env.logger).to_not receive(:debug).with(/purging unmanaged puppetfile content/i)
```
Magisus, line 270: So you know what's bizarre? I don't think this expectation is firing either. I flipped it to `to receive` and the test still passes. I also added a `expect(true).to eq(false)` to this block and it also didn't fail.

Magisus: However, if I drop the `expect` on `env.logger`, the `expect(true)` does trip...

spec/unit/action/deploy/environment_spec.rb
Magisus: According to pry, that `env` object doesn't have a `logger` method (or instance var, for that matter). So I _think_ this block is actually throwing and the exception is being squashed, but it's messing with other stuff. I don't know _why_ it doesn't seem to have the method, something to do with weird test doubles maybe?
```
NoMethodError: undefined method `logger' for #<R10K::Environment::Mock:0x00007fcc6bb11220>
```

justinstoller: Ah ha! So it looks like the logger functionality is included in all the child Environment objects, rather than the Environment::Base class. This deployment uses Environment::Mock which doesn't include the logging framework.

Including the R10K::Logging module into the Environment::Mock class causes _a different_ test to fail.... So I think I should move the inclusion of the Logging framework into the Environment::Base and then audit the usages of loggers & the mock environment...

Magisus: That makes sense to me. Yeah, it took me a bit to track down exactly where `logger` was coming from. I expected it to be part of the `Base` environment...

justinstoller: Okay, I moved the logging inclusion for the Environments and while I was at it for the Sources too (since they also have a Mock version that doesn't include a logger). There was also an issue with the test data not being passed down from Source::Mock to Environment::Mock correctly. I think that and the silently failing logger calls were the two issues conspiring against my tests before.

justinstoller:
Also did some functional testing against the dev branch of https://github.com/justinstoller/r10k-testing (with puppetfile and environment purge levels). Looks good.

Magisus (approved):
Did you want other review on this?

justinstoller:
No, I think the general shape of was already agreed upon. When I get to refactoring environments/puppetfile usage I'd like a larger audience.

PR #1165, opened 2021-05-16
(main) Only deploy modules to requested envs

Magisus:
This commit fixes a logic error in the check for which environments to
deploy modules to when using `r10k deploy module -e <env> <mod>`.
Previously, it would deploy to all envs except the one you asked for.
Now it will only deploy to the requested env.

```
```


Magisus:
There might be testing missing here that would have caught this. I'll look on Monday.

Magisus:
Man adding expectations on those loggers is finicky... if you expect the wrong thing the errors are totally off from what they should be sometimes.

justinstoller (approved):

PR #1166, opened 2021-06-01
(maint) Allow modules to sync even if one fails

justinstoller:
When removing the visitor from module synchronization we lost the
blanket rescue around every module sync call. The visitor's visit
method would rescue any exception, log it and then mark the `@ok_visit`
variable as false. This would then cause the CLI command to exit with
a 1 after every other action (and module sync) had finished.

A failed module sync should mark the command invocation as failed,
however some users may depend on the fact that invocation failure
would still result in all modules capable of being synced being synced.

This updates the synchronizer to do so.

```
```

### DIFF
--- a/lib/r10k/content_synchronizer.rb
+++ b/lib/r10k/content_synchronizer.rb
@@ -8,9 +8,9 @@ module R10K
     end
 
     def self.serial_sync(modules)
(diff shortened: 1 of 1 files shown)


justinstoller:
needs tests

PR #1167, opened 2021-06-14
(maint) Use a module loader within the Puppetfile class

justinstoller:
This series of commits moves the logic for loading the module content from a Puppetfile and into a new concept called a ModuleLoader. The idea being to standardize the interface for loading modules, since there are multiple ways to load module specifications (eg, puppetfile, exec, yamldir).

This approach would allow Environments to no longer care _how_ modules were loaded but simply call `load` on any ModuleLoaders they are configured to use. The environment would then have a canonical list of all its environments (instead of deferring to the Puppetfile class when available) and be able to pass all of them to ContentSynchronizer at once (currently only Puppetfile content can be concurrent synced). As well as a list of managed directories, desired content, and purge exclusions for usage with the Cleaner util.

Refactoring the environment to use Module Loaders is a somewhat larger refactor and may break internal APIs. This changeset only refactors the Puppetfile to use the ModuleLoader under the hood. This should be a backwards compatible refactor save that `load` now returns the hash of loaded content information rather than `true` (we only check for it's truthiness - not a hard type check, so it remains backwards compatible for our uses). 

```
```

### DIFF
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -13,22 +13,24 @@ module R10K
       # @param basedir [String] The path that contains the moduledir &
       #     Puppetfile by default. May be an environment, project, or
       #     simple directory.
-      # @param puppetfile [String] The full path to the Puppetfile
-      # @param moduledir [String] The full path to the moduledir
+      # @param puppetfile [String] The path to the Puppetfile, either an
+      #     absolute full path or a relative path with regards to the basedir.
+      # @param moduledir [String] The path to the moduledir, either an
+      #     absolute full path or a relative path with regards to the basedir.
       # @param forge [String] The url (without protocol) to the Forge
       # @param overrides [Hash] Configuration for loaded modules' behavior
       # @param environment [R10K::Environment] The environment loading may be
       #     taking place within
       def initialize(basedir:,
-                     moduledir: File.join(basedir, DEFAULT_MODULEDIR),
-                     puppetfile: File.join(basedir, DEFAULT_PUPPETFILE_NAME),
+                     moduledir: DEFAULT_MODULEDIR,
+                     puppetfile: DEFAULT_PUPPETFILE_NAME,
                      forge: DEFAULT_FORGE_API,
                      overrides: {},
                      environment: nil)
 
-        @basedir     = basedir
-        @moduledir   = moduledir
-        @puppetfile  = puppetfile
+        @basedir     = cleanpath(basedir)
+        @moduledir   = resolve_path(@basedir, moduledir)
+        @puppetfile  = resolve_path(@basedir, puppetfile)
         @forge       = forge
         @overrides   = overrides
         @environment = environment
@@ -76,11 +78,7 @@ module R10K
 
       # @param [String] moduledir
       def set_moduledir(moduledir)
-        @moduledir = if Pathname.new(moduledir).absolute?
-          moduledir
-        else
-          File.join(@basedir, moduledir)
-        end
+        @moduledir = resolve_path(@basedir, moduledir)
       end
 
       # @param [String] name
@@ -97,7 +95,7 @@ module R10K
         args[:overrides] = @overrides
 
         if install_path = args.delete(:install_path)
-          install_path = resolve_install_path(install_path)
+          install_path = resolve_path(@basedir, install_path)
           validate_install_path!(install_path, name)
         else
           install_path = @moduledir
@@ -136,29 +134,24 @@ module R10K
         end
       end
 
-      def resolve_install_path(path)
-        pn = Pathname.new(path)
-
-        unless pn.absolute?
-          pn = Pathname.new(File.join(@basedir, path))
+      def resolve_path(base, path)
+        if Pathname.new(path).absolute?
+          cleanpath(path)
+        else
+          cleanpath(File.join(base, path))
         end
-
-        # .cleanpath is as good as we can do without touching the filesystem.
-        # The .realpath methods will also choke if some of the intermediate
-        # paths are missing, even though we will create them later as needed.
-        pn.cleanpath.to_s
       end
 
       def validate_install_path!(path, modname)
-        unless /^#{Regexp.escape(real_basedir)}.*/ =~ path
-          raise R10K::Error.new("Puppetfile cannot manage content '#{modname}' outside of containing environment: #{path} is not within #{real_basedir}")
+        unless /^#{Regexp.escape(@basedir)}.*/ =~ path
+          raise R10K::Error.new("Puppetfile cannot manage content '#{modname}' outside of containing environment: #{path} is not within #{@basedir}")
         end
 
         true
       end
 
       def determine_managed_directories(managed_content)
-        managed_content.keys.reject { |dir| dir == real_basedir }
+        managed_content.keys.reject { |dir| dir == @basedir }
       end
 
       # Returns an array of the full paths to all the content being managed.
@@ -177,8 +170,12 @@ module R10K
         end
       end
 
-      def real_basedir
-        Pathname.new(@basedir).cleanpath.to_s
+      # .cleanpath is as good as we can do without touching the filesystem.
+      # The .realpath methods will choke if some of the intermediate paths
+      # are missing, even though in some cases we will create them later as
+      # needed.
+      def cleanpath(path)
+        Pathname.new(path).cleanpath.to_s
       end
 
       # For testing purposes only
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -19,8 +19,8 @@ module R10K
       #     absolute full path or a relative path with regards to the basedir.
       # @param forge [String] The url (without protocol) to the Forge
       # @param overrides [Hash] Configuration for loaded modules' behavior
-      # @param environment [R10K::Environment] The environment loading may be
-      #     taking place within
+      # @param environment [R10K::Environment] When provided, the environment
+      #     in which loading takes place
       def initialize(basedir:,
                      moduledir: DEFAULT_MODULEDIR,
                      puppetfile: DEFAULT_PUPPETFILE_NAME,
@@ -52,16 +52,16 @@ module R10K
 
         managed_content = @modules.group_by(&:dirname).freeze
 
-        @managed_directories = determine_managed_directories(managed_content).freeze
-        @desired_contents = determine_desired_contents(managed_content).freeze
-        @purge_exclusions = determine_purge_exclusions(@managed_directories.clone).freeze
+        @managed_directories = determine_managed_directories(managed_content)
+        @desired_contents = determine_desired_contents(managed_content)
+        @purge_exclusions = determine_purge_exclusions(@managed_directories)
 
         {
           modules: @modules,
           managed_directories: @managed_directories,
           desired_contents: @desired_contents,
           purge_exclusions: @purge_exclusions
-        }.freeze
+        }
 
       rescue SyntaxError, LoadError, ArgumentError, NameError => e
         raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile})
@@ -110,7 +110,7 @@ module R10K
           install_path = @moduledir
         end
 
-        if @default_branch_override != nil
+        if @default_branch_override
           module_info[:default_branch_override] = @default_branch_override
         end
 
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -42,11 +42,11 @@ module R10K
         @purge_exclusions = []
       end
 
-      def load!
+      def load
         dsl = R10K::ModuleLoader::Puppetfile::DSL.new(self)
         dsl.instance_eval(puppetfile_content(@puppetfile), @puppetfile)
 
-        validate_no_duplicate_names!(@modules)
+        validate_no_duplicate_names(@modules)
         @modules.freeze
 
         managed_content = @modules.group_by(&:dirname).freeze
@@ -104,7 +104,7 @@ module R10K
 
         if install_path = module_info.delete(:install_path)
           install_path = resolve_path(@basedir, install_path)
-          validate_install_path!(install_path, name)
+          validate_install_path(install_path, name)
         else
           install_path = @moduledir
         end
@@ -128,10 +128,10 @@ module R10K
      private
 
       # @param [Array<R10K::Module>] modules
-      def validate_no_duplicate_names!(modules)
+      def validate_no_duplicate_names(modules)
         dupes = modules
                 .group_by { |mod| mod.name }
-                .select { |_, v| v.size > 1 }
+                .select { |_, mods| mods.size > 1 }
                 .map(&:first)
         unless dupes.empty?
           msg = _('Puppetfiles cannot contain duplicate module names.')
@@ -149,7 +149,7 @@ module R10K
         end
       end
 
-      def validate_install_path!(path, modname)
+      def validate_install_path(path, modname)
         unless /^#{Regexp.escape(@basedir)}.*/ =~ path
           raise R10K::Error.new("Puppetfile cannot manage content '#{modname}' outside of containing environment: #{path} is not within #{@basedir}")
         end
@@ -177,10 +177,10 @@ module R10K
         end
       end
 
-      # .cleanpath is as good as we can do without touching the filesystem.
-      # The .realpath methods will choke if some of the intermediate paths
-      # are missing, even though in some cases we will create them later as
-      # needed.
+      # .cleanpath is as close to a canonical path as we can do without touching
+      # the filesystem. The .realpath methods will choke if some of the
+      # intermediate paths are missing, even though in some cases we will create
+      # them later as needed.
       def cleanpath(path)
         Pathname.new(path).cleanpath.to_s
       end
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -82,19 +82,27 @@ module R10K
       end
 
       # @param [String] name
-      # @param [Hash, String, Symbol] args Calling with anything but a Hash is
-      #   deprecated. The DSL will now convert String and Symbol versions to
-      #   Hashes of the shape
+      # @param [Hash, String, Symbol, nil] module_info Calling with
+      #   anything but a Hash is deprecated. The DSL will now convert
+      #   String and Symbol versions to Hashes of the shape
       #     { version: <String or Symbol> }
       #
-      def add_module(name, args)
-        if !args.is_a?(Hash)
-          args = { version: args }
+      #   String inputs should be valid module versions, the Symbol
+      #   `:latest` is allowed, as well as `nil`.
+      #
+      #   Non-Hash inputs are only ever used by Forge modules. In
+      #   future versions this method will require the caller (the
+      #   DSL class, not the Puppetfile author) to do this conversion
+      #   itself.
+      #
+      def add_module(name, module_info)
+        if !module_info.is_a?(Hash)
+          module_info = { version: module_info }
         end
 
-        args[:overrides] = @overrides
+        module_info[:overrides] = @overrides
 
-        if install_path = args.delete(:install_path)
+        if install_path = module_info.delete(:install_path)
           install_path = resolve_path(@basedir, install_path)
           validate_install_path!(install_path, name)
         else
@@ -102,10 +110,10 @@ module R10K
         end
 
         if @default_branch_override != nil
-          args[:default_branch_override] = @default_branch_override
+          module_info[:default_branch_override] = @default_branch_override
         end
 
-        mod = R10K::Module.new(name, install_path, args, @environment)
+        mod = R10K::Module.new(name, install_path, module_info, @environment)
         mod.origin = :puppetfile
 
         # Do not load modules if they would conflict with the attached
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -1,31 +1,39 @@
-require 'r10k/logging'
-
 module R10K
   module ModuleLoader
     class Puppetfile
 
-      include R10K::Logging
+      DEFAULT_MODULEDIR = 'modules'
+      DEFAULT_PUPPETFILE_NAME = 'Puppetfile'
+      DEFAULT_FORGE_API = 'forgeapi.puppetlabs.com'
 
       attr_accessor :default_branch_override, :environment
       attr_reader :modules, :moduledir,
         :managed_directories, :desired_contents, :purge_exclusions
 
-      # @param [Hash] options
-      # @option options [String] :puppetfile
-      # @option options [String] :moduledir
-      # @option options [String] :basedir
-      # @option options [String] :forge
-      # @option options [Hash] :overrides
-      # @option options [R10K::Environment] :environment
-      def initialize(puppetfile:, moduledir:, forge:, basedir:, overrides:, environment:)
-        @puppetfile  = puppetfile
-        @moduledir   = moduledir
+      # @param basedir [String] The path that contains the moduledir &
+      #     Puppetfile by default. May be an environment, project, or
+      #     simple directory.
+      # @param puppetfile [String] The full path to the Puppetfile
+      # @param moduledir [String] The full path to the moduledir
+      # @param forge [String] The url (without protocol) to the Forge
+      # @param overrides [Hash] Configuration for loaded modules' behavior
+      # @param environment [R10K::Environment] The environment loading may be
+      #     taking place within
+      def initialize(basedir:,
+                     moduledir: File.join(basedir, DEFAULT_MODULEDIR),
+                     puppetfile: File.join(basedir, DEFAULT_PUPPETFILE_NAME),
+                     forge: DEFAULT_FORGE_API,
+                     overrides: {},
+                     environment: nil)
+
         @basedir     = basedir
+        @moduledir   = moduledir
+        @puppetfile  = puppetfile
+        @forge       = forge
         @overrides   = overrides
         @environment = environment
 
         @modules = []
-        @managed_content = {}
 
         @managed_directories = []
         @desired_contents = []
@@ -33,29 +41,33 @@ module R10K
       end
 
       def load!
-        if !File.readable?(@puppetfile)
-          logger.debug _("Puppetfile %{path} missing or unreadable") % {path: @puppetfile.inspect}
-          return false
-        end
-
         dsl = R10K::ModuleLoader::Puppetfile::DSL.new(self)
-        dsl.instance_eval(File.read(@puppetfile), @puppetfile)
+        dsl.instance_eval(puppetfile_content(@puppetfile), @puppetfile)
 
-        validate_no_duplicate_names(@modules)
-        set_managed_directories
-        set_desired_contents
-        set_purge_exclusions
+        validate_no_duplicate_names!(@modules)
+        @modules.freeze
+
+        managed_content = @modules.group_by(&:dirname).freeze
+
+        @managed_directories = determine_managed_directories(managed_content).freeze
+        @desired_contents = determine_desired_contents(managed_content).freeze
+        @purge_exclusions = determine_purge_exclusions(@managed_directories.clone).freeze
 
         {
           modules: @modules,
           managed_directories: @managed_directories,
           desired_contents: @desired_contents,
           purge_exclusions: @purge_exclusions
-        }
+        }.freeze
 
       rescue SyntaxError, LoadError, ArgumentError, NameError => e
         raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile})
       end
+
+
+      ##
+      ## set_forge, set_moduledir, and add_module are used directly by the DSL class
+      ##
 
       # @param [String] forge
       def set_forge(forge)
@@ -86,7 +98,7 @@ module R10K
 
         if install_path = args.delete(:install_path)
           install_path = resolve_install_path(install_path)
-          validate_install_path(install_path, name)
+          validate_install_path!(install_path, name)
         else
           install_path = @moduledir
         end
@@ -105,18 +117,13 @@ module R10K
           return @modules
         end
 
-        # Keep track of all the content this Puppetfile is managing to enable purging.
-        @managed_content[install_path] = Array.new unless @managed_content.has_key?(install_path)
-        @managed_content[install_path] << mod.name
-
         @modules << mod
       end
 
-
-
      private
-      # @param [Array<String>] modules
-      def validate_no_duplicate_names(modules)
+
+      # @param [Array<R10K::Module>] modules
+      def validate_no_duplicate_names!(modules)
         dupes = modules
                 .group_by { |mod| mod.name }
                 .select { |_, v| v.size > 1 }
@@ -142,7 +149,7 @@ module R10K
         pn.cleanpath.to_s
       end
 
-      def validate_install_path(path, modname)
+      def validate_install_path!(path, modname)
         unless /^#{Regexp.escape(real_basedir)}.*/ =~ path
           raise R10K::Error.new("Puppetfile cannot manage content '#{modname}' outside of containing environment: #{path} is not within #{real_basedir}")
         end
@@ -150,32 +157,34 @@ module R10K
         true
       end
 
-      def set_managed_directories
-        @managed_directories = @managed_content.keys.reject { |dir| dir == real_basedir }
+      def determine_managed_directories(managed_content)
+        managed_content.keys.reject { |dir| dir == real_basedir }
       end
 
       # Returns an array of the full paths to all the content being managed.
       # @return [Array<String>]
-      def set_desired_contents
-        @desired_contents = @managed_content.flat_map do |install_path, modnames|
-          modnames.collect { |name| File.join(install_path, name) }
+      def determine_desired_contents(managed_content)
+        managed_content.flat_map do |install_path, mods|
+          mods.collect { |mod| File.join(install_path, mod.name) }
         end
       end
 
-      def set_purge_exclusions
-        exclusions = managed_directories
-
+      def determine_purge_exclusions(managed_dirs)
         if environment && environment.respond_to?(:desired_contents)
-          exclusions += environment.desired_contents
+          managed_dirs + environment.desired_contents
+        else
+          managed_dirs
         end
-
-        @purge_exclusions = exclusions
       end
 
       def real_basedir
         Pathname.new(@basedir).cleanpath.to_s
       end
 
+      # For testing purposes only
+      def puppetfile_content(path)
+        File.read(path)
+      end
     end
   end
 end
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -7,7 +7,8 @@ module R10K
       include R10K::Logging
 
       attr_accessor :default_branch_override, :environment
-      attr_reader :modules, :managed_content, :moduledir
+      attr_reader :modules, :moduledir,
+        :managed_directories, :desired_contents, :purge_exclusions
 
       # @param [Hash] options
       # @option options [String] :puppetfile
@@ -25,6 +26,10 @@ module R10K
 
         @modules = []
         @managed_content = {}
+
+        @managed_directories = []
+        @desired_contents = []
+        @purge_exclusions = []
       end
 
       def load!
@@ -37,6 +42,17 @@ module R10K
         dsl.instance_eval(File.read(@puppetfile), @puppetfile)
 
         validate_no_duplicate_names(@modules)
+        set_managed_directories
+        set_desired_contents
+        set_purge_exclusions
+
+        {
+          modules: @modules,
+          managed_directories: @managed_directories,
+          desired_contents: @desired_contents,
+          purge_exclusions: @purge_exclusions
+        }
+
       rescue SyntaxError, LoadError, ArgumentError, NameError => e
         raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile})
       end
@@ -132,6 +148,28 @@ module R10K
         end
 
         true
+      end
+
+      def set_managed_directories
+        @managed_directories = @managed_content.keys.reject { |dir| dir == real_basedir }
+      end
+
+      # Returns an array of the full paths to all the content being managed.
+      # @return [Array<String>]
+      def set_desired_contents
+        @desired_contents = @managed_content.flat_map do |install_path, modnames|
+          modnames.collect { |name| File.join(install_path, name) }
+        end
+      end
+
+      def set_purge_exclusions
+        exclusions = managed_directories
+
+        if environment && environment.respond_to?(:desired_contents)
+          exclusions += environment.desired_contents
+        end
+
+        @purge_exclusions = exclusions
       end
 
       def real_basedir
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -1,0 +1,143 @@
+require 'r10k/logging'
+
+module R10K
+  module ModuleLoader
+    class Puppetfile
+
+      include R10K::Logging
+
+      attr_accessor :default_branch_override, :environment
+      attr_reader :modules, :managed_content, :moduledir
+
+      # @param [Hash] options
+      # @option options [String] :puppetfile
+      # @option options [String] :moduledir
+      # @option options [String] :basedir
+      # @option options [String] :forge
+      # @option options [Hash] :overrides
+      # @option options [R10K::Environment] :environment
+      def initialize(puppetfile:, moduledir:, forge:, basedir:, overrides:, environment:)
+        @puppetfile  = puppetfile
+        @moduledir   = moduledir
+        @basedir     = basedir
+        @overrides   = overrides
+        @environment = environment
+
+        @modules = []
+        @managed_content = {}
+      end
+
+      def load!
+        if !File.readable?(@puppetfile)
+          logger.debug _("Puppetfile %{path} missing or unreadable") % {path: @puppetfile.inspect}
+          return false
+        end
+
+        dsl = R10K::ModuleLoader::Puppetfile::DSL.new(self)
+        dsl.instance_eval(File.read(@puppetfile), @puppetfile)
+
+        validate_no_duplicate_names(@modules)
+      rescue SyntaxError, LoadError, ArgumentError, NameError => e
+        raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile})
+      end
+
+      # @param [String] forge
+      def set_forge(forge)
+        @forge = forge
+      end
+
+      # @param [String] moduledir
+      def set_moduledir(moduledir)
+        @moduledir = if Pathname.new(moduledir).absolute?
+          moduledir
+        else
+          File.join(@basedir, moduledir)
+        end
+      end
+
+      # @param [String] name
+      # @param [Hash, String, Symbol] args Calling with anything but a Hash is
+      #   deprecated. The DSL will now convert String and Symbol versions to
+      #   Hashes of the shape
+      #     { version: <String or Symbol> }
+      #
+      def add_module(name, args)
+        if !args.is_a?(Hash)
+          args = { version: args }
+        end
+
+        args[:overrides] = @overrides
+
+        if install_path = args.delete(:install_path)
+          install_path = resolve_install_path(install_path)
+          validate_install_path(install_path, name)
+        else
+          install_path = @moduledir
+        end
+
+        if @default_branch_override != nil
+          args[:default_branch_override] = @default_branch_override
+        end
+
+        mod = R10K::Module.new(name, install_path, args, @environment)
+        mod.origin = :puppetfile
+
+        # Do not load modules if they would conflict with the attached
+        # environment
+        if @environment && @environment.module_conflicts?(mod)
+          mod = nil
+          return @modules
+        end
+
+        # Keep track of all the content this Puppetfile is managing to enable purging.
+        @managed_content[install_path] = Array.new unless @managed_content.has_key?(install_path)
+        @managed_content[install_path] << mod.name
+
+        @modules << mod
+      end
+
+
+
+     private
+      # @param [Array<String>] modules
+      def validate_no_duplicate_names(modules)
+        dupes = modules
+                .group_by { |mod| mod.name }
+                .select { |_, v| v.size > 1 }
+                .map(&:first)
+        unless dupes.empty?
+          msg = _('Puppetfiles cannot contain duplicate module names.')
+          msg += ' '
+          msg += _("Remove the duplicates of the following modules: %{dupes}" % { dupes: dupes.join(' ') })
+          raise R10K::Error.new(msg)
+        end
+      end
+
+      def resolve_install_path(path)
+        pn = Pathname.new(path)
+
+        unless pn.absolute?
+          pn = Pathname.new(File.join(@basedir, path))
+        end
+
+        # .cleanpath is as good as we can do without touching the filesystem.
+        # The .realpath methods will also choke if some of the intermediate
+        # paths are missing, even though we will create them later as needed.
+        pn.cleanpath.to_s
+      end
+
+      def validate_install_path(path, modname)
+        unless /^#{Regexp.escape(real_basedir)}.*/ =~ path
+          raise R10K::Error.new("Puppetfile cannot manage content '#{modname}' outside of containing environment: #{path} is not within #{real_basedir}")
+        end
+
+        true
+      end
+
+      def real_basedir
+        Pathname.new(@basedir).cleanpath.to_s
+      end
+
+    end
+  end
+end
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -34,6 +34,7 @@ module R10K
         @forge       = forge
         @overrides   = overrides
         @environment = environment
+        @default_branch_override = @overrides.dig(:environments, :default_branch_override)
 
         @modules = []
 
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -116,10 +116,9 @@ module R10K
         mod = R10K::Module.new(name, install_path, module_info, @environment)
         mod.origin = :puppetfile
 
-        # Do not load modules if they would conflict with the attached
+        # Do not save modules if they would conflict with the attached
         # environment
         if @environment && @environment.module_conflicts?(mod)
-          mod = nil
           return @modules
         end
 
--- a/lib/r10k/puppetfile.rb
+++ b/lib/r10k/puppetfile.rb
@@ -5,10 +5,13 @@ require 'r10k/util/purgeable'
 require 'r10k/errors'
 require 'r10k/content_synchronizer'
 require 'r10k/module_loader/puppetfile/dsl'
+require 'r10k/module_loader/puppetfile'
 
 module R10K
 class Puppetfile
   # Defines the data members of a Puppetfile
+
+  NotGiven = BasicObject.new
 
   include R10K::Settings::Mixin
 
@@ -20,25 +23,17 @@ class Puppetfile
   #   @return [String] The URL to use for the Puppet Forge
   attr_reader :forge
 
-  # @!attribute [r] modules
-  #   @return [Array<R10K::Module>]
-  attr_reader :modules
-
   # @!attribute [r] basedir
   #   @return [String] The base directory that contains the Puppetfile
   attr_reader :basedir
-
-  # @!attribute [r] moduledir
-  #   @return [String] The directory to install the modules #{basedir}/modules
-  attr_reader :moduledir
 
   # @!attrbute [r] puppetfile_path
   #   @return [String] The path to the Puppetfile
   attr_reader :puppetfile_path
 
-  # @!attribute [rw] environment
+  # @!attribute [r] environment
   #   @return [R10K::Environment] Optional R10K::Environment that this Puppetfile belongs to.
-  attr_accessor :environment
+  attr_reader :environment
 
   # @!attribute [rw] force
   #   @return [Boolean] Overwrite any locally made changes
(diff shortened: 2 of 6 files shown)


CHANGELOG.mkd
```diff
@@ -4,6 +4,11 @@ CHANGELOG
 Unreleased
 ----------
 
+3.9.1
+-----
+
+- Invalid module specifications in a Puppetfile will cause the R10K run to abort earlier than before. Prior to this release the R10K run would complete, sync all other modules, and return an exit code of one, R10K will now stop syncing modules and abort immediately.
```
mwaggett: Is there a ticket associated with this change? If so, we should include it. If not, we should at least link to this PR for context.

CHANGELOG.mkd
```diff
@@ -4,6 +4,11 @@ CHANGELOG
 Unreleased
 ----------
 
+3.9.1
```
mwaggett: Should this be a feature release? I could be convinced either way.

CHANGELOG.mkd
```suggestion
- Invalid module specifications in a Puppetfile will cause the R10K run to abort earlier than before. Prior to this release, the R10K run would complete, sync all other modules, and return an exit code of 1. R10K will now stop syncing modules and abort immediately.
```

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+      DEFAULT_MODULEDIR = 'modules'
+      DEFAULT_PUPPETFILE_NAME = 'Puppetfile'
+      DEFAULT_FORGE_API = 'forgeapi.puppetlabs.com'
+
+      attr_accessor :default_branch_override, :environment
+      attr_reader :modules, :moduledir,
+        :managed_directories, :desired_contents, :purge_exclusions
+
+      # @param basedir [String] The path that contains the moduledir &
+      #     Puppetfile by default. May be an environment, project, or
```
mwaggett, line 14: How is the `basedir` used if the `moduledir` and Puppetfile are elsewhere?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        @managed_directories = determine_managed_directories(managed_content).freeze
+        @desired_contents = determine_desired_contents(managed_content).freeze
+        @purge_exclusions = determine_purge_exclusions(@managed_directories.clone).freeze
+
+        {
+          modules: @modules,
+          managed_directories: @managed_directories,
+          desired_contents: @desired_contents,
+          purge_exclusions: @purge_exclusions
+        }.freeze
```
mwaggett: Is this `freeze` necessary if we're not saving this hash as a variable anywhere?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile})
+      end
+
+
+      ##
+      ## set_forge, set_moduledir, and add_module are used directly by the DSL class
+      ##
+
+      # @param [String] forge
+      def set_forge(forge)
```
mwaggett, line 76: Any reason for this not to just be an `attr_accessor`?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        end
+      end
+
+      # @param [String] name
+      # @param [Hash, String, Symbol] args Calling with anything but a Hash is
+      #   deprecated. The DSL will now convert String and Symbol versions to
+      #   Hashes of the shape
+      #     { version: <String or Symbol> }
+      #
+      def add_module(name, args)
```
mwaggett: Can we call this param something more descriptive than `args`? The yardoc also doesn't really tell me what that param is for.

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        @moduledir = if Pathname.new(moduledir).absolute?
+          moduledir
+        else
+          File.join(@basedir, moduledir)
+        end
+      end
+
+      # @param [String] name
+      # @param [Hash, String, Symbol] args Calling with anything but a Hash is
+      #   deprecated. The DSL will now convert String and Symbol versions to
```
mwaggett: Is this a user-facing deprecation?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+          args[:default_branch_override] = @default_branch_override
+        end
+
+        mod = R10K::Module.new(name, install_path, args, @environment)
+        mod.origin = :puppetfile
+
+        # Do not load modules if they would conflict with the attached
+        # environment
+        if @environment && @environment.module_conflicts?(mod)
+          mod = nil
```
mwaggett: I don't think this is necessary?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        @modules << mod
+      end
+
+     private
+
+      # @param [Array<R10K::Module>] modules
+      def validate_no_duplicate_names!(modules)
+        dupes = modules
+                .group_by { |mod| mod.name }
+                .select { |_, v| v.size > 1 }
```
mwaggett: Can we use something more descriptive than `v`? maybe `modules_by_name` or something, so it's clear what's being counted?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+                .map(&:first)
+        unless dupes.empty?
+          msg = _('Puppetfiles cannot contain duplicate module names.')
```
mwaggett, line 138: Can you remind me what wrapping this in `_()` does? It's not very easy to google haha

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+          return @modules
+        end
+
+        @modules << mod
+      end
+
+     private
+
+      # @param [Array<R10K::Module>] modules
+      def validate_no_duplicate_names!(modules)
```
mwaggett: Why does this have a `!`? It doesn't modify `modules` at all, does it?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+      end
+
+      def resolve_install_path(path)
+        pn = Pathname.new(path)
+
+        unless pn.absolute?
+          pn = Pathname.new(File.join(@basedir, path))
+        end
+
+        # .cleanpath is as good as we can do without touching the filesystem.
```
mwaggett: as good as we can do for what? (I looked it up, so I know what `.cleanpath` does now, but since we're leaving this comment, I think it would be good to specify.)

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        # The .realpath methods will also choke if some of the intermediate
+        # paths are missing, even though we will create them later as needed.
+        pn.cleanpath.to_s
+      end
+
+      def validate_install_path!(path, modname)
```
mwaggett: Same question about `!` in the method name - it doesn't look like this modifies any of the args?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        if environment && environment.respond_to?(:desired_contents)
+          managed_dirs + environment.desired_contents
+        else
+          managed_dirs
+        end
+      end
+
+      def real_basedir
+        Pathname.new(@basedir).cleanpath.to_s
+      end
```
mwaggett: Is there a reason to have this as a separate method as opposed to just setting `@basedir = Pathname.new(basedir).cleanpath.to_s` on initialization?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+
+      # For testing purposes only
+      def puppetfile_content(path)
```
mwaggett, line 190: I still don't really see the purpose of this.. Why not just call `File.read` directly if this is just for testing?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        true
+      end
+
+      def determine_managed_directories(managed_content)
+        managed_content.keys.reject { |dir| dir == real_basedir }
+      end
+
+      # Returns an array of the full paths to all the content being managed.
+      # @return [Array<String>]
+      def determine_desired_contents(managed_content)
```
mwaggett, line 167: possibly out of scope for this if you just copied it from elsewhere, but I don't find `desired_contents` to be a particularly informative name for what we're actually collecting here.. `managed_content` and `managed_directories` are not much better..

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        end
+      end
+
+      def determine_purge_exclusions(managed_dirs)
+        if environment && environment.respond_to?(:desired_contents)
+          managed_dirs + environment.desired_contents
+        else
+          managed_dirs
+        end
+      end
```
mwaggett, line 179: Does this mean we exclude all modules from purging??

Magisus (approved):
So just to sum up, the goal of this is to establish the new API of a ModuleLoader, but deliberately does not update any consumers of module loading to use it; everything is still delegated via the `Puppetfile` class.

I think this makes sense as an intermediate step. Do you anticipate being able to switch over to calling things via this new API directly, at least internally, without breaking things for users? Is Bolt the only known consumer of r10k as a library that we need to worry about breaking?

lib/r10k/puppetfile.rb
```diff
@@ -67,142 +72,109 @@ def initialize(basedir, options_or_moduledir = nil, deprecated_path_arg = nil, d
     @moduledir       = deprecated_moduledir_arg || options.delete(:moduledir)       || File.join(basedir, 'modules')
     @puppetfile_name = deprecated_name_arg      || options.delete(:puppetfile_name) || 'Puppetfile'
     @puppetfile_path = deprecated_path_arg      || options.delete(:puppetfile_path) || File.join(basedir, @puppetfile_name)
+    @environment     = options.delete(:environment)
 
     @overrides       = options.delete(:overrides) || {}
+    @default_branch_override = @overrides.dig(:environments, :default_branch_overrides)
```
mwaggett: Should this be `:default_branch_override` singular? If so, seems not great if tests didn't catch that..

lib/r10k/puppetfile.rb
```diff
@@ -67,142 +72,109 @@ def initialize(basedir, options_or_moduledir = nil, deprecated_path_arg = nil, d
...
+    }
+
     @loaded = false
   end
 
-  def load(default_branch_override = nil)
-    return true if self.loaded?
-    if File.readable? @puppetfile_path
-      self.load!(default_branch_override)
+  def load(default_branch_override = NotGiven)
```
mwaggett: Why set up this `NotGiven` object instead of continuing to use `nil`?

lib/r10k/puppetfile.rb
```diff
@@ -67,142 +72,109 @@ def initialize(basedir, options_or_moduledir = nil, deprecated_path_arg = nil, d
...
-      logger.debug _("Puppetfile %{path} missing or unreadable") % {path: @puppetfile_path.inspect}
+      self.load!(default_branch_override)
     end
   end
 
-  def load!(default_branch_override = nil)
-    @default_branch_override = default_branch_override
+  def load!(dbo = NotGiven)
+    if (dbo != NotGiven) && (dbo != @default_branch_override)
+      logger.warn("Mismatch between passed and initialized default branch overrides, preferring passed value.")
```
mwaggett: I think this message could be a little more user-friendly. When would this situation arise? Where would these two values be coming from?

lib/r10k/puppetfile.rb
```diff
@@ -67,142 +72,109 @@ def initialize(basedir, options_or_moduledir = nil, deprecated_path_arg = nil, d
...
+      return @loaded_content
     else
-      logger.debug _("Puppetfile %{path} missing or unreadable") % {path: @puppetfile_path.inspect}
+      self.load!(default_branch_override)
     end
   end
 
-  def load!(default_branch_override = nil)
-    @default_branch_override = default_branch_override
+  def load!(dbo = NotGiven)
```
mwaggett: I'd love if we left `default_branch_override` instead of `dbo`.

lib/r10k/puppetfile.rb
```diff
@@ -67,142 +72,109 @@ def initialize(basedir, options_or_moduledir = nil, deprecated_path_arg = nil, d
...
+    if (dbo != NotGiven) && (dbo != @default_branch_override)
+      logger.warn("Mismatch between passed and initialized default branch overrides, preferring passed value.")
+      @loader.default_branch_override = dbo
```
mwaggett: Do we still want to set this if we can't read the Puppetfile and therefore never actually load?

spec/unit/action/deploy/module_spec.rb
```diff
@@ -202,14 +202,20 @@
...
+          loader.add_module('mod3', { git: 'git://remote' })
+
+          allow(loader).to receive(:puppetfile_content).and_return('')
+          loaded_content = loader.load!
+          puppetfile.instance_variable_set(:@loaded_content, loaded_content)
+          puppetfile.instance_variable_set(:@loaded, true)
         end
 
-        pf.modules.each do |mod|
+        loader.modules.each do |mod|
```
mwaggett: I'm confused about why this needed to change.. Shouldn't we still be able to act on the Puppetfile the way we were before? 
Are we intending to eventually get rid of the `Puppetfile` class altogether at some point? Even then, I feel like this change should go with that removal? Continuing to go through `Puppetfile` for now seems like it could be a good test of backwards compatibility?

Or am I just totally not understanding something?

spec/unit/action/deploy/module_spec.rb
```diff
@@ -182,7 +182,7 @@
     subject { described_class.new({ config: '/some/nonexistent/path' }, ['mod1', 'mod2'], {}) }
 
     let(:cache) { instance_double("R10K::Git::Cache", 'sanitized_dirname' => 'foo', 'cached?' => true, 'sync' => true) }
-    let(:repo) { instance_double("R10K::Git::StatefulRepository", cache: cache, resolve: 'main') }
+    let(:repo) { instance_double("R10K::Git::StatefulRepository", cache: cache, resolve: 'main', tracked_paths: []) }
```
mwaggett, line 185: What is `tracked_paths` for?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
+require 'spec_helper'
+require 'r10k/module_loader/puppetfile'
+
+describe R10K::ModuleLoader::Puppetfile do
+  describe 'initial parameters' do
+    describe 'honor' do
+      subject do
+        R10K::ModuleLoader::Puppetfile.new(basedir: '/test/basedir/env',
+                                           moduledir: '/test/basedir/env/dist-modules',
+                                           puppetfile: '/test/basedir/env/puppetfile.prod',
```
mwaggett: Would it be worth using values that are not inside the `basedir`, just to really be sure that we're not twiddling things we shouldn't?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+    end
+
+    it 'should not accept Forge modules with a version comparison' do
+      allow(R10K::Module).to receive(:new).with('puppet/test_module', subject.moduledir, hash_including(version: '< 1.2.0'), anything).and_call_original
+
+      expect {
+        subject.add_module('puppet/test_module', '< 1.2.0')
+      }.to raise_error(RuntimeError, /module puppet\/test_module.*doesn't have an implementation/i)
+
+      expect(subject.modules.collect(&:name)).not_to include('test_module')
```
mwaggett, line 107: since we're using the same `subject` for all these tests, I'm confused about how the 'test_module' from the previous test wouldn't still be in `subject.modules` for this test..

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: Or are these required to be inside the `basedir`? I saw that we validate that a module's `install_path` is inside the `basedir`, but I didn't see any of that validation for the `moduledir` or `puppetfile`.

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+      expect(subject.modules.collect(&:name)).to include('test_module')
+    end
+
+    it 'should reject non-Forge modules with an invalid relative :install_path option' do
+      module_opts = {
+        install_path: '../../vendor',
+        git: 'git@example.com:puppet/test_module.git',
+      }
+
+      allow(R10K::Module).to receive(:new).with('puppet/test_module', File.join(basedir, 'vendor'), module_opts, anything).and_call_original
```
mwaggett: I think we can remove this `allow`, right? since `validate_install_path!` gets called before we initialize the module, we would NOT expect this to get called at all, right?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+      expect { subject.add_module('puppet/test_module', module_opts) }.to raise_error(R10K::Error, /cannot manage content.*is not within/i).and not_change { subject.modules }
+    end
+
+    it 'should reject non-Forge modules with an invalid absolute :install_path option' do
+      module_opts = {
+        install_path: '/tmp/mydata/vendor',
+        git: 'git@example.com:puppet/test_module.git',
+      }
+
+      allow(R10K::Module).to receive(:new).with('puppet/test_module', File.join(basedir, 'vendor'), module_opts, anything).and_call_original
```
mwaggett: same re: removing this `allow`

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,190 @@
...
+        if @default_branch_override != nil
+          args[:default_branch_override] = @default_branch_override
+        end
+
+        mod = R10K::Module.new(name, install_path, args, @environment)
+        mod.origin = :puppetfile
+
+        # Do not load modules if they would conflict with the attached
+        # environment
+        if @environment && @environment.module_conflicts?(mod)
```
mwaggett, line 122: Should we be warning at all in this case?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+
+      expect { subject.add_module('puppet/test_module', module_opts) }.to raise_error(R10K::Error, /cannot manage content.*is not within/i).and not_change { subject.modules }
+    end
+
+    it 'should disable and not add modules that conflict with the environment' do
+      env = instance_double('R10K::Environment::Base')
+      mod = instance_double('R10K::Module::Base', name: 'conflict', origin: :puppetfile)
+      allow(mod).to receive(:origin=).and_return(nil)
```
mwaggett: What is this for?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+      loader = described_class.new(basedir: basedir, environment: env)
+      allow(env).to receive(:'module_conflicts?').with(mod).and_return(true)
+
+      allow(R10K::Module).to receive(:new).with('test', anything, anything, anything).and_return(mod)
+      expect { loader.add_module('test', {}) }.not_to change { loader.modules }
```
mwaggett: It's confusing to me that we're using 'test' as the name here, but we use 'conflict' inside `mod`.

justinstoller:
Bolt is the only user I know of and they don't use much of the API (I think they define an action and create a deployment).

I think it's possible to update the Environments to use the loaders directly and effectively abandoning the Puppetfile in a completely backwards compatible way. To be clear, the Puppetfile now uses the ModuleLoader, so all Puppetfile module loading goes through the ModuleLoader, the consumers are just unaware of the change.

I'm torn on how much the Environments refactor after that can or should be done in a backwards compatible way. The primary breaking change is that the class hierarchy will change. We currently have Environments descending from Base for two reasons: how the environment is backed (Git, Svn, Bare), vs how module content is loaded (WithModules). The change I'm thinking of would put the ModuleLoading, ContentSynchronizing, and Cleanup in the Base class. All environments could have Puppetfiles or extra modules defined, and Environments would be subclassed based on their backing store (Git, etc...). I believe the WithModules class and the Environment hierarchy was modified in R10K 3.4.0 though, so I don't think it's a huge deal to change it again in the 3 series. However that diff will be huge.

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+    end
+  end
+
+  describe '#purge_exclusions' do
+    let(:managed_dirs) { ['dir1', 'dir2'] }
+    subject { R10K::ModuleLoader::Puppetfile.new(basedir: '/test/basedir') }
+
+    it 'includes managed_directories' do
+      expect(subject.send(:determine_purge_exclusions, managed_dirs)).to match_array(managed_dirs)
```
mwaggett, line 179: Is the reason you need to use `send` because `determine_purge_exclusions` is private? 

I feel like it might make more sense to test the behavior we want to see rather than reaching into private methods? Like, as is, this test doesn't seem very valuable to me..

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+  describe '#managed_directories' do
+
+    let(:basedir) { '/test/basedir' }
+    subject { R10K::ModuleLoader::Puppetfile.new(basedir: basedir) }
+
+    before do
+      allow(subject).to receive(:puppetfile_content).and_return('')
+    end
+
+    it 'returns an array of paths that can be purged' do
```
mwaggett: Should this reference purging?? I was gonna blame copy pasta, but it doesn't even look like this is used in the `purge_exclusions` test.. ¯\_(ツ)_/¯

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+        subject.load!
+
+        expect(subject.modules.length).to be 1
+        expect(subject.managed_directories).to be_empty
+      end
+    end
+  end
+
+  describe 'evaluating a Puppetfile' do
+    def expect_wrapped_error(orig, pf_path, wrapped_error)
```
mwaggett: Would it make sense to call this param `error` instead of `orig`? And then maybe `wrapped_error` could be `wrapped_error_type`?

Magisus:
I like that description of the new class hierarchy 👍 I was always a bit thrown off by the `WithModules` subclassing. I don't feel like there's a strong reason to do it in a backwards-compatible way, so long as most/all user-facing behavior stays the same (which it should, as I understand it).

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+      expect { subject.load! }.not_to raise_error
+    end
+
+    it 'accepts a forge module without a version' do
+      @path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'valid-forge-without-version')
+      pf_path = File.join(@path, 'Puppetfile')
+      expect { subject.load! }.not_to raise_error
+    end
+
+    it 'creates a git module and applies the default branch sepcified in the Puppetfile' do
```
mwaggett:
```suggestion
    it 'creates a git module and applies the default branch specified in the Puppetfile' do
```

mwaggett (commented):
I realize a lot of these comments are about things that were already in here, so feel free to ignore, but figured I'd mention everything 🙃

lib/r10k/module_loader/puppetfile.rb
Magisus: This is for i18n, it's the string extraction function, like the `trs`/`tru` from our Clojure.

lib/r10k/module_loader/puppetfile.rb
Magisus: This is a clojurism that doesn't mean the same thing in Ruby, agree it should be dropped.

lib/r10k/module_loader/puppetfile.rb
mwaggett: ahh gotcha, no wonder I didn't find it by googling 😅

CHANGELOG.mkd
justinstoller: No, this is an internal refactor.

lib/r10k/module_loader/puppetfile.rb
justinstoller: Prior to this you could pass relative paths for these, however I took that out based on how convoluted the original implementation was. I've added that ability back (which leverages the basedir more thoroughly) and cleaned up the implementation.

lib/r10k/module_loader/puppetfile.rb
justinstoller: This isn't necessary. I was toying with different refactors that would make this class a one-shot loader, or would return an immutable object that contained loaded content. I made the results of loading immutable as a start and without a strong push towards one implementation or the other I haven't change it yet.

lib/r10k/module_loader/puppetfile.rb
justinstoller: I think it's good especially call out that this, `set_moduledir` and `add_module` are part of an interface contract with the DSL (vs, making an `attr_writer` and updating the DSL to call that).

lib/r10k/module_loader/puppetfile.rb
justinstoller: This is a deprecation that would affect those writing their own Puppetfile parsers but want to use most of our internals.

Previously, we would pass whatever crap users put after `mod` in a Puppetfile down to the Module implementations and hope they could figure out what was what. This will, in a future version, require Puppetfile DSL implementations to convert all that nonsense into a consumable hash with some consistent structure.

lib/r10k/module_loader/puppetfile.rb
justinstoller: Sure!

lib/r10k/module_loader/puppetfile.rb
justinstoller: I'm not clear enough on this behavior to know if users frequently ignore a large subset of Puppetfile content because of this (and are alright with that). The person who set users up with this workflow implemented it so I'm going to assume it's working okay for their use case. (Reid)

lib/r10k/module_loader/puppetfile.rb
justinstoller: Probably not, I think we used to unconditionally add the module and later on compact `@modules`. I'll remove it.

lib/r10k/module_loader/puppetfile.rb
justinstoller: A method that ends in a bang means that the method is more dangerous or surprising than one would expect. Typically in core classes it is used for methods that mutate the receiver. It is also used for methods that raise on failure instead of return data, or methods that may not be chainable in ways that similar methods are. Bang methods are always paired with a non-bang version. https://stackoverflow.com/questions/612189/why-are-exclamation-marks-used-in-ruby-methods

In Clojure the bang means a method should not be used in a transaction - I think we use it especially in file-sync for things that modify the file system (though I wouldn't be surprised if we use it in functions that raise as well). https://stackoverflow.com/questions/20606249/when-to-use-exclamation-mark-in-clojure-or-lisp

The exclamation mark is there because its a validation function that raises rather than returns data, which I think can be an accurate usage in Ruby, but it isn't paired with a non-bang version, so I agree it should be dropped.

lib/r10k/module_loader/puppetfile.rb
justinstoller: sure

lib/r10k/module_loader/puppetfile.rb
justinstoller: updated.

lib/r10k/module_loader/puppetfile.rb
justinstoller: Ditto to above. Raising without returning data is a valid reason to append an exclamation mark to a method name, but only if there's also a "safe" variant.

In that spirit I've also removed the exclamation mark from `load!` (the `R10K::Puppetfile` object has both `load!` and `load` but this object does not).

lib/r10k/module_loader/puppetfile.rb
justinstoller: Yeah, I don't either, but `managed_directories`, `desired_contents`, and `purge_exclusions` make up an interface for interacting with the Purgeable mixin that's used in several places. I think `determine_<interface name>` (unless you have a better word for "determine") is good.

lib/r10k/module_loader/puppetfile.rb
justinstoller: Yeah, a Puppetfile "manages" the moduledir and will purge anything inside it that _isn't_ a "module".

lib/r10k/module_loader/puppetfile.rb
justinstoller: I didn't like this before and went ahead and refactored it in response to your comment regarding basedir/moduledir/puppetfile. Let me know what you think.

lib/r10k/module_loader/puppetfile.rb
justinstoller: So specs can do:
```rb
test_puppetfile_content = <<EOPF
moduledir "my/moduledir"

mod "my-module", "1.0"
EOPF

expect(puppetfile).to receive(:puppetfile_content).and_return(test_puppetfile_content)
puppetfile.load
```

spec/unit/action/deploy/module_spec.rb
justinstoller: No, that makes sense. Though this (and the next test) aren't really testing Puppetfile compatibility, they really just have a bunch of setup so that we can test that when you call `r10k deploy module` with a list of modules and/or environments to deploy only those modules/environment combinations are deployed.

We just stub out all the Puppetfile interactions so we get realistic modules and can put expectations on their internal Repos.

Turns out the change itself wasn't needed and I changed it to `puppetfile`. I've also added comments explaining more.

spec/unit/action/deploy/module_spec.rb
justinstoller: Determining `desired_contents` on the environment which is now preemptively done in the `load` method rather than on demand during purging.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: I've refactored this test as part of addressing your first comment regarding basedir/moduledir/puppetfile parameters.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: `subject` is just a distinguished name in the specs. It is effectively
```ruby
let(:subject) <block>
```
where `<block>` if not set would default to:
```rb
R10K::ModuleLoader::Puppetfile.new
```
since that is the class passed into the top level `describe` block.
(and `let`s are memoized within a nested test block but are reset between them)

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: good catch!

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: It was a stub, I updated the `instance_double` line above to set it with the reader.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: good catch, I updated it.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: I think it makes sense to reference purging here though wording is weird and the test does a bit more than it really needs to. This is part of the inputs to the Purgeable mixin, which will find everything in `managed_directories` (here the moduledir) and then purge everything except `desired_contents` (modules) + `purge_exclusions` (things potentially generated by a post_run hook like `.generated_types` for environments). This is from the original Puppetfile specs.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: yeah

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: That's fair. These were existing tests on Puppetfile that I updated minimally to work with the ModuleLoader, I'll re-work them to be a bit higher level if I can.

lib/r10k/puppetfile.rb
justinstoller: This is largely for debugging/development purposes and I don't expect users to come across it in the wild.

lib/r10k/puppetfile.rb
justinstoller: I was unclear if `nil` had meaningful semantics as a `default_branch_override`. I don't think so, but I wanted to validate the case between a `nil` passed explicitly into `load` and the default value for it. I've removed it.

lib/r10k/puppetfile.rb
justinstoller: I want to set this at the ModuleLoader initialization and have passing this into deprecated. The warning above is largely for debugging/API usage. Likewise, the `NotGiven` construct was to ensure `nil` didn't have special semantics as an override. I've added docs deprecating this signature.

lib/r10k/puppetfile.rb
justinstoller: Yeah it should be. Though what this is doing is saving the initialized value for the conditional that warns below. With this bug we will always warn. I've corrected this and added tests around the default_branch_override behavior.

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,195 @@
...
+      #     Puppetfile by default. May be an environment, project, or
+      #     simple directory.
+      # @param puppetfile [String] The path to the Puppetfile, either an
+      #     absolute full path or a relative path with regards to the basedir.
+      # @param moduledir [String] The path to the moduledir, either an
+      #     absolute full path or a relative path with regards to the basedir.
+      # @param forge [String] The url (without protocol) to the Forge
+      # @param overrides [Hash] Configuration for loaded modules' behavior
+      # @param environment [R10K::Environment] The environment loading may be
+      #     taking place within
```
mwaggett:
```suggestion
      # @param environment [R10K::Environment] When provided, the environment in which loading
      #     takes place
```
or something like that, to make clear that loading definitely (not may) happens within the environment, when that param is provided.

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,195 @@
...
+        dsl.instance_eval(puppetfile_content(@puppetfile), @puppetfile)
+
+        validate_no_duplicate_names(@modules)
+        @modules.freeze
+
+        managed_content = @modules.group_by(&:dirname).freeze
+
+        @managed_directories = determine_managed_directories(managed_content).freeze
+        @desired_contents = determine_desired_contents(managed_content).freeze
+        @purge_exclusions = determine_purge_exclusions(@managed_directories.clone).freeze
```
mwaggett: Why does this need to be cloned?

lib/r10k/module_loader/puppetfile.rb
mwaggett: hmm okay. so then could you maybe update your comment to say that? as it is, I don't really see why we couldn't just update the DSL to call an `attr_writer`.

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -0,0 +1,195 @@
...
+        module_info[:overrides] = @overrides
+
+        if install_path = module_info.delete(:install_path)
+          install_path = resolve_path(@basedir, install_path)
+          validate_install_path(install_path, name)
+        else
+          install_path = @moduledir
+        end
+
+        if @default_branch_override != nil
```
mwaggett:
```suggestion
        if @default_branch_override
```

lib/r10k/module_loader/puppetfile.rb
mwaggett: fair, maybe at least some debug logging would be useful?

lib/r10k/module_loader/puppetfile.rb
mwaggett: that works for me!

lib/r10k/module_loader/puppetfile.rb
mwaggett: okay that makes sense

lib/r10k/puppetfile.rb
```diff
@@ -67,142 +70,118 @@ def initialize(basedir, options_or_moduledir = nil, deprecated_path_arg = nil, d
...
 
+  # @param [String] default_branch_override The default branch to use
+  #   instead of one specified in the module declaration, if applicable.
+  #   Deprecated, use R10K::ModuleLoader::Puppetfile directly and pass
+  #   the default_branch_override as an option on initialization.
   def load!(default_branch_override = nil)
-    @default_branch_override = default_branch_override
+    if default_branch_override && (default_branch_override != @default_branch_override)
+      logger.warn("Mismatch between passed and initialized default branch overrides, preferring passed value.")
+      @loader.default_branch_override = default_branch_override
```
mwaggett: since GH doesn't seem to want to let me reply to my previous comment, Do we still want to set this if we can't read the Puppetfile and therefore never actually load?

Your reply referenced the conditional above, but I'm asking about the conditional below where we check to see if the Puppetfile is readable. I'm wondering if setting `@loader.default_branch_override = default_branch_override` should come after that check. Does that make sense? Apologies if I'm just misunderstanding your reply.

lib/r10k/puppetfile.rb
```diff
@@ -67,142 +70,118 @@ def initialize(basedir, options_or_moduledir = nil, deprecated_path_arg = nil, d
...
   include R10K::Util::Purgeable
 
   def managed_directories
-    self.load unless @loaded
+    self.load
 
-    dirs = @managed_content.keys
-    dirs.delete(real_basedir)
-    dirs
+    @loader.managed_directories
```
mwaggett: Should this be `@loaded_content[:managed_directories]` instead of reaching into the `@loader`, like we do for `desired_contents` and `purge_exclusions` below? I assume both would work, but I figure we should be consistent unless there's a reason to prefer one over the other.

spec/unit/action/deploy/module_spec.rb
mwaggett: I'm just not seeing how/where it's being used..

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,330 @@
...
+      let(:options) do
+        {
+          basedir: '/test/basedir/env',
+          forge: 'localforge.internal.corp',
+          overrides: { modules: { deploy_modules: true } },
+          environment: R10K::Environment::Git.new('env',
+                                                  '/test/basedir/',
+                                                  'env',
+                                                  { remote: 'git://foo/remote',
+                                                    ref: 'env' })
```
mwaggett, line 16: (nit) might be good to use a ref that's different from the environment name?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,330 @@
...
+          expect(puppetfile.instance_variable_get(:@puppetfile)).to eq('/opt/puppetlabs/special/Puppetfile')
+        end
+
+        it 'roots the Puppetfile in the basepath if a relative path is specified' do
+          relative_options = options.merge({puppetfile: 'Puppetfile.global'})
+          puppetfile = R10K::ModuleLoader::Puppetfile.new(**relative_options)
+          expect(puppetfile.instance_variable_get(:@puppetfile)).to eq('/test/basedir/env/Puppetfile.global')
+        end
+      end
+
```
mwaggett, line 49: Could we move setting `subject` to down here, since the above tests don't use it?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,330 @@
...
+    end
+  end
+
+  describe 'adding modules' do
+    let(:basedir) { '/test/basedir' }
+
+    subject { R10K::ModuleLoader::Puppetfile.new(basedir: basedir) }
+
+    it 'should transform Forge modules with a string arg to have a version key' do
+      allow(R10K::Module).to receive(:new).with('puppet/test_module', subject.moduledir, hash_including(version: '1.2.3'), anything).and_call_original
```
mwaggett: apologies as it's been a while since I've had to write rspec tests, but why do we need this `allow`? We don't seem to be doing anything special that the original method doesn't do?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,330 @@
...
+
+      expect { subject.add_module('puppet/test_module', '1.2.3') }.to change { subject.modules }
+      expect(subject.modules.collect(&:name)).to include('test_module')
+    end
+
+    it 'should not accept Forge modules with a version comparison' do
```
mwaggett, line 100: Is that just something we've never supported?

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: oh, so even though it's defined within the `describe`, each `it` block resets the value?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,330 @@
...
+        install_path: '/tmp/mydata/vendor',
+        git: 'git@example.com:puppet/test_module.git',
+      }
+
+      expect { subject.add_module('puppet/test_module', module_opts) }.to raise_error(R10K::Error, /cannot manage content.*is not within/i).and not_change { subject.modules }
+    end
+
+    it 'should disable and not add modules that conflict with the environment' do
+      env = instance_double('R10K::Environment::Base')
+      mod = instance_double('R10K::Module::Base', name: 'conflict', origin: :puppetfile, 'origin=': nil)
```
mwaggett, line 165: What does this `'origin=': nil` bit do?

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -0,0 +1,330 @@
...
+      loader = described_class.new(basedir: basedir, environment: env)
```
mwaggett: Any reason you use `described_class` here instead of directly instantiating `R10K::ModuleLoader::Puppetfile` like you do elsewhere?

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: ahh okay, I think I understand. so then what if we said something like 'returns an array of directories that could potentially be purged'?

lib/r10k/module_loader/puppetfile.rb
mwaggett: yeah, I'm just not really understanding why that's even necessary. This isn't acting on anything within this class.. 

In this example, you're acting on `puppetfile`, but this is the `loader` class.. Should this method go in the `R10K::Puppetfile` class? Or should this not take a param and instead call `File.read(@puppetfile)`?

lib/r10k/module_loader/puppetfile.rb
justinstoller: What - in addition to the comment four lines above this - would like to see?

spec/unit/action/deploy/module_spec.rb
justinstoller: When we call `determine_purge_exclusions` we ask the environment for its `desired_contents`. A Git backed environment will call `tracked_paths` on it's backing repo (here: https://github.com/puppetlabs/r10k/blob/main/lib/r10k/environment/git.rb#L90)

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: yep!

lib/r10k/module_loader/puppetfile.rb
justinstoller: Looking into it more it seems that the `module_conflicts?` method handles that already https://github.com/puppetlabs/r10k/blob/main/lib/r10k/environment/with_modules.rb#L52

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: Okay, updated locally to be:
"returns an array of paths that #purge! will operate within"

lib/r10k/module_loader/puppetfile.rb
justinstoller: accepted locally

lib/r10k/module_loader/puppetfile.rb
justinstoller: The method that `determine_purge_exclusions` is based on used to mutate the argument. I did this, then updated it to no longer mutate the argument. I'll get rid of the clone.

lib/r10k/puppetfile.rb
justinstoller: No, that makes sense. Really my reply was "I don't care if it's not in the best spot because I want to deprecate and remove it when I can anyways". I updated the code earlier to try to be more clear in my intentions but I'll update the location in the meantime.

lib/r10k/puppetfile.rb
justinstoller: good catch

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: The ref is usually the branch name that backs the environment, which is also the environment name. So while this is confusing, it is accurate, here's the code that creates Git environments (where `bn` represents the "branch name" object:
```rb
        envs << R10K::Environment::Git.new(bn.name,
                                           @basedir,
                                           bn.dirname,
                                           {remote: remote,
                                            ref: bn.name,
                                            puppetfile_name: puppetfile_name,
                                            overrides: @options[:overrides]})
```

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: I don't know, do we have guidance around that?

I have a fairly ingrained habit of putting (and looking for) `before`s, `let`s, etc at the top of the describe/context block they're active for.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: I think the original author meant to use `expect` though once you put argument expectations on an `allow` I think it works essentially the same. I'll update this to `expect`.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: I believe the Forge metadata and potentially librarian-puppet (an r10k precursor) did. But I think it has always been out of r10k's scope.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: each key becomes a method that takes any args and returns the value given. This defines a stub setter for `origin`.

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: Nope! I'll normalize that.

lib/r10k/module_loader/puppetfile.rb
justinstoller: So the fact that it takes a param instead of acting on an instance variable is the confusing part? It's six of one, half dozen of the other to me. Though I typically like methods parameterized as much as possible, which is why it has a param.

Here is an example of how this is used in a test: https://github.com/puppetlabs/r10k/pull/1167/files#diff-1bc365ddca4e6e4a7ba95ff4d655c036faa16ec1025e35bc559476265a5719aaR292

spec/unit/module_loader/puppetfile_spec.rb
justinstoller: Did I not do this?

justinstoller: Looks like no. I believe after I push my local changes up, this will be the one outstanding change that I still need to make, however I won't get to it till next week.

lib/r10k/module_loader/puppetfile.rb
mwaggett: Is there a reason we're treating the DSL's usage as an API contract? i.e. _why_ can't we update the DSL to call a different method?

Or, to put it another way, why is it especially important to note that these methods are used by the DSL?

spec/unit/action/deploy/module_spec.rb
mwaggett: ah, so previously we were getting away with not including `tracked_paths` here because this test used to not reach a point where we needed the `desired_contents`, yes? this is just some stubbing weirdness and a real `R10K::Git::StatefulRepository` would have always had `tracked_paths`, even before this PR?

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: cool, thanks for explaining!

lib/r10k/module_loader/puppetfile.rb
mwaggett: ahh nice, thanks for finding that.

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: ahh, yeah, that makes sense, thanks

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: I don't know of any guidance / style guide around that kind of thing. I'm fine leaving it if you prefer. My habit is to just look for things like that nearby the things that use it. 😅 Could just switch the order of the tests haha, but doesn't really matter.

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: yeah, the `allow` vs. `expect` and how they exactly behave has always confused me, so using the one that semantically expresses the intention is my preference. 😁

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: gotcha, just making sure this wasn't new from this PR

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: ahh, okay, so since we're stubbing this, `R10K::Module::Base` doesn't have its normal methods defined, and we need the `origin` setting for `add_module`. Is that an accurate understanding of why this is needed?

lib/r10k/module_loader/puppetfile.rb
mwaggett: No, that's not what bothering/confusing me.

I was mostly objecting to the premise of adding a class method just for testing purposes, especially since I wasn't really understanding how calling it in the tests was actually doing anything, but now that I look again, I see that this method actually _is_ being used within the `load` method, not just in tests, in which case I think you should remove this comment, since it clearly confused me a lot lol.

lib/r10k/module_loader/puppetfile.rb
Magisus: I'm with Justin, I think the clarity of having them be "real" methods helps with realizing that they're some kind of API. There's too much wishy-washiness as it is with that. In addition, not all of them are pure setters. So it's kind of nice to see them as a set here.

Plus, I feel like updating the usage of the DSL class is outside the scope of this, besides.

lib/r10k/module_loader/puppetfile.rb
Magisus: I agree this makes things better for testing, it gets messy stubbing things like `File.read` because once you do it once in a test, you have to do it for all instances of it being called, which often means stubbing a lot more than is strictly necessary. So far as I can tell, it _is_ called up above, where it's being passed the `puppetfile` instance var. The little abstraction over `File.read` in that context is kinda nice actually, IMO.

lib/r10k/module_loader/puppetfile.rb
mwaggett: yeah, that's fine, I just want it to be clear that we're treating this as an API contract. so the comment could say something like
> The `set_forge`, `set_moduledir`, and `add_module` methods must be defined as part of an API contract with the DSL class.`

mwaggett: Am I making sense?

spec/unit/module_loader/puppetfile_spec.rb
Magisus: I'd much rather `subject` be set at the top, that's definitely the way it's done everywhere I've seen.

lib/r10k/module_loader/puppetfile.rb
mwaggett: yeah, I'm totally fine with this method now that I see how it's being used. The "for testing purposes only" comment totally threw me off.

Magisus (approved):
I think I'm good on this.

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: Is this still something you're planning to do as part of this PR?

spec/unit/action/deploy/module_spec.rb
Magisus: Yeah.

spec/unit/module_loader/puppetfile_spec.rb
Magisus: Fwiw, as much as I agree that this test isn't super useful, I also think it would be good to get this in, so if this is the only pending comment, I'm inclined to leave writing new tests for another time. We'll get better validation running it in real environments.

mwaggett: Do we want/need a ticket for writing better automated tests?

The only other pending comments were around code comments that I found confusing, one of them being straight up not true (the one about the `puppetfile_content` method). so, not the biggest deal, but could be extra confusing for future us. ¯\_(ツ)_/¯

mwaggett:
@justinstoller some of the commits should probably be squashed?

justinstoller:
I'm fine with the commits if y'all are. The whole PR is a refactor and each commit feels fairly contained to a functional aspect, set of related methods, or theme.

PR #1169, opened 2021-06-17
(maint) Fix forge module implementation def bug

reidmv:
The implement?() method for Forge modules would fail if the args weren't a hash. Fix bug, since for Forge modules specifically, args can sometimes be a scalar.

### DIFF
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -13,12 +13,16 @@ class R10K::Module::Forge < R10K::Module::Base
   R10K::Module.register(self)
 
   def self.implement?(name, args)
-    args[:type].to_s == 'forge' ||
-      (!!
-       ((args.keys & %i{git svn type}).empty? &&
-        args.has_key?(:version) &&
-        name.match(%r[\w+[/-]\w+]) &&
-        valid_version?(args[:version])))
+    if args.is_a?(Hash)
+      args[:type].to_s == 'forge' ||
+        (!!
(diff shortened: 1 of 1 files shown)


justinstoller:
Is this being instantiated using `Module.new` with a single string parameter outside of the Puppetfile class?

Wondering the workflow where you ran into this?

reidmv:
Oh wait... it looks like we're moving to make specifying a module with a non-hash params argument obsolete... so, maybe this doesn't need to validate. I'm gonna go that direction instead.

reidmv:
> Is this being instantiated using `Module.new` with a single string parameter outside of the Puppetfile class?
> 
> Wondering the workflow where you ran into this?

Yeah, I ran into this outside of Puppetfile. I peeled the first commit off of #1159 when I was rebasing it so I don't remember exactly what I was doing, but it was probably specifying a Forge module in a yaml file trying to use a scalar arg.

I'm perfectly happy just saying that in the "new" place, yaml files, you don't get to do that. You have to specify hash args.

I'm also going to suggest going one step further and requiring a type key for Forge modules, and not assuming Forge just because there's no other type. Will file that suggestion as a PR so it can be visualized.

justinstoller:
Cool.

FWIW, I'm happy to put it back for the remainder of the 3.x series. But I do think requiring a Hash argument is what we'll want in 4.0. It took a bit of spelunking to figure out that every Module implementation had to duck type the args because one (and what I thought was only one) of them also allowed a String instead of a Hash.

PR #1170, opened 2021-06-17
(maint) Simplify forge module implementation identification

reidmv:
Anywhere except the Puppetfile, a module input must specify {type: forge} to indicate a Forge module. In the Puppetfile, we convert a scalar version arg to a hash already, before passing it to R10K::Module.new(). So, let's simplify the implementation selection logic.

In the Puppetfile, specify both version: and type: for scalar module args, which by convention we know to be Forge modules.

End result: internally, modules args should ALWAYS be of type Hash. Puppetfile module specifications, which permit a String form of args, now convert found string args to hashes on intake.

### DIFF
--- a/lib/r10k/environment/with_modules.rb
+++ b/lib/r10k/environment/with_modules.rb
@@ -96,7 +96,7 @@ class R10K::Environment::WithModules < R10K::Environment::Base
   def load_modules(module_hash)
     module_hash.each do |name, args|
       if !args.is_a?(Hash)
-        args = { version: args }
+        args = { type: 'forge', version: args }
       end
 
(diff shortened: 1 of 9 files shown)


Magisus:
I think some of the spots this touches are being changed by @justinstoller 's ongoing refactors. If you're still interested in doing this, it might need to be revisited when those have finished.

reidmv:
@Magisus revisited and re-worked post Justin's refactors.

Magisus (approved):

Magisus:
I'll push up a changelog entry for this separately (it would create conflicts, and I want to get this in).

Issue #1173, opened 2021-06-25
(Initial investigation) PR #1161 breaks all controlrepos that use `control_branch`


dylanratcliffe:
As of the current version the R10K::Puppetfile#add_module method does [the following](https://github.com/puppetlabs/r10k/blob/main/lib/r10k/puppetfile.rb#L155-L157) which was introduced in #1161 :

```ruby
    if @default_branch_override != nil
      args[:default_branch_override] = @default_branch_override
    end
```

This means that all module classes will have the `:default_branch_override` argument set if the `R10K::Puppetfile#load` was called with a non-nil `default_branch_override`. The `:default_branch_override` option is file for `R10K::Module::Git` modules but not for `R10K::Module::Forge for example and leats to deployment errors such as:

```
2021-06-25T14:07:54.515+02:00 INFO  [qtp1153833797-509] [p.c.app] Queuing deploy for production_cdpe_ia_1624622874129
2021-06-25T14:07:54.530+02:00 INFO  [deploy-pool-1] [p.c.core] Attempting to deploy :environment 'production_cdpe_ia_1624622874129'...
2021-06-25T14:07:55.225+02:00 ERROR [deploy-pool-1] [p.c.core] Errors while deploying environment 'production_cdpe_ia_1624622874129' (exit code: 1):
/opt/puppetlabs/puppet/lib/ruby/gems/2.7.0/gems/rugged-0.27.7/lib/rugged/repository.rb:258: warning: Using the last argument as keyword parameters is deprecated
ERROR	 -> Failed to evaluate /etc/puppetlabs/code-staging/environments/production_cdpe_ia_1624622874129/Puppetfile
Original exception:
R10K::Module::Forge cannot handle option 'default_branch_override'
2021-06-25T14:07:55.406+02:00 INFO  [deploy-pool-1] [p.c.core] Finished deploy attempt for :environment 'production_cdpe_ia_1624622874129'.
```

## Steps to Reproduce

I won't have time to do a replication case but I suspect any Puppetfile that contains something like this will cause it:

```ruby
mod 'puppetlabs/sqlserver',                   :latest

mod 'someone/something',
    git: 'git@somewhere.com/whatever',
    branch: :control_branch,
    default_branch: 'main'
```

## Environment

PE 2021.2.0

## Additional Context

This is in a released version of PE and will affect a great many customers. Let's get onto this ASAP

Magisus:
Thanks Dylan, we're taking a look!

[...]

PR #1174, opened 2021-06-25
(#1173) Allow default_branch_override argument for all module types

justinstoller:
In 1ff9995, we moved the setting of the `default_branch_override` from
all module declarations that were Hashes (ie, not Puppetfile declared
Forge modules) to all modules. This broke Forge modules for users who
pass the `default_branch_override` flag.

It should be noted that code-wise, users who set the
`default_branch_override` flag when deploying YAML based environments
(which always use Hash based module declarations) or had SVN modules in
their Puppetfile (also Hash based) would have failed. However, those
scenarios are either not supported (in the case of YAML based envs) or
essentially unused (in the case of SVN).

Other options to fix this regretion include:
  * Have the caller of Module.new first determine if the module will be
  a Git module and then conditionally add the option, or
  * Have the caller of Module.new set the default_branch_override after
  creating the module, if the module responds to the setter

Both of these options involve the caller having to introspect enough
about the type of module being created that it destroys what value we
get from the Module.new facade. Consequently, this patch takes the
approach to loosen Module argument strictness, allowing potentially
unused options to passed to any module.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,7 +4,13 @@ CHANGELOG
 Unreleased
 ----------
 
+3.9.3
+-----
+
+- Fixes a regression when using `--default_branch_override` with Puppetfiles containing Forge modules. [#1173](https://github.com/puppetlabs/r10k/issues/1173)
+
 3.9.2
+-----
 
 - Makes the third parameter to R10K::Actions optional, restoring backwards compatability broken in 3.9.1.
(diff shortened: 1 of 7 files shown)


Magisus (commented):
Code changes look good to me 👍

Magisus (approved):

mwaggett (approved):

PR #1182, opened 2021-07-08
Always sync git module on ref HEAD

skoef:
This PR solves Jira ticket [RK-323](https://tickets.puppetlabs.com/projects/RK/issues/RK-323) for which no GH was made yet.

It allows you to always checkout the latest available version of a Git-type module in Puppetfile by setting `:ref` to `:latest`. Basically what it does is force a sync of the cache and update the desired checkout to `HEAD` while resolving to a specific commit.

### DIFF
--- a/doc/puppetfile.mkd
+++ b/doc/puppetfile.mkd
@@ -130,7 +130,8 @@ with a git commit, branch reference, or a tag.
 When a module is installed using `:tag` or `:commit`, r10k assumes that the
 given object is a tag or commit and can do some optimizations around fetching
 the object. If the tag or commit is already available r10k will skip network
-operations when updating the repo, which can speed up install times.
+operations when updating the repo, which can speed up install times. When
+`:commit` is set to track `HEAD`, it will synchronize the module on each run.
 
 Module versions can also be specified using `:branch` to track a specific
 branch reference.
--- a/doc/puppetfile.mkd
+++ b/doc/puppetfile.mkd
@@ -131,7 +131,7 @@ When a module is installed using `:tag` or `:commit`, r10k assumes that the
 given object is a tag or commit and can do some optimizations around fetching
 the object. If the tag or commit is already available r10k will skip network
 operations when updating the repo, which can speed up install times. When
-`:commit` is set to track `HEAD`, it will synchronize the module on each run.
+`:ref` is set to track `HEAD`, it will synchronize the module on each run.
 
 Module versions can also be specified using `:branch` to track a specific
 branch reference.
--- a/lib/r10k/git/stateful_repository.rb
+++ b/lib/r10k/git/stateful_repository.rb
@@ -93,6 +93,7 @@ class R10K::Git::StatefulRepository
   # @api private
   def sync_cache?(ref)
     return true if !@cache.exist?
+    return true if ref == 'HEAD'
     return true if !([:commit, :tag].include? @cache.ref_type(ref))
     return false
   end
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,8 @@ CHANGELOG
 Unreleased
 ----------
 
+- Always sync git cache on `ref: 'HEAD'` [#1182](https://github.com/puppetlabs/r10k/pull/1182)
+
 3.10.0
 ------
 
--- a/spec/unit/git/stateful_repository_spec.rb
+++ b/spec/unit/git/stateful_repository_spec.rb
@@ -19,6 +19,11 @@ describe R10K::Git::StatefulRepository do
       expect(subject.sync_cache?(ref)).to eq true
     end
 
+    it "is true if the ref is HEAD" do
+      expect(cache).to receive(:exist?).and_return true
+      expect(subject.sync_cache?('HEAD')).to eq true
+    end
+
     it "is true if the ref is unresolvable" do
       expect(cache).to receive(:exist?).and_return true
       expect(cache).to receive(:ref_type).with('0.9.x').and_return(:unknown)


skoef:
Given a Puppetfile like this:
```
mod 'foo',
    :git => 'git@github.com:skoef/foo.git'
mod 'bar',
    :git => 'git@github.com:skoef/bar.git',
    :ref => :latest
```

running r10k with `--verbose=debug` outputs:

```
[2021-07-08 09:45:54 - DEBUG] /etc/puppetlabs/code/environments/testing/modules/foo is already at Git ref master
[2021-07-08 09:45:54 - DEBUG] Module thread 47216422044020 exiting: queue empty
[2021-07-08 09:45:54 - DEBUG] /etc/puppetlabs/code/environments/testing/modules/bar is already at Git ref bd5ff0c095d3a7588cf6eef5939453845d4039da
[2021-07-08 09:45:54 - DEBUG] Module thread 47216422043860 exiting: queue empty
```

mwaggett:
hey @skoef thanks for your contribution!

Gonna need to check in with our Product folks around this, but my personal opinion is that `:latest` is potentially misleading. Latest for what branch? I assume the default branch of the repo, but I think it's generally better to be explicit about what branch you want. 

Also, this doesn't _really_ solve RK-323 - you'll still have this problem if you're using a branch as the desired `ref`.

I think the better solution would be to fix this for branches and leave out the `:latest` option.

I believe we've had conversations around this kind of thing in the past, but I'd love to hear your opinions @npwalker @reidmv.

npwalker:
Is HEAD not already an allowed reference?

mwaggett:
I believe it is, based on this issue, which has some context: https://github.com/puppetlabs/r10k/issues/1110

lib/r10k/git/stateful_repository.rb
```diff
@@ -31,6 +31,10 @@ def initialize(remote, basedir, dirname)
   end
 
   def resolve(ref)
+    if ref == :latest
+      @cache.sync!
+      ref = "HEAD"
+    end
     @cache.sync if sync_cache?(ref)
```
reidmv: Shouldn't this line already be doing what the stanza above does?

1. This calls `#sync` on the cache IF `#sync_cache?` returns true for `ref`.  
Note that `R10K::Git::Cache#sync` is just a wrapper for `R10K::Git::Cache#sync!` with a guard against multiple sync activities in the same r10k run.
2. `#sync_cache?` [returns true](https://github.com/puppetlabs/r10k/blob/e27a5d14c05324a89820ef8c382f63cb407410fb/lib/r10k/git/stateful_repository.rb#L96) if the cache doesn't exist, or if the ref type is NOT `:commit` or `:tag`, according to  [`R10K::Git::Cache#ref_type`](https://github.com/puppetlabs/r10k/blob/7d8b5b51b533ea3c45016d54e837aecc6f657588/lib/r10k/git/rugged/base_repository.rb#L36-L49).
3. Therefore, `@cache.sync` should already be being called here. If it isn't, that's a bug that needs to be hunted down and fixed, not a change in intended behavior.

skoef:
I digged deeper into this today. @reidmv mentions that the cache should already be updated, since it exists and the ref type is NOT `:commit` or `:tag`. I'm working with a fresh checkout of r10k 3.10.0 and when I would use `HEAD` rather then `:latest`, it *does* resolve to `:commit` in `ref_type()` so the cache is not synced. In `ref_type` what happens is `git rev-parse HEAD^{commit}` basically and that will return the commit of HEAD at the time.

So to me, this does seem buggy. Instead of introducing the `:latest` a better fix imho would be to change https://github.com/puppetlabs/r10k/blob/e27a5d14c05324a89820ef8c382f63cb407410fb/lib/r10k/git/stateful_repository.rb#L96

to
```ruby
return true if !([:commit, :tag].include? @cache.ref_type(ref)) || ref == "HEAD"
```

This also solves my issue and is perhaps much cleaner. Let me know if you guys can confirm this bug and if you want me to update the MR to address this particular issue!

Magisus:
I think that makes sense to me! `HEAD` is a floating ref, so we would want to make sure we update it every time.

skoef:
@reidmv and @Magisus how about this?

reidmv (approved):
LGTM! 👍

doc/puppetfile.mkd
```diff
@@ -130,7 +130,8 @@ with a git commit, branch reference, or a tag.
 When a module is installed using `:tag` or `:commit`, r10k assumes that the
 given object is a tag or commit and can do some optimizations around fetching
 the object. If the tag or commit is already available r10k will skip network
-operations when updating the repo, which can speed up install times.
+operations when updating the repo, which can speed up install times. When
+`:commit` is set to track `HEAD`, it will synchronize the module on each run.
```
mwaggett: The sync will happen regardless of what key you use (`:ref`, `:commit`, `:branch`, etc.), right?
```suggestion
`:ref` is set to track `HEAD`, it will synchronize the module on each run.
```

mwaggett (commented):
I assume that on a first sync, `HEAD` will refer to the head of the default branch.
But what happens if you switch from referencing a specific commit or tag to referencing `HEAD`? I'm thinking it might not get updated, if I'm understanding `HEAD` behavior correctly.

doc/puppetfile.mkd
reidmv: True. All of `:ref`, `:commit`, and `:branch` are really just syntactic sugar for the same thing these days.

Magisus (approved):
Okay, I just verified that this works as described in your latest message, namely that if you have a git module and are tracking `HEAD`, it will now correctly try to pull the latest commits from the default branch on every run. (Note that this is subject to the weirdness around "default branch", which is [always `master`](https://github.com/puppetlabs/r10k/blob/cc238496452fe3e7819de9380110ccffe2fd84b2/lib/r10k/module/git.rb#L60) regardless of what GH considers the default.)

Magisus:
I'm going to resolve the changelog conflict and merge this. Thanks for the contribution!

Magisus:
I also tested that scenario Molly raised, and this is what it does:
* Going from `HEAD` to a branch switches you from the `HEAD` of `master` to the newly specified branch
* Going back to `HEAD` from that branch puts you back on the `HEAD` of `master`.

Whether that's expected or not, I feel like it's at least consistent: `HEAD` will always refer to the head of `master`, and r10k will correctly swap to and from it.

mwaggett (approved):

skoef:
Thanks @mwaggett!

PR #1183, opened 2021-07-09
(maint) Fix broken acceptance test

justinstoller:
Previously, the single_env_purge_unmanaged_modules test said that
it was testing that agent runs would fail because the motd module
had been successfully purged.

However, it was actually testing that a) the purge command failed and
b) the agent run succeeded but the assertion failed, not because of
anything that happened on SUT, but because it was testing the output
against the regex /Blah/ which should not appear in normal agent output.
The test then claimed that the failure was part of CODEMGMT-78, which
was a feature improvement to `r10k puppetfile` to respect `--moduledir`
and `--puppetfile` flags. However, that feature was implemented in 2016
and the test has continued to "pass" because of a failing purge command
because the test never actually attempted to use either the moduledir
or puppetfile flags.

This patch updates the purge invocation to actually use moduledir and
puppetfile flags, updates the expectation on the purge invocation to
be successful. It also updates the expectation on agent runs to be
unsuccessful but removes the duplicitous expectation on output matching
/Blah/.

```
```

### DIFF
--- a/integration/tests/user_scenario/basic_workflow/single_env_purge_unmanaged_modules.rb
+++ b/integration/tests/user_scenario/basic_workflow/single_env_purge_unmanaged_modules.rb
@@ -1,10 +1,11 @@
 require 'git_utils'
 require 'r10k_utils'
 require 'master_manipulator'
-test_name 'CODEMGMT-73 - C63184 - Single Environment Purge Unmanaged Modules'
+test_name 'CODEMGMT-78 - Puppetfile Purge --puppetfile & --moduledir flag usage'
 
 #Init
 master_certname = on(master, puppet('config', 'print', 'certname')).stdout.rstrip
+moduledir = on(master, puppet('config', 'print', 'environmentpath')).stdout.strip + '/production/modules'
 git_environments_path = '/root/environments'
 last_commit = git_last_commit(master, git_environments_path)
 r10k_fqp = get_r10k_fqp(master)
@@ -13,8 +14,6 @@ r10k_fqp = get_r10k_fqp(master)
 motd_path = '/etc/motd'
 motd_contents = 'Hello!'
 motd_contents_regex = /\A#{motd_contents}\z/
-
-error_message_regex = /Blah/
 
 #File
 puppet_file = <<-PUPPETFILE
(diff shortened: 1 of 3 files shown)


justinstoller:
still attempting to test this locally

mwaggett:
Would it be worth testing against an appropriate error message or do you think it's better to just leave that out? Just want to make sure we're not failing for a different reason that would indicate a bug.

justinstoller:
> Would it be worth testing against an appropriate error message or do you think it's better to just leave that out?

I think it would be an improvement, though probably not necessary. I was hoping to quickly run the acceptance test and then see what the failure is and update the expectation, but that's proven more difficult than I thought. I _think_ the next run will work :crossed_fingers: and will do so then.

justinstoller:
I think this is ready to go

integration/tests/user_scenario/basic_workflow/single_env_purge_unmanaged_modules.rb
```diff
@@ -1,10 +1,11 @@
 require 'git_utils'
 require 'r10k_utils'
 require 'master_manipulator'
-test_name 'CODEMGMT-73 - C63184 - Single Environment Purge Unmanaged Modules'
+test_name 'CODEMGMT-78 - Puppetfile Purge --puppetfile & --moduledir flag usage'
 
 #Init
 master_certname = on(master, puppet('config', 'print', 'certname')).stdout.rstrip
+moduledir = on(master, puppet('config', '--environment', 'production', 'print', 'environmentpath')).stdout.strip + '/production/modules'
```
mwaggett: iirc, you were just calling `config print moduledir` before, right? Why didn't that work?

I'm surprised you need to include the `production/` dir in the path you're appending - I thought the `environmentpath` would already include the `production/` dir?

mwaggett (approved):

mwaggett (approved):

integration/tests/user_scenario/basic_workflow/single_env_purge_unmanaged_modules.rb
justinstoller: Good catch.

Before I was calling moduledir, which isn't a Puppet thing. It's an actually an r10k thing and I was confused, and modulepath could return a ':' separated list of directories to search for modules, so I decided against using that.

I then put in `--environment production print environmentpath` thinking I'd get back something that ended in production. However, this is returning something I thought was called an "environmentspath" but it looks like "environmentspath" isn't a thing. So we have to append the environment name to it. In that case I don't believe the `--environment production` is necessary anymore and can be removed. I'll do that.

Working in file-sync has destroyed my recollection of this part of Puppet's settings, and I was being sloppy because this acceptance angered me so much.

mwaggett: haha all good! I've never had an understanding of Puppet's settings, so no worries 😅 
thanks for explaining!

mwaggett (approved):

PR #1184, opened 2021-07-12
(maint) Raise when calling Puppetfile#load! if file is missing

justinstoller:
As part of a recent refactor the conditional behavior to check if a
Puppetfile existed was moved from `load` to `load!`. However, the
puppetfile actions only call `load!` and depend on its behavior of
raising for parts of their user experience.

This commit moves the check for Puppetfile existence back to `load` so
that it can be skipped by the puppetfile actions and they will properly
error for users when ran in a directory that does not contain a
Puppetfile.

It adds tests for the `load!` method that exercises its lack of
idempotence. It also removes the `freeze` from remaining internal module
loader objects as the method will still fail, but fails with a more user
friendly error (R10K::Error, "duplicate content") rather than
FreezeError.

```
```

### DIFF
--- a/spec/unit/puppetfile_spec.rb
+++ b/spec/unit/puppetfile_spec.rb
@@ -68,34 +68,57 @@ describe R10K::Puppetfile do
   end
 
   describe "loading a Puppetfile" do
-    it "returns the loaded content" do
-      path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'valid-forge-with-version')
-      subject = described_class.new(path, {})
+    context 'using load' do
+      it "returns the loaded content" do
+        path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'valid-forge-with-version')
+        subject = described_class.new(path, {})
 
-      loaded_content = subject.load
-      expect(loaded_content).to be_an_instance_of(Hash)
+        loaded_content = subject.load
+        expect(loaded_content).to be_an_instance_of(Hash)
 
-      has_some_data = loaded_content.values.none?(&:empty?)
-      expect(has_some_data).to be true
+        has_some_data = loaded_content.values.none?(&:empty?)
+        expect(has_some_data).to be true
+      end
+
+      it "is idempotent" do
+        path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'valid-forge-with-version')
+        subject = described_class.new(path, {})
+
+        expect(subject.loader).to receive(:load).and_call_original.once
+
+        loaded_content1 = subject.load
+        expect(subject.loaded?).to be true
+        loaded_content2 = subject.load
+
+        expect(loaded_content2).to eq(loaded_content1)
+      end
+
+      it "returns nil if Puppetfile doesn't exist" do
+        path = '/rando/path/that/wont/exist'
+        subject = described_class.new(path, {})
+        expect(subject.load).to eq nil
+      end
     end
 
-    it "is idempotent" do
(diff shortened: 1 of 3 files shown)


spec/unit/puppetfile_spec.rb
```diff
@@ -68,34 +68,70 @@
...
+        path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'valid-forge-with-version')
+        subject = described_class.new(path, {})
+
+        expect(subject.loader).to receive(:load).and_call_original.twice
+
+        loaded_content1 = subject.load!
+        expect(subject.loaded?).to be true
+        expect {
+          subject.load!
+        }.to raise_error(R10K::Error)
```
mwaggett: Why does this error? Is there a specific error message we should be looking for?

spec/unit/puppetfile_spec.rb
```diff
@@ -68,34 +68,70 @@
...
+      end
+
+      it "raises if Puppetfile doesn't exist" do
```
mwaggett, line 115: Should this test go under `load` instead of `load!`?

justinstoller: No, `load` will not raise, `load!` will.

spec/unit/puppetfile_spec.rb
justinstoller: It raises saying that duplicate modules are not allowed, however I don't know that that error is important so long as it does error with an nice exception. Here I use R10K::Error as a stand in for a "nice" error since that means the error will be caught and wrapped with appropriate logging. I can add a comment to that effect?

justinstoller: added something to that effect

spec/unit/puppetfile_spec.rb
mwaggett: oh.. okay so yeah `load!` will raise, but we shouldn't ever actually hit this case, right? If everything calls `load` rather than calling `load!` directly, we'll hit the case where we log that the Puppetfile is missing or unreadable and never call `load!`. Am I missing something?

spec/unit/puppetfile_spec.rb
mwaggett: hmm okay. I guess I'm not really understanding why it's important to ensure that `load!` is not idempotent at all..

justinstoller: that's fair, was I really just mirroring the existing tests.

spec/unit/puppetfile_spec.rb
justinstoller: No, the puppetfile actions, eg `R10K::Action::Puppetfile::Purge` all call `load!` and it is the ENOENT that was being raised by `puppetfile_content` that cause the action to abort, print a message at the ERROR level, and return 1. Nothing calls `load` within R10K itself, I don't really know why it exists.

mwaggett: 😑 fun.
Should we update those things to call `load` instead and then have `load` fail with a "nice" error that matches the debug message?

justinstoller: No.

I think the refactor was a slight regression in UX and the implementation of that older, better UX was a little wonky. I've returned it to the better UX with a wonkier implementation. My end goal with this work it stop callers from using the Puppetfile and instead use the ModuleLoader impelementation (which behaves like `load!`) and have the callers check their inputs and handle them nicely.

I'd rather put a button on this part of the refactor and move on towards the eventual goal.

spec/unit/puppetfile_spec.rb
justinstoller: removed

spec/unit/puppetfile_spec.rb
mwaggett: okay, sounds good

mwaggett (approved):

Issue #982, opened 2019-09-18
postrun after deploying a module


amateo:
There is a `postrun` option that enables running a script after deploying an environment. But this script is not run after deploying a module (not an environment).
I think it would be useful to be able to run this script after deploying a module.

chrisongthb:
We need this feature, too. Any news here?

packetcafe:
+1  for this feature.. it would be really handy for the postrun to run for all deployment include a single module.

chrisongthb:
This issue is still important to us.

kendall-moore:
+1 for this feature. Deploying a module is still a deployment, so there's no reason postrun shouldn't be triggered.

packetcafe:
+1  I'm also waiting for the feature, it not stable at all!.

Magisus:
Agreed, this should be fixed. Related jira ticket: https://tickets.puppetlabs.com/browse/RK-369

Magisus:
The `postrun` command for environments will read a variable `$modifiedenvs` and [sub in](https://github.com/puppetlabs/r10k/blob/ded1045f29f99a624c19eae25d96217cd0dee179/lib/r10k/action/deploy/environment.rb#L126-L131) the environments being deployed. We probably want to respect this for `deploy module` too, but should it only supply environments that were both requested _and_ actually had the module in question? Or all requested envs, regardless of whether they were actually updated?

alexjfisher:
Preferably, just the environments that were updated?

Magisus:
Cool, that makes sense to me too.

Magisus:
If it turns out that no environments were actually updated (e.g. because none of the requested ones contained the module), do we still want to run the command at all? Or only if it does not include `$modifiedenvs`?

Magisus:
I have a PR up that only runs the command if any envs actually synced modules, and populates `$modifiedenvs` with the changed environments, see https://github.com/puppetlabs/r10k/pull/1190. Still working on tests but would appreciate thoughts on the approach and functionality.

Magisus:
This has been merged, and will be released soon (in the next month), or we could do sooner if there's interest, please let us know. Also if anyone has any issues with it or needs any tweaks to the functionality.

PR #1190, opened 2021-07-23
(RK-369) Run postrun command for module deploys

Magisus:
This commit updates the `deploy module` command to run the configured
`postrun` command that is normally run after an environment deploy. It
will only run the command if any of the requested environments were
actually modified, i.e. if they contained any of the modules we asked to
deploy. Similarly, if the command contains the `$modifiedenvs` variable,
it will be populated only with environments that actually had modules
deployed. This mirrors how we run `puppet generate types` for module
deploys when that setting is enabled.

Still needs tests, they're a little involved, but would appreciate thoughts on the approach and functionality!

Associated issue: https://github.com/puppetlabs/r10k/issues/982

### DIFF
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -29,6 +29,7 @@ module R10K
           super
 
           requested_env = @opts[:environment] ? [@opts[:environment].gsub(/\W/, '_')] : []
+          @modified_envs = []
 
           @settings = @settings.merge({
             overrides: {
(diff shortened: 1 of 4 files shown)


Magisus:
Did some manual testing of this, though, and it seems to work as described.

tvpartytonight:
LGTM, should update the `Definition` description too: https://github.com/puppetlabs/r10k/blob/29f163b3ee257e2590ebf7ca41086ea6ed5ebd0f/lib/r10k/settings.rb#L183

Magisus:
Tests added.

lib/r10k/action/deploy/module.rb
```diff
@@ -69,6 +70,19 @@ def call
...
+            if postcmd.grep('$modifiedenvs').any?
+              envs_to_run = @modified_envs.join(' ')
+              logger.debug "Running postrun command for environments #{envs_to_run}."
+              postcmd = postcmd.map { |e| e.gsub('$modifiedenvs', envs_to_run) }
+            end
+            subproc = R10K::Util::Subprocess.new(postcmd)
+            subproc.logger = logger
+            subproc.execute
+          else
+            logger.debug "No environments were modified, not executing postrun command."
```
justinstoller: This could happen because environments were modified but the postrun setting is nil.

Magisus: Oh good catch, I'll split that out.

Magisus: Updated.

justinstoller (approved):

PR #1192, opened 2021-07-28
(PF-2437) Allow r10k to authenticate to the Forge

binford2k:
In the near future, the Forge will be adding support for authenticated
module downloads. This extends support for token based authentication to
that use case.

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -167,9 +167,8 @@ This defaults to 'https://forgeapi.puppetlabs.com'
 
 #### authorization_token
 
-The 'authorization_token' setting allows you to provide a token for authenticating to a
-custom Forge server. When set, 'baseurl' must also be set.
-You will need to prepend your token with 'Bearer ' if using Artifactory as your Forge server.
+The 'authorization_token' setting allows you to provide a token for authenticating to a Forge server.
+You will need to prepend your token with 'Bearer ' to authenticate to the Forge or when using your own Artifactory server.
 
 ```yaml
 forge:
--- a/spec/unit/action/runner_spec.rb
+++ b/spec/unit/action/runner_spec.rb
@@ -310,14 +310,6 @@ describe R10K::Action::Runner do
         runner.setup_settings
         runner.setup_authorization
       end
-
-      it 'errors if no custom forge URL is set' do
-        options = { config: "spec/fixtures/unit/action/r10k_forge_auth_no_url.yaml" }
-        runner = described_class.new(options, %w[args yes], action_class)
-        expect(PuppetForge::Connection).not_to receive(:authorization=).with('faketoken')
-
-        expect { runner.setup_settings }.to raise_error(R10K::Error, /Cannot specify a Forge auth/)
-      end
     end
 
     context "license auth" do
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -5,6 +5,7 @@ Unreleased
 ----------
 
 - (CODEMGMT-1454) Ensure missing repo caches are re-synced [#1210](https://github.com/puppetlabs/r10k/pull/1210)
+- (PF-2437) Allow token authentication to be used with the Forge. [#1192](https://github.com/puppetlabs/r10k/pull/1192)
 
 3.11.0
 ------
--- a/lib/r10k/initializers.rb
+++ b/lib/r10k/initializers.rb
@@ -66,13 +66,7 @@ module R10K
       def call
         with_setting(:baseurl) { |value| PuppetForge.host = value }
         with_setting(:proxy) { |value| PuppetForge::Connection.proxy = value }
-        with_setting(:authorization_token) { |value|
-          if @settings[:baseurl]
-            PuppetForge::Connection.authorization = value
-          else
-            raise R10K::Error, "Cannot specify a Forge authorization token without configuring a custom baseurl."
-          end
-        }
+        with_setting(:authorization_token) { |value| PuppetForge::Connection.authorization = value }
       end
     end
   end


Magisus (approved):

Magisus:
Oh cool, good timing on us adding this support then.

doc/dynamic-environments/configuration.mkd
```diff
@@ -167,9 +167,8 @@ This defaults to 'https://forgeapi.puppetlabs.com'
 
 #### authorization_token
 
-The 'authorization_token' setting allows you to provide a token for authenticating to a
-custom Forge server. When set, 'baseurl' must also be set.
-You will need to prepend your token with 'Bearer ' if using Artifactory as your Forge server.
+The 'authorization_token' setting allows you to provide a token for authenticating to a Forge server.
+You will need to prepend your token with 'Bearer ' to authenticate to the Forge or when using your own Artifactory server.
```
mwaggett, line 171:
```suggestion
You will need to prepend your token with 'Bearer ' to authenticate to the Forge or when using Artifactory as your Forge server.
```

spec/unit/action/runner_spec.rb
```diff
@@ -230,14 +230,6 @@ def call
         runner.setup_settings
         runner.setup_authorization
       end
-
-      it 'errors if no custom forge URL is set' do
```
mwaggett, line 314: Would it be valuable to test that setting `authorization_token` now _does_ work if `baseurl` isn't set?

mwaggett (commented):
Have you done any manual testing?

Is there Forge work that needs to land before this can be released?

spec/unit/action/runner_spec.rb
Magisus: Considering how mocked up this all is, not really IMO.

Magisus:
Is it even possible to do manual testing right now, or does the forge not yet support token auth?

binford2k:
There is Forge WIP that will need to land before this is actually functional. It shouldn't actually break anything, so it won't _hurt_ anything to release early, but I'd wait until @carabasdaniel gives the go-ahead.

mwaggett:
okay cool. yeah @binford2k I'd like to wait for that to land so that this can be manually tested before merging.

caseywilliams:
This won't be ready for testing until PF-2447 is done - I've updated the jira ticket for this PR to reflect that. I'd prefer to wait until we can test with real data before we merge these changes.

geoffnichols:
@binford2k and @mwaggett - when y'all get a chance to take a look at this again, it looks like PF-2447 was resolved yesterday!

mwaggett:
awesome, @binford2k would you mind rebasing and then doing some manual testing?

doc/dynamic-environments/configuration.mkd
caseywilliams: I suggest waiting to merge this until https://github.com/puppetlabs/forge-ruby/pull/89 is merged and released, since it obviates the need to include 'Bearer'

Magisus: Looks like that should work for Artifactory-based Forge too?

caseywilliams: I think so, although I haven't checked that manually - I did manually verify just now that the version of forge-ruby I just released (3.1.0) will allow this to work with the forge API without prepending `Bearer ` in the r10k.yaml setting. Otherwise, the only thing stopping r10k from authenticating before was the check for a `baseurl` that's removed in this PR.

Unless I'm missing something, it seems like the options here are to:
* Update this README to suggest specifying `Bearer <your-key>` as the `authorization_token` setting for now (regardless of whether you're authenticating to the forge or to artifactory), and leave the forge-ruby dependency as-is
or
* Update the forge-ruby dependency to 3.1.0 and don't mention `Bearer` at all - the  biggest impact from this would probably be that the default forge API url set by forge-ruby would update from https://forgeapi.puppetlabs.com to https://forgeapi.puppet.com.

caseywilliams: I guess a third option would be to add new logic in r10k to prepend `Bearer ` to the key as needed - it wouldn't affect forge-ruby behavior in 3.1.0, and it would let people omit `Bearer ` in their r10k.yaml file if they're using an older forge-ruby version.

Magisus: We talked about that and decided we didn't know enough about what token requirements other custom Forge implementations might have, and didn't want to introduce magic that was going to bite us. Though I guess if the Forge gem is doing that now, that point is kinda moot.

mwaggett: @caseywilliams Would the URL update in 3.1.0 that you mentioned cause breakages? I know we had some trouble updating that URL in r10k, so we're still using the `puppetlabs` one.

caseywilliams: Yeah, it might - if I remember correctly, the PE acceptance tests were breaking last time we tested using the updated URL in r10k. forgeapi.puppet.com is really the current canonical place to find the API, so everyone should switch over eventually, but the two biggest issues in updating would be (as far as I can tell) (1) if a user had allowlisted access in their network to forgeapi.puppetlabs.com and then suddenly lost access when the url updated but their rules had not, and (2) needing to schedule the acceptance test update work

binford2k:
@mwaggett manually tested, works as expected 👍

doc/dynamic-environments/configuration.mkd
mwaggett: looks like bumping the puppet_forge gem to 3.1.0 will involve packaging a bunch of additional dependencies in pe-r10k-vanagon, since faraday 1.3+ has a bunch more dependencies than the version we're currently on.

I'm thinking it might be best to open a separate ticket/issue for bumping puppet_forge, removing the docs around `Bearer `, and updating the Forge API url. Would you mind doing that @caseywilliams?

mwaggett (approved):

doc/dynamic-environments/configuration.mkd
caseywilliams: Sure, wasn't sure exactly which jira project was appropriate but I filed [PE-32770](https://tickets.puppetlabs.com/browse/PE-32770).

PR #1194, opened 2021-07-29
(maint) Remove rspec testing from Travis

Magisus:
We run these tests in GH Actions now, so we don't need to duplicate them
in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,17 +11,7 @@ before_install: gem install bundler -v '< 2' --no-document
 matrix:
   include:
     - stage: r10k tests
-      rvm: 2.7.0
-    - stage: r10k tests
-      rvm: 2.6.5
-    - stage: r10k tests
-      rvm: 2.5.0
-    - stage: r10k tests
-      rvm: 2.4.0
-    - stage: r10k tests
       rvm: 2.3.0
-    - stage: r10k tests
-      rvm: jruby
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,18 +10,6 @@ jdk:
...
-    - stage: r10k tests
-      rvm: 2.7.0
-    - stage: r10k tests
-      rvm: 2.6.5
-    - stage: r10k tests
-      rvm: 2.5.0
-    - stage: r10k tests
-      rvm: 2.4.0
-    - stage: r10k tests
-      rvm: 2.3.0
```
mwaggett, line 22: oh right.. we had talked about ideally dropping ruby 2.3.0, but I think we decided that we needed to do that at a major version boundary. We didn't add ruby 2.3.0 to the Github Actions test matrix: https://github.com/puppetlabs/r10k/blob/main/.github/workflows/rspec_tests.yml#L12 Should we?

Magisus: Yeah probably, I'll see about adding it.

Magisus:
I'm not sure why GHA's Ruby 2.3 fails so horribly, but I'm not sure it's worth investing in to figure it out, so I'm keeping that cell in Travis instead, where it works.

mwaggett (approved):

PR #1196, opened 2021-08-03
CRI: Allow latest versions on newer ruby

bastelfreak:
People with a more modern operating system/ruby version might already
have a newer CRI version. Given that Ruby 2.4 is dead since years I
think it's a good idea to support newer versions.

```
```

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -23,7 +23,8 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored2',   '3.1.2'
-  s.add_dependency 'cri', '2.15.10'
+  s.add_dependency 'cri', '2.15.10' if RUBY_VERSION < '2.5.0'
+  s.add_dependency 'cri', '>= 2.15.10', '< 3' if RUBY_VERSION >= '2.5.0'
 
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'


r10k.gemspec
```diff
@@ -23,7 +23,8 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored2',   '3.1.2'
-  s.add_dependency 'cri', '2.15.10'
+  s.add_dependency 'cri', '2.15.10' if RUBY_VERSION < '2.5.0'
+  s.add_dependency 'cri' if RUBY_VERSION >= '2.5.0'
```
Magisus: I think we want to at least pin the X version here: `~> 2.15`. I know `cri` has always respected semver, but it still seems good to play it a little safe (and match what we're doing for most other gems).

bastelfreak: updated it

Magisus (approved):

r10k.gemspec
alexjfisher: I'm wondering if this is a valid thing to do.  ie does this just set the dependency based on what version of ruby is used when `gem build` is run?

r10k.gemspec
```diff
@@ -23,7 +23,8 @@ Gem::Specification.new do |s|
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored2',   '3.1.2'
-  s.add_dependency 'cri', '2.15.10'
+  s.add_dependency 'cri', '2.15.10' if RUBY_VERSION < '2.5.0'
+  s.add_dependency 'cri', '>= 2.15.10', '< 3' if RUBY_VERSION >= '2.5.0'
```
ekohl, line 27: This doesn't work. A gemspec is evaluated during build time, which means it will now depend on when `gem build` runs what the dependencies end up being.

r10k.gemspec
bastelfreak: I want to add that I thought that's safe to use because I saw it earlier on https://github.com/puppetlabs/forge-ruby/blob/main/puppet_forge.gemspec#L37-L38

alexjfisher: Suggest reverting this PR for now and opening a new discussion if you want to drop ruby 2.4 support.

Magisus: Oh man, this is my bad too, I tested it just with a bundle install and saw that work. Didn't realize the implications for building the gem. Will revert.

Magisus:
PR for revert: https://github.com/puppetlabs/r10k/pull/1197

Thanks for catching this, everyone!

PR #1199, opened 2021-08-09
(maint) Update acceptance tests for missing environments

mwaggett:
This commit updates the expected behavior when running puppet against
environments that don't exist. Previously, puppet runs would fail if the
requested environment wasn't found, but recent changes in puppet (PUP-10539)
changed this behavior to fall back to the environment specified by the server.

### DIFF
--- a/integration/Rakefile
+++ b/integration/Rakefile
@@ -1,4 +1,5 @@
 require 'rototiller'
+require 'fileutils'
 
 namespace :ci do
   namespace :test do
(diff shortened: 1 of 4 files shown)


integration/Rakefile
```diff
@@ -1,10 +1,12 @@
 require 'rototiller'
+require 'fileutils'
 
 namespace :ci do
   namespace :test do
 
     desc 'Tests at the component level for the pe-r10k project'
     task :component => [:check_pe_r10k_env_vars] do
+      FileUtils.mkdir_p 'configs'
```
justinstoller: Should this go in the `beaker_hostgenerator` task?

mwaggett: hmm yeah I guess so

mwaggett: updated!

Magisus (approved):

PR #1200, opened 2021-08-10
(CODEMGMT-1415) Allow assuming modules are unchanged between deploys

justinstoller:
The first 7 commits are follow work from earlier refactors that remove the Puppetfile class from being used anywhere within the main code paths of R10K. You can see them by themselves here: https://github.com/justinstoller/r10k/compare/61c589ec62851ba427fc7c110a3762a032f43ac4...45d485822fd3b789ce1e074bb434b84b75584dbf The commits are all relatively small and self contained refactors.

The last commit implements most of the this feature (and can be reviewed on its own). Comments on approach specifically are much appreciated.

### DIFF
--- a/doc/dynamic-environments/usage.mkd
+++ b/doc/dynamic-environments/usage.mkd
@@ -62,6 +62,32 @@ useful if you want to make sure that a given environment is fully up to date.
 
 - - -
 
+There is also a middle ground between updating all modules and updating no modules.
+It is often desirable to update the environment and then update only those modules
+whose definitions have changed in the Puppetfile, or whose content _could_ have
+changed since the last deployment (eg, Forge modules with their version set to
+`:latest` or Git modules who point to a `branch` ref).
+
+This can be achieved by assuming content is unchanged locally on disk. This is the
+opposite of what one would assume during a module development cycle, when a user
+might be making local edits to test code changes. However, in production, access
+to puppet code is usually locked down, and updates are deployed through automated
+invocations of R10K.
+
+In these cases, deploys where most modules are unchanged and reference exact
+versions (ie, not `:latest` or a branch as mentioned above), this invocation
+may shorten deployment times dozens of seconds if not minutes depending on how
+many modules meet the above criteria (approximately 1 minute for every 400 modules).
+
+To take advantage of this, set as many modules as possible in the Puppetfile to
+explicit, static version. These are released Forge versions, or Git modules using
+the `:tag`, or `:commit` keys. Git `:ref`s containing only the full 40 character
+commit SHA will also be treated as static versions. Then invoke a deploy with:
+
+    r10k deploy environment production --modules --assume-unchanged
+
+- - -
+
 Update a single environment and specify a default branch override:
 
     r10k deploy environment my_working_environment --modules --default-branch-override default_branch_override
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -41,7 +41,8 @@ module R10K
                 requested_environments: @argv.map { |arg| arg.gsub(/\W/,'_') },
                 default_branch_override: @default_branch_override,
                 generate_types: @generate_types || settings.dig(:deploy, :generate_types) || false,
-                preload_environments: true
+                preload_environments: true,
+                assume_unchanged: @assume_unchanged
               },
               modules: {
                 exclude_spec: settings.dig(:deploy, :exclude_spec),
@@ -238,6 +239,7 @@ module R10K
           super.merge(puppetfile: :modules,
                       modules: :self,
                       cachedir: :self,
+                      'assume-unchanged': :self,
                       'no-force': :self,
                       'exclude-spec': :self,
                       'generate-types': :self,
--- a/lib/r10k/action/deploy/environment.rb
+++ b/lib/r10k/action/deploy/environment.rb
@@ -47,6 +47,7 @@ module R10K
                 exclude_spec: settings.dig(:deploy, :exclude_spec),
                 requested_modules: [],
                 deploy_modules: @modules,
+                pool_size: @settings[:pool_size] || 4,
                 force: !@no_force, # force here is used to make it easier to reason about
               },
               purging: {
--- a/lib/r10k/action/puppetfile/check.rb
+++ b/lib/r10k/action/puppetfile/check.rb
@@ -14,7 +14,7 @@ module R10K
 
           loader = R10K::ModuleLoader::Puppetfile.new(**options)
           begin
-            loader.load
+            loader.load!
             $stderr.puts _("Syntax OK")
             true
           rescue => e
--- a/lib/r10k/action/puppetfile/check.rb
+++ b/lib/r10k/action/puppetfile/check.rb
@@ -1,6 +1,6 @@
-require 'r10k/puppetfile'
 require 'r10k/action/base'
 require 'r10k/errors/formatting'
+require 'r10k/module_loader/puppetfile'
 
 module R10K
   module Action
@@ -8,11 +8,13 @@ module R10K
       class Check < R10K::Action::Base
 
         def call
-          pf = R10K::Puppetfile.new(@root,
-                                    {moduledir: @moduledir,
-                                     puppetfile_path: @puppetfile})
+          options = { basedir: @root }
+          options[:moduledir] = @moduledir if @moduledir
+          options[:puppetfile] = @puppetfile if @puppetfile
+
+          loader = R10K::ModuleLoader::Puppetfile.new(**options)
           begin
-            pf.load!
+            loader.load
             $stderr.puts _("Syntax OK")
             true
           rescue => e
--- a/lib/r10k/action/puppetfile/purge.rb
+++ b/lib/r10k/action/puppetfile/purge.rb
@@ -1,7 +1,7 @@
-require 'r10k/puppetfile'
-require 'r10k/util/cleaner'
 require 'r10k/action/base'
 require 'r10k/errors/formatting'
+require 'r10k/module_loader/puppetfile'
+require 'r10k/util/cleaner'
 
 module R10K
   module Action
@@ -9,13 +9,16 @@ module R10K
       class Purge < R10K::Action::Base
 
         def call
-          pf = R10K::Puppetfile.new(@root,
-                                    {moduledir: @moduledir,
-                                     puppetfile_path: @puppetfile})
-          pf.load!
-          R10K::Util::Cleaner.new(pf.managed_directories,
-                                  pf.desired_contents,
-                                  pf.purge_exclusions).purge!
+          options = { basedir: @root }
+
+          options[:moduledir]  = @moduledir  if @moduledir
+          options[:puppetfile] = @puppetfile if @puppetfile
+
+          loader = R10K::ModuleLoader::Puppetfile.new(**options)
+          loaded_content = loader.load
+          R10K::Util::Cleaner.new(loaded_content[:managed_directories],
+                                  loaded_content[:desired_contents],
+                                  loaded_content[:purge_exclusions]).purge!
 
           true
         rescue => e
--- a/lib/r10k/action/puppetfile/purge.rb
+++ b/lib/r10k/action/puppetfile/purge.rb
@@ -15,7 +15,7 @@ module R10K
           options[:puppetfile] = @puppetfile if @puppetfile
 
           loader = R10K::ModuleLoader::Puppetfile.new(**options)
-          loaded_content = loader.load
+          loaded_content = loader.load!
           R10K::Util::Cleaner.new(loaded_content[:managed_directories],
                                   loaded_content[:desired_contents],
                                   loaded_content[:purge_exclusions]).purge!
--- a/lib/r10k/environment/base.rb
+++ b/lib/r10k/environment/base.rb
@@ -1,5 +1,8 @@
-require 'r10k/util/subprocess'
+require 'r10k/content_synchronizer'
 require 'r10k/logging'
+require 'r10k/module_loader/puppetfile'
+require 'r10k/util/cleaner'
+require 'r10k/util/subprocess'
 
 # This class defines a common interface for environment implementations.
 #
@@ -64,7 +67,12 @@ class R10K::Environment::Base
 
     loader_options = { basedir: @full_path, overrides: @overrides, environment: self }
     loader_options[:puppetfile] = @puppetfile_name if @puppetfile_name
+
     @loader = R10K::ModuleLoader::Puppetfile.new(**loader_options)
+
+    if @overrides.dig(:environments, :assume_unchanged)
+      @loader.load_metadata
+    end
 
     @base_modules = nil
     @managed_directories = [ @full_path ]
@@ -120,8 +128,7 @@ class R10K::Environment::Base
   #   associated with this environment.
   def modules
     if @base_modules.nil?
-      loaded_content = @loader.load
-      @base_modules = loaded_content[:modules]
+      load_puppetfile_modules
     end
 
     @base_modules
@@ -141,23 +148,31 @@ class R10K::Environment::Base
   end
 
   def deploy
-    loaded_content = @loader.load
-    @base_modules = loaded_content[:modules]
+    if @base_modules.nil?
+      load_puppetfile_modules
+    end
 
     if ! @base_modules.empty?
       pool_size = @overrides.dig(:modules, :pool_size)
-      R10K::ContentSynchronizer.concurrent_sync(loaded_content[:modules], pool_size, logger)
+      R10K::ContentSynchronizer.concurrent_sync(@base_modules, pool_size, logger)
     end
+
+    if (@overrides.dig(:purging, :purge_levels) || []).include?(:puppetfile)
+      logger.debug("Purging unmanaged Puppetfile content for environment '#{dirname}'...")
+      @puppetfile_cleaner.purge!
+    end
+  end
+
+  def load_puppetfile_modules
+    loaded_content = @loader.load
+    @base_modules = loaded_content[:modules]
 
     @purge_exclusions = determine_purge_exclusions(loaded_content[:managed_directories],
                                                    loaded_content[:desired_contents])
 
-    if (@overrides.dig(:purging, :purge_levels) || []).include?(:puppetfile)
-      logger.debug("Purging unmanaged Puppetfile content for environment '#{dirname}'...")
-      R10K::Util::Cleaner.new(loaded_content[:managed_directories],
-                              loaded_content[:desired_contents],
-                              loaded_content[:purge_exclusions]).purge!
-    end
+    @puppetfile_cleaner = R10K::Util::Cleaner.new(loaded_content[:managed_directories],
+                                                  loaded_content[:desired_contents],
+                                                  loaded_content[:purge_exclusions])
   end
 
   def whitelist(user_whitelist=[])
--- a/lib/r10k/environment/base.rb
+++ b/lib/r10k/environment/base.rb
@@ -66,6 +66,7 @@ class R10K::Environment::Base
     loader_options[:puppetfile] = @puppetfile_name if @puppetfile_name
     @loader = R10K::ModuleLoader::Puppetfile.new(**loader_options)
 
+    @base_modules = nil
     @managed_directories = [ @full_path ]
     @desired_contents = []
     @purge_exclusions = []
@@ -118,8 +119,12 @@ class R10K::Environment::Base
   # @return [Array<R10K::Module::Base>] All modules defined in the Puppetfile
   #   associated with this environment.
   def modules
-    @puppetfile.load
-    @puppetfile.modules
+    if @base_modules.nil?
+      loaded_content = @loader.load
+      @base_modules = loaded_content[:modules]
+    end
+
+    @base_modules
   end
 
   # @return [Array<R10K::Module::Base>] Whether or not the given module
@@ -137,9 +142,9 @@ class R10K::Environment::Base
 
   def deploy
     loaded_content = @loader.load
-    @modules = loaded_content[:modules]
+    @base_modules = loaded_content[:modules]
 
-    if ! @modules.empty?
+    if ! @base_modules.empty?
       pool_size = @overrides.dig(:modules, :pool_size)
       R10K::ContentSynchronizer.concurrent_sync(loaded_content[:modules], pool_size, logger)
     end
--- a/lib/r10k/environment/base.rb
+++ b/lib/r10k/environment/base.rb
@@ -34,6 +34,10 @@ class R10K::Environment::Base
   #   @return [String] The puppetfile name (relative)
   attr_reader :puppetfile_name
 
+  attr_reader :managed_directories, :purge_exclusions, :desired_contents
+
+  attr_reader :loader
+
   # Initialize the given environment.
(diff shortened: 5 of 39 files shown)


spec/unit/action/puppetfile/purge_spec.rb
```diff
@@ -37,23 +39,23 @@ def purger(opts = {}, argv = [], settings = {})
   end
 
   describe "using custom paths" do
-    before(:each) do
-      allow(puppetfile).to receive(:purge!)
-    end
+    #before(:each) do
+    #  allow(puppetfile).to receive(:purge!)
+    #end
```
Magisus: Can this be deleted?

lib/r10k/action/puppetfile/purge.rb
```diff
@@ -1,5 +1,5 @@
-require 'r10k/puppetfile'
 require 'r10k/util/cleaner'
+require 'r10k/content_synchronizer'
```
Magisus: Is this actually needed? I don't see it used...

spec/unit/action/puppetfile/install_spec.rb
```diff
@@ -51,14 +58,16 @@ def installer(opts = {}, argv = [], settings = {})
 
   describe "purging" do
     before do
-      allow(puppetfile).to receive(:modules).and_return([])
+      allow(loader).to receive(:load).and_return({
+        modules:             [],
+        desired_contents:    [ 'root/foo' ],
+        managed_directories: [ 'root' ],
+        purge_exclusions:    [ 'root/**/**.rb' ]
+      })
```
Magisus, line 66: I don't really mind, but is there a reason this (and the test below) have these in `before` blocks, even though there's only one test case?

lib/r10k/action/puppetfile/check.rb
```diff
@@ -1,4 +1,4 @@
-require 'r10k/puppetfile'
+require 'r10k/module_loader/puppetfile'
```
Magisus: I notice you add this require here, but not in the actions updated by the previous two commits (purge and install). Are we getting lucky there, and something else has already required it in those code paths, but not this one? Is that okay, or should we add this require to those too, to be explicit?

lib/r10k/action/deploy/environment.rb
```diff
@@ -47,6 +47,7 @@ def initialize(opts, argv, settings = {})
                 deploy_spec: settings.dig(:deploy, :deploy_spec),
                 requested_modules: [],
                 deploy_modules: @modules,
+                pool_size: @settings[:pool_size] || 4,
```
Magisus, line 51: How does this interact with the [default](https://github.com/puppetlabs/r10k/blob/3a8c6e80e69cad6aa1198241f3c7b49aeaf4f07c/lib/r10k/settings.rb#L235) defined in the r10k settings object? The comment above this section implies that this should fall back to that, so I would kind of hope we wouldn't need to specify a default separately here. Or are there scenarios where we don't have access to those (maybe tests...?)?

lib/r10k/module_loader/stub_puppetfile.rb
```diff
@@ -0,0 +1,28 @@
+require 'r10k/module_loader/puppetfile'
+require 'r10k/module/stub'
+
+module R10K
+  module ModuleLoader
+
+    # This sub classes the Puppetfile and provides the same functionality
+    # except non of the modules instantiated are sync-able. They are only
```
Magisus: "non" -> "none"

Magisus:
Looks like this needs a rebase, too.

spec/unit/action/puppetfile/purge_spec.rb
justinstoller: yep!

lib/r10k/action/puppetfile/purge.rb
justinstoller: I mixed up a few requires when I initially did this. I went back and updated the requires for the puppetfile actions today.

spec/unit/action/puppetfile/install_spec.rb
justinstoller: No, it's a weird pattern this whole test file used. I just made it more obvious by making the single line stub be six lines. I've updated the whole file to pull them into the `it` blocks when possible.

lib/r10k/action/puppetfile/check.rb
justinstoller: I added them to the missing actions.

lib/r10k/module_loader/stub_puppetfile.rb
justinstoller: :+1:

lib/r10k/action/deploy/environment.rb
justinstoller: Yeah the `settings` parameter to this object is set when invoked by our runner and populated with a settings definition parsed from the r10k.yaml. However, tests and other users of this class directly (notably Bolt) pass hashes (or nothing).

Magisus (commented):
I think what you have here so far looks good!

Seems like you might still be working on tests related to the new functionality?

lib/r10k/environment/base.rb
```diff
@@ -48,6 +54,7 @@ def initialize(name, basedir, dirname, options = {})
     @options = options
     @puppetfile_name = options.delete(:puppetfile_name)
     @overrides = options.delete(:overrides) || {}
+    @modules = []
```
justinstoller: This can be removed now

lib/r10k/environment/base.rb
```diff
@@ -57,6 +64,22 @@ def initialize(name, basedir, dirname, options = {})
...
+    loader_options = { basedir: @full_path, overrides: @overrides, environment: self }
+    loader_options[:puppetfile] = @puppetfile_name if @puppetfile_name
+
+    if @overrides.dig(:environments, :assume_unchanged)
+      @pre_loader = R10K::ModuleLoader::StubPuppetfile.new(**loader_options)
+      existing_modules = @pre_loader.load[:modules].map {|mod| [ mod.name, mod.version ] }.to_h
+      loader_options[:existing_modules] = existing_modules
+    end
+
+    @loader = R10K::ModuleLoader::Puppetfile.new(**loader_options)
```
justinstoller: With this we should comment above where we create the `@puppetfile` that it is deprecated and below that `accept` is deprecated.

lib/r10k/environment/base.rb
```diff
@@ -57,6 +64,22 @@ def initialize(name, basedir, dirname, options = {})
                                          force: @overrides.dig(:modules, :force),
                                          puppetfile_name: @puppetfile_name})
     @puppetfile.environment = self
+
+    loader_options = { basedir: @full_path, overrides: @overrides, environment: self }
+    loader_options[:puppetfile] = @puppetfile_name if @puppetfile_name
+
+    if @overrides.dig(:environments, :assume_unchanged)
+      @pre_loader = R10K::ModuleLoader::StubPuppetfile.new(**loader_options)
+      existing_modules = @pre_loader.load[:modules].map {|mod| [ mod.name, mod.version ] }.to_h
```
justinstoller: this is pretty dense line compared to most of this code base. the formatting is also my own odd style.

lib/r10k/environment/base.rb
```diff
@@ -124,28 +151,37 @@ def accept(visitor)
   end
 
   def deploy
-    puppetfile.load(@overrides.dig(:environments, :default_branch_override))
+    loaded_content = @loader.load
```
justinstoller: It's weird to have `modules` call `@loader.load` as well as `deploy`. Can we factor that into a shared method?

lib/r10k/environment/base.rb
```diff
@@ -106,8 +129,12 @@ def info
   # @return [Array<R10K::Module::Base>] All modules defined in the Puppetfile
   #   associated with this environment.
   def modules
```
justinstoller, line 129: note: this method is overidden by `Environment::WithModules` which Git environments inherit

justinstoller: Could this method be called before an environment.sync?

lib/r10k/environment/base.rb
```diff
@@ -124,28 +151,37 @@ def accept(visitor)
   end
 
   def deploy
-    puppetfile.load(@overrides.dig(:environments, :default_branch_override))
+    loaded_content = @loader.load
+    @base_modules = loaded_content[:modules]
 
-    puppetfile.sync
```
justinstoller, line 129: It might make this PR more understandable to do the content sync in one commit (by getting the modules from the puppetfile object. Then adding loader in another commit.

lib/r10k/module/stub.rb
```diff
@@ -0,0 +1,103 @@
+require 'r10k/module'
+
+class R10K::Module::Stub < R10K::Module::Base
```
justinstoller: Stub feels like a bad name for these, but not sure what else would work better?
I thought about making this these changes to the Base module as this kinda represents an unsync-able, abstract module only as it is represented in a Puppetfile. However this module has a grab bag of knowledge about other module implimentations, which isn't great in its own right and feels especially bad in a `Base` module. But maybe it would be worth having implementations register version parsing when they register themselves?

lib/r10k/module_loader/stub_puppetfile.rb
```diff
@@ -0,0 +1,36 @@
+require 'r10k/module_loader/puppetfile'
+require 'r10k/module/stub'
+
+module R10K
+  module ModuleLoader
+
+    # This sub classes the Puppetfile and provides the same functionality
+    # except none of the modules instantiated are sync-able. They are only
+    # used to compare versions to decide if the real modules should be synced.
+    class StubPuppetfile < R10K::ModuleLoader::Puppetfile
```
justinstoller: Also don't like Stub here. And since we don't need any of the purge info maybe it'd be better to factor out the parsing functionality and include it both here and in the normal puppetfile module loader rather than subclassing?

lib/r10k/environment/base.rb
Magisus: I don't mind it 🤷‍♀️

lib/r10k/module/stub.rb
Magisus: I agree that `Stub` isn't quite right. Makes this almost sound like a testing implementation, rather than something we actually use.

I see what you mean about the issues with putting it into the base module class. Are you thinking that if you had module implementations "register" a version parsing schema, we could do away with a separate class altogether (and therefore don't need to think of a name)?

If we want to keep this class... we could call it something like `Metadata`? To imply that it's just the external properties of the module and not an actual module?

lib/r10k/module_loader/stub_puppetfile.rb
Magisus: I think I kind of like that idea, if it's not too difficult. Seems a little weird to have a subclass that is supposed to do _less_ than the parent.

lib/r10k/module/definition.rb
```diff
@@ -0,0 +1,56 @@
...
+      # This logic is clear and included for completeness sake, though at
+      # this time Stub modules do not support SVN versions.
+      type = 'svn'
+    else
+      type = 'forge'
+    end
+
+    {
+      expected: version,
+      actual: 'assumed',
```
Magisus: "assumed" here refers to the fact that we're assuming the actual version is also the expected version? From what I can tell, this is just used to write info to the `.r10k-deploy.json`, is that correct? If so, this seems fine, just wanted to understand how this arbitrary string is used 🙂

lib/r10k/module/definition.rb
```diff
@@ -40,7 +40,7 @@ def properties
       type = 'git'
     elsif @args[:svn]
       # This logic is clear and included for completeness sake, though at
-      # this time Stub modules do not support SVN versions.
+      # this time modules definition do not support SVN versions.
```
Magisus: "module definitions"

lib/r10k/environment/with_modules.rb
```diff
@@ -126,13 +125,6 @@ def add_module(name, args)
 
   include R10K::Util::Purgeable
 
-  # Returns an array of the full paths that can be purged.
-  # @note This implements a required method for the Purgeable mixin
-  # @return [Array<String>]
-  def managed_directories
-    [@full_path]
-  end
-
```
Magisus, line 135: Note to self, this went away because the base environment now defines `managed_directories` as `[@full_path]` by default.

lib/r10k/module.rb
```diff
@@ -22,17 +22,35 @@ def self.register(klass)
...
     end
   end
 
+
   require 'r10k/module/base'
   require 'r10k/module/git'
   require 'r10k/module/svn'
   require 'r10k/module/local'
   require 'r10k/module/forge'
+  require 'r10k/module/definition'
```
Magisus, line 55: I find it a little odd that we require this one at the bottom of the file, even though we use it explicitly up above. But I guess it still works so 🤷‍♀️

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -74,6 +82,30 @@ def load
...
+
+      def load_metadata!
+        dsl = R10K::ModuleLoader::Puppetfile::DSL.new(self, metadata_only: true)
+        dsl.instance_eval(puppetfile_content(@puppetfile), @puppetfile)
+
+        @existing_module_versions_by_name = @existing_module_metadata.map {|mod| [ mod.name, mod.version ] }.to_h
+        empty_load_output.merge(modules: @existing_module_metadata)
+
+      rescue SyntaxError, LoadError, ArgumentError, NameError => e
+        logger.warn "Unable to preload Puppetfile"
```
Magisus: The other version of this includes a little more info about why things failed. I know we don't want to rethrow here, but do you think it'd be valuable to include at least the message from the original exception?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -134,11 +148,60 @@ def add_module(name, module_info)
...
+        if File.readable?(puppetfile_path)
+          block.call
+        else
+          logger.debug _("Puppetfile %{path} missing or unreadable") % {path: puppetfile_path.inspect}
+
+          empty_load_output
+        end
+      end
+
+      def manage_inputs(name, info)
```
Magisus: I don't love this method name, it's unclear to me what is being managed, per se. Maybe something like "configure_deploy_options" or something? Or is there another way you were thinking about it?

Magisus (commented):
I agree with this refactor, the names are much clearer than before, and I feel like the new stuff fits pretty well into the overall patterns here. The key pieces (like the bit where you conditionally turn a module definition into a real module) feel unobtrusive and elegant to me.

lib/r10k/module/definition.rb
justinstoller: Yes, I can document that.

FWIW, we don't use this info directly in code-manager, though we serve it via an endpoint `deploys/data`. We don't document that endpoint publicly, but CD4PE does use it as part of IA. This change should match the same "type", ie this field for modules in general is an arbitrary string. However, I'm confirming with CD4PE that this won't have negative affects on IA, because to get the "actual" value would require resolving the ref, which will do a fetch, which is the thing we're trying to avoid.

lib/r10k/module.rb
justinstoller: This is dumb, but follows the existing pattern. I wrote and deleted a small novel, I'm happy to talk about what I think is going on in chat.

lib/r10k/module_loader/puppetfile.rb
justinstoller: Updated

lib/r10k/module_loader/puppetfile.rb
justinstoller: Yeah, I couldn't come up with a good name for it because all the ideas I had were too specific (ie we're not coercing or validating inputs) and so I just gave it a very generic name. I've given it another go based on feedback, let me know what you think.

justinstoller:
Sorry, I adopted these suggestions locally almost without exception, but failed to respond. I believe I've addressed them, as well as added needed tests.

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -96,11 +96,11 @@ def load_metadata!
         empty_load_output.merge(modules: @existing_module_metadata)
 
       rescue SyntaxError, LoadError, ArgumentError, NameError => e
-        logger.warn "Unable to preload Puppetfile"
+        logger.warn _("Unable to preload Puppetfile because of %{msg}" % { msg: e.message })
       end
 
       def add_module_metadata(name, info)
-        install_path, metadata_info, _ = manage_inputs(name, info)
+        install_path, metadata_info, _ = parse_module_definition(name, info)
```
Magisus, line 103: I like this, I was going to suggest `parse` but wasn't sure if this counted as parsing. But works for me 👍

Magisus (approved):

Magisus:
One thing I just thought of/realized, is this deliberately a CLI flag only? Or do we want a config-file setting too?

lib/r10k/module/definition.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    mod.origin = origin
+    mod.spec_deletable = spec_deletable
+
+    mod
+  end
+
+  # syncing is a noop for module definitions
+  def sync(args = {})
+    logger.debug1(_("Not updating module %{name}, assuming content unchanged") % {name: name})
```
Magisus, line 27: This is at a lower log level than I expected. Did you have some reasoning for putting it at `debug1`?

Right now at `debug` it looks like this:
```
[2021-08-25 15:45:39 - INFO] Environment prod2 is now at 627332f215d718754ad47ab5c9e924293a3d537d
[2021-08-25 15:45:39 - DEBUG] Updating modules with 4 threads
[2021-08-25 15:45:39 - DEBUG] Module thread 2780 exiting: queue empty
[2021-08-25 15:45:39 - DEBUG] Module thread 2820 exiting: queue empty
[2021-08-25 15:45:39 - DEBUG] Module thread 2800 exiting: queue empty
[2021-08-25 15:45:39 - DEBUG] Module thread 2840 exiting: queue empty
```
While at `debug1` you get:
```
[2021-08-25 15:43:45 - INFO] Environment prod2 is now at 627332f215d718754ad47ab5c9e924293a3d537d
[2021-08-25 15:43:45 - DEBUG] Updating modules with 4 threads
[2021-08-25 15:43:45 - DEBUG1] Not updating module stdlib, assuming content unchanged
[2021-08-25 15:43:45 - DEBUG1] Not updating module apache, assuming content unchanged
[2021-08-25 15:43:45 - DEBUG1] Not updating module zookeeper, assuming content unchanged
[2021-08-25 15:43:45 - DEBUG1] Not updating module concat, assuming content unchanged
[2021-08-25 15:43:45 - DEBUG1] Not updating module inifile, assuming content unchanged
[2021-08-25 15:43:45 - DEBUG1] Not updating module motd, assuming content unchanged
[2021-08-25 15:43:45 - DEBUG] Module thread 2780 exiting: queue empty
[2021-08-25 15:43:45 - DEBUG] Module thread 2800 exiting: queue empty
[2021-08-25 15:43:45 - DEBUG] Module thread 2820 exiting: queue empty
[2021-08-25 15:43:45 - DEBUG] Module thread 2840 exiting: queue empty
```

I guess I have a slight preference for moving it to normal debug, so that when I supply the flag, I don't have to dig as deep to see what it's skipping. But I also don't have a sense if there is already a pattern in this library about what goes at which level.

Magisus: There's honestly not that much difference in output generally between `debug` and `debug1`, things don't get crazy till `debug2`. So maybe it's not a big deal.

justinstoller: I don't have a great reason.

I looked at some other logging that I felt was similar - the skipping module or environment syncing because of we're given a specific list of either - which are also at `debug1` and copied them. However, there's a lot of things around purging and syncing that I would put at similar levels but that are all over the place as far as `debug` vs `debug1` vs `debug2`.

Magisus:
Other than those couple of questions, I'm satisfied with this from my manual testing 👍

justinstoller:
I hadn't thought of it as a config file option as it feels _very_ specific to the behavior of `deploy environment` action, but I'm open to adding that if folks think we should do so.

Magisus:
I see what you mean, and that's fine with me. Would you mind adding a little blurb to https://github.com/puppetlabs/r10k/blob/main/doc/dynamic-environments/usage.mkd#deploying-environments about it though?

lib/r10k/module/definition.rb
Magisus: Okay, I think it's fine where it is then, if it's like other things at that level.

doc/dynamic-environments/usage.mkd
```diff
@@ -62,6 +62,22 @@ useful if you want to make sure that a given environment is fully up to date.
...
 - - -
 
+There is also a middle ground between updating all modules and updating no modules.
+It is often desirable to update the environment and then update only those modules
+whose definitions have changed in the Puppetfile, or whose content _could_ have
+changed since the last deployment (eg, they are Forge modules with their version
+set to `:latest` or Git modules who point to a `branch` ref).
+
+This can be achieved by assuming content is unchanged locally on disk. This is the
+opposite of what one would assume during a module development cycle. However, in
```
Magisus: "...development cycle, when a user might be making local edits to test code changes."

justinstoller:
Added some documentation, though I'd love some feedback on it.

doc/dynamic-environments/usage.mkd
```diff
@@ -62,6 +62,22 @@ useful if you want to make sure that a given environment is fully up to date.
...
+production Puppet compilers access is often locked down and code deployment only
+happens through automated invocations of R10K. In those cases the following
+invocation is ideal:
+
+    r10k deploy environment production --modules --assume-unchanged
+
```
Magisus, line 88: Can we add a note here that with this option, deploys are much more efficient/faster, because we skip a lot of unnecessary checking? Otherwise I think it's not obvious why you would use this.

doc/dynamic-environments/usage.mkd
```diff
@@ -62,6 +62,22 @@ useful if you want to make sure that a given environment is fully up to date.
...
 
+There is also a middle ground between updating all modules and updating no modules.
+It is often desirable to update the environment and then update only those modules
+whose definitions have changed in the Puppetfile, or whose content _could_ have
+changed since the last deployment (eg, they are Forge modules with their version
+set to `:latest` or Git modules who point to a `branch` ref).
+
+This can be achieved by assuming content is unchanged locally on disk. This is the
+opposite of what one would assume during a module development cycle. However, in
+production Puppet compilers access is often locked down and code deployment only
```
Magisus: "... in production, access to puppet code is usually locked down, and updates are deployed through automated..."

doc/dynamic-environments/usage.mkd
justinstoller: Okay, maybe that's better?

Magisus (approved):

justinstoller:
squashed and rebased

Magisus:
Looks like that last commit message still uses the "Stub" terminology a lot, as well as mentioning the StubPuppetfile, which we don't have anymore. Should probably update that to the new "module definition" wording.

justinstoller:
Oh man, thank you.

Magisus (approved):
Hell yeah

PR #1202, opened 2021-08-11
(maint) Restore the ability to load a Puppetfile from a relative `basedir`

mwaggett:
This commit updates the `R10K::Puppetfile` class to pass just the Puppetfile
name to the module loader, rather than passing the whole path. Prior to this
change, if you used a relative `basedir`, we'd pass a relative Puppetfile path
to `R10K::ModuleLoader::Puppetfile`, which would then try to resolve the path
against the `basedir` again. This resulted in attempting to load the Puppetfile
from a path like `my/cool/basedir/my/cool/basedir/Puppetfile`, which doesn't
exist, so loading would fail.


mwaggett:
confirmed that the test I added fails prior to this change.

justinstoller (approved):

PR #1203, opened 2021-08-12
(maint) Pass the appropriate Puppetfile path to module loader

mwaggett:

### DIFF
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -11,7 +11,7 @@ module R10K
       DEFAULT_FORGE_API = 'forgeapi.puppetlabs.com'
 
       attr_accessor :default_branch_override, :environment
-      attr_reader :modules, :moduledir,
+      attr_reader :modules, :moduledir, :puppetfile_path,
         :managed_directories, :desired_contents, :purge_exclusions
 
       # @param basedir [String] The path that contains the moduledir &
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -34,7 +34,7 @@ module R10K
 
         @basedir     = cleanpath(basedir)
         @moduledir   = resolve_path(@basedir, moduledir)
-        @puppetfile  = resolve_path(@basedir, puppetfile)
+        @puppetfile_path  = resolve_path(@basedir, puppetfile)
         @forge       = forge
         @overrides   = overrides
         @environment = environment
@@ -46,13 +46,13 @@ module R10K
         @desired_contents = []
         @purge_exclusions = []
 
-        logger.info _("Using Puppetfile '%{puppetfile}'") % {puppetfile: @puppetfile}
+        logger.info _("Using Puppetfile '%{puppetfile}'") % {puppetfile: @puppetfile_path}
         logger.debug _("Using moduledir '%{moduledir}'") % {moduledir: @moduledir}
       end
 
       def load
         dsl = R10K::ModuleLoader::Puppetfile::DSL.new(self)
-        dsl.instance_eval(puppetfile_content(@puppetfile), @puppetfile)
+        dsl.instance_eval(puppetfile_content(@puppetfile_path), @puppetfile_path)
 
         validate_no_duplicate_names(@modules)
         @modules
@@ -71,7 +71,7 @@ module R10K
         }
 
       rescue SyntaxError, LoadError, ArgumentError, NameError => e
-        raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile})
+        raise R10K::Error.wrap(e, _("Failed to evaluate %{path}") % {path: @puppetfile_path})
       end
 
 
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -1,6 +1,10 @@
+require 'r10k/logging'
+
 module R10K
   module ModuleLoader
     class Puppetfile
+
+      include R10K::Logging
 
       DEFAULT_MODULEDIR = 'modules'
       DEFAULT_PUPPETFILE_NAME = 'Puppetfile'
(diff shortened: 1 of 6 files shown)


lib/r10k/module_loader/puppetfile.rb
```diff
@@ -41,11 +45,14 @@ def initialize(basedir:,
         @managed_directories = []
         @desired_contents = []
         @purge_exclusions = []
+
+        logger.info _("Using Puppetfile '%{puppetfile}'") % {puppetfile: @puppetfile_path}
+        logger.info _("Using moduledir '%{moduledir}'") % {moduledir: @moduledir}
```
justinstoller: I think this should be at debug.

Magisus: I was thinking that too.

mwaggett: Both or just the `moduledir` one? Either way is fine with me, just that the `puppetfile` one was at `info` before, which is why I put them both at that level here, but I don't disagree that debug makes sense.

lib/r10k/puppetfile.rb
```diff
@@ -104,10 +104,10 @@ def load(default_branch_override = nil)
     if self.loaded?
       return @loaded_content
     else
-      if !File.readable?(@puppetfile_path)
```
justinstoller, line 109: Previous to this a missing Puppetfile would cause this portion the deployment to be a noop, while a syntax error in your Puppetfile would still cause the run to fail, right?

It seems we may want to either fix the `@puppetfile_path` to be the same kind of path that the loader will use. Or catch SystemCallError (or whatever else a missing/unreadable file might throw) here?

lib/r10k/module_loader/puppetfile.rb
justinstoller: I'd just do the moduledir one, I think they both feel like debug statements but since the puppetfile one was info we should probably unfortunately keep it there.

lib/r10k/puppetfile.rb
mwaggett: yes.

I have another PR I'm gonna put up that will change the path used here, but it seemed better to me to do the validation/rescuing in the method that would actually throw, especially since the `ModuleLoader` methods are the ones we actually plan to keep. Are you just thinking that, by putting it here, we'd avoid unnecessarily trying to load altogether? Or is there another reason you think that stuff should go here?

justinstoller: I think previously the run would fail if their were syntax errors in the Puppetfile, while now those errors will be printed at debug and the run will pass. (I might be wrong about that, but if that's the case we shouldn't do it).

Good point about the exception handling, I was just spitballing ideas that would keep us from doing a blanket rescue here.

mwaggett: yeah, you're right. I guess I wasn't really clear on _why_ we just log and move on if the Puppetfile is unreadable but throw if there's another kind of error - I assumed that we didn't want to fail the run if there was anything wrong with the Puppetfile in whatever case(s) `load` is used (as opposed to `load!`), but didn't investigate at all. I can bring back the old behavior if you think that's better. I guess a change would probably be considered breaking by _someone_ huh..

justinstoller: So I don't think you need a Puppetfile in an environment if you either don't have dependencies or you vendor them all. In that case the load becomes a no-op and the run continues. However, the attempting to deploy external dependencies is an opt-in feature so maybe it should be an error if you opt-in to deploying modules but there is not Puppetfile? I suppose you could do a deploy all kinda thing and want it to deploy modules if the Puppetfile exists in an environment?

lib/r10k/module_loader/puppetfile.rb
mwaggett: updated

lib/r10k/puppetfile.rb
mwaggett: I removed this change for now.

mwaggett:
This should be good to go now.

mwaggett:
This passed for me locally.

mwaggett:
well.. some iteration of it lol 👀

mwaggett:
okay that should be better 😅

Magisus (approved):
I think this looks alright, but Justin has more context.

justinstoller:
Do  you want to add a unit test for the puppetfile spec that passing `{ puppetfile_path: "foo/Puppetfile" }` etc is honored?

mwaggett:
@justinstoller tests added 😊

spec/unit/puppetfile_spec.rb
```diff
@@ -10,14 +10,35 @@
...
     it "is the basedir joined with '/Puppetfile.r10k' path" do
       expect(subject.puppetfile_path).to eq '/some/nonexistent/basedir/Puppetfile.r10k'
     end
   end
 
 end
 
+describe R10K::Puppetfile do
+
+  relative_subject = described_class.new('/some/nonexistent/basedir',
```
justinstoller: I think these subjects should go in the `it` blocks, right?

mwaggett: I copied this from the test above, which has the `subject` outside of the `it` block. Let me know if you still think I should move them.

justinstoller: The above tests use the `subject` RSpec DSL method, see: https://relishapp.com/rspec/rspec-core/docs/subject/explicit-subject .

These appear to be local variables that for some reason are still be in scope when the it block is evaluated.

mwaggett: yes, I know. I assume you're saying "yes, change it"?

mwaggett: updated

justinstoller: Yes.

spec/unit/puppetfile_spec.rb
```diff
@@ -10,14 +10,35 @@
...
       expect(subject.puppetfile_path).to eq '/some/nonexistent/basedir/Puppetfile.r10k'
     end
   end
 
 end
 
+describe R10K::Puppetfile do
+
+  describe "a custom relative puppetfile_path" do
+    relative_subject = described_class.new('/some/nonexistent/basedir',
```
justinstoller: These should be in the `it` block.

justinstoller (approved):

PR #1206, opened 2021-08-13
(maint) Improve acceptance test verbosity

mwaggett:

### DIFF
--- a/integration/tests/user_scenario/basic_workflow/single_env_purge_unmanaged_modules.rb
+++ b/integration/tests/user_scenario/basic_workflow/single_env_purge_unmanaged_modules.rb
@@ -57,7 +57,7 @@ step 'Push Changes'
 git_add_commit_push(master, 'production', 'Update site.pp and add module.', git_remote_environments_path)
 
 step 'Deploy Environments via r10k'
-on(master, "#{r10k_fqp} deploy environment --modules --verbose debug")
+on(master, "#{r10k_fqp} deploy environment --modules --verbose debug --trace")
 
 step 'Manually Install the "motd" Module from the Forge'
 on(master, puppet("module install puppetlabs-motd --modulepath #{moduledir}"))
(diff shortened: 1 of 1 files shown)


Magisus (approved):

integration/tests/user_scenario/basic_workflow/single_env_purge_unmanaged_modules.rb
```diff
@@ -57,7 +57,7 @@ class { 'motd':
 git_add_commit_push(master, 'production', 'Update site.pp and add module.', git_remote_environments_path)
 
 step 'Deploy Environments via r10k'
-on(master, "#{r10k_fqp} deploy environment --modules --verbose")
+on(master, "#{r10k_fqp} deploy environment --modules --verbose debug --trace")
```
justinstoller, line 60: Kinda iffy on this.

If this fails then there's something fundamentally wrong with the system and test set up (ie. this r10k invocation is incidental to the point of the test)

mwaggett: iffy on what? still seems like it could be useful if this _does_ fail for some reason..

justinstoller: Sure, but that should be no more likely than any other part of the test setup. ie, wouldn't it be just as helpful to turn debug output on for any other command in the file, or all the cli commands in the test?

mwaggett: I just turned it on for all `r10k` invocations and that's it ¯\_(ツ)_/¯

Magisus: Agreed that it would be better to do this more globally as a slightly more involved piece of work.

mwaggett: you're not wrong, just not gonna do that now 😝 

All these tests could use an audit, but yeah that is a bigger, more-involved effort.

justinstoller: That didn't go how I thought it would.

I think collecting all the logs is only effective if it's trivial to collect, store, and filter/process those logs. I would not expect a human to watch and/or manually go through a beaker log for a test where every command line invocation (of r10k, puppet, git, etc) was logging at debug level). I think that would add so much noise that any signal would be drowned out. I, frankly, feel like the tests are pretty close to that already. I agree with having debug output for the commands under test, but I generally think things like test setup and teardown should log at warn or info type levels.

PR #1208, opened 2021-08-16
Drop EoL Ruby 2.3/2.4 support

bastelfreak:
This is a follow up of https://github.com/puppetlabs/r10k/pull/1196

```
```

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -7,6 +7,7 @@ Unreleased
 - (RK-383) Remove deprecated `basedir` method from Puppetfile DSL. Users should use `environment_name` instead. [#1254](https://github.com/puppetlabs/r10k/pull/1254)
 - (RK-386) Remove deprecated `bare` environment type. [#1235](https://github.com/puppetlabs/r10k/issues/1235)
 - Refactor internal module creation to always expect a hash, even for Forge modules, which can be specified in the Puppetfile with just a version string. [#1170](https://github.com/puppetlabs/r10k/pull/1170)
+- Drop EoL Ruby 2.3/2.4 support [#1280](https://github.com/puppetlabs/r10k/pull/1208)
 
 3.14.0
 ------
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -18,23 +18,19 @@ Gem::Specification.new do |s|
     dynamic environments.
   DESCRIPTION
 
-  s.required_ruby_version = '>= 2.3.0'
+  s.required_ruby_version = '>= 2.5.0'
 
   s.license  = 'Apache-2.0'
 
   s.add_dependency 'colored2',   '3.1.2'
-  s.add_dependency 'cri', '2.15.10'
+  s.add_dependency 'cri', '>= 2.15.10'
 
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'
 
-  s.add_dependency 'puppet_forge', '~> 2.3.0'
+  s.add_dependency 'puppet_forge', '>= 2.3.0', '< 4'
 
   s.add_dependency 'gettext-setup', '~>0.24'
-  # These two pins narrow what is allowed by gettext-setup,
-  # to preserve compatability with Ruby 2.4
-  s.add_dependency 'fast_gettext', '~> 1.1.0'
-  s.add_dependency 'gettext', ['>= 3.0.2', '< 3.3.0']
 
   s.add_dependency 'jwt', '~> 2.2.3'
   s.add_dependency 'minitar', '~> 0.9'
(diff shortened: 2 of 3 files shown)


mwaggett:
hey @bastelfreak, thanks for opening this! Unfortunately, we don't feel comfortable dropping support for these Ruby versions until a major release, which we are not prepared to do right now. I wish I could give you an estimate of when that might be, but there are a lot of other competing priorities right now. 😕

bastelfreak:
Hey @mwaggett ,
can you provide any update on this? What's the issue with releasing a new major version? Because of the recovation of https://github.com/puppetlabs/r10k/pull/1197 it's not possible to install the gem on Ruby 3 which is the default system version on multiple operating systems (also I don't understand why you want to keep supporting EoL Ruby 2.3/2.4).

mwaggett:
@bastelfreak Thanks for your patience on this. We have a lot we were hoping to do for our next major release, but not a lot of time to do it, which is why it hasn't happened yet. But we'll look into possibly releasing a new major version with just updated ruby versions and get to the rest later - we'll just have to check our Product Owner.

Magisus:
So I went ahead and created a `4.x` branch, would you mind retargeting this there? I'm hoping to start accumulating changes for that major release, and it makes sense to do it on a new branch stream.

Magisus:
Related tickets: https://tickets.puppetlabs.com/browse/RK-373 and https://tickets.puppetlabs.com/browse/RK-377.

bastelfreak:
@mwaggett awesome to see some progress here! I changed the target branch of the PR and rebased.

Magisus:
Oh right, we do Ruby 2.3 testing in Travis. I'm going to rework that config to drop the Ruby 2.3 unit testing and just run container tests.

I think I also need to update the Github Actions config to run on the 4.x branch.

I'll post back here when that's ready.

Magisus:
Okay those fixes are here: https://github.com/puppetlabs/r10k/pull/1256. Once that goes in, we can rebase this again and see stuff run for real.

CHANGELOG.mkd
```diff
@@ -23,6 +23,7 @@ Unreleased
 - Purge should remove unmanaged directories, in addition to unmanaged files. [#1222](https://github.com/puppetlabs/r10k/pull/1222)
 - Rename experimental environment type "bare" to "plain". [#1228](https://github.com/puppetlabs/r10k/pull/1228)
 - Add support for specifying additional logging ouputs. [#1230](https://github.com/puppetlabs/r10k/issues/1230)
+- Drop EoL Ruby 2.3/2.4 support [#1280](https://github.com/puppetlabs/r10k/pull/1208)
```
mwaggett: This needs to move to the `Unreleased` section.

r10k.gemspec
```diff
@@ -18,23 +18,21 @@ Gem::Specification.new do |s|
...
-  s.add_dependency 'puppet_forge', '~> 2.3.0'
+  s.add_dependency 'puppet_forge', '>= 2.3.0', '< 4'
 
   s.add_dependency 'gettext-setup', '~>0.24'
-  # These two pins narrow what is allowed by gettext-setup,
-  # to preserve compatability with Ruby 2.4
-  s.add_dependency 'fast_gettext', '~> 1.1.0'
-  s.add_dependency 'gettext', ['>= 3.0.2', '< 3.3.0']
+  s.add_dependency 'fast_gettext', '>= 1.1.0', '< 3'
+  s.add_dependency 'gettext', '>= 3.0.2', '< 4.0.0'
```
mwaggett: Do we need these pins at all anymore?

bastelfreak: good question. for the upper pinning this is a good practice in many modules. Recently psych and faraday hat major releases with a lot of breaking changes. I think upper limits are fine, but I don't have a string opinion here. for the lower limits I am not sure whats the oldest working version / if we should only have an upper limit.

CHANGELOG.mkd
bastelfreak: ah that got reordered during my rebase, will fix that later.

r10k.gemspec
mwaggett: I just meant that it looks like these are transitive dependencies via `gettext-setup`, rather than direct dependencies of r10k, I don't think we should pin them explicitly unless we need to. I would hope that `gettext-setup` does its own pinning to prevent breaking changes from its dependencies and that we could just pin `gettext-setup` if we need to.

CHANGELOG.mkd
bastelfreak: fixed it

r10k.gemspec
bastelfreak: okay, I think this is good to go now.

mwaggett: thanks so much!

PR #1210, opened 2021-08-19
(CODEMGMT-1454) Ensure missing repo caches are re-synced

mwaggett:
This commit updates the `status` method for stateful git repos to check for the
presence of the git cache when determining whether repo updates are necessary.
Previously, if the repo's working dir and git dir were present, but the cache
was missing, the `@repo.dirty?` check would fail when attempting to diff the
workdir (`object not found - no match for id`). This would occur during an
environment deploy if the environment used the same remote repo as another
environment that had been deleted (and thus its artifacts cleaned up). By
setting the status to `:mismatched` when the cache doesn't exist, we ensure that
the cache is resynced and the deployment may proceed.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,8 @@ CHANGELOG
 Unreleased
 ----------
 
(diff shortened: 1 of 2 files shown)


Magisus (approved):

Magisus:
We're going to add a test for the scenario that revealed this bug at a different level (in PE).

PR #1211, opened 2021-08-30
(CODEMGMT-1415) Sundry Improvements

justinstoller:
* Fixes regression where on some module content would be deleted on subsequent runs when not using `--modules`
* Renames `--assume-unchanged` flag to `--incremental`
* Adds CHANGELOG entry

Still to do
* [ ] Acceptance test

### DIFF
--- a/integration/tests/basic_functionality/basic_deployment.rb
+++ b/integration/tests/basic_functionality/basic_deployment.rb
@@ -7,7 +7,7 @@ test_name 'Basic Environment Deployment Workflows'
 
 # This isn't a block because we want to use the local variables throughout the file
 step 'init'
-  env_path              = on(master, puppet('config print environmentpath')).stdout.rstrip
+  @env_path             = on(master, puppet('config print environmentpath')).stdout.rstrip
   r10k_fqp              = get_r10k_fqp(master)
 
   control_repo_gitdir   = '/git_repos/environments.git'
@@ -29,7 +29,7 @@ step 'init'
       provider: '#{git_provider}'
     sources:
       control:
-        basedir: "#{env_path}"
+        basedir: "#{@env_path}"
         remote: "#{control_repo_gitdir}"
     deploy:
       purge_levels: ['deployment','environment','puppetfile']
@@ -38,7 +38,7 @@ step 'init'
 
 
 def and_stdlib_is_correct
-  metadata_path = "#{env_path}/production/modules/stdlib/metadata.json"
+  metadata_path = "#{@env_path}/production/modules/stdlib/metadata.json"
   on(master, "test -f #{metadata_path}", accept_all_exit_codes: true) do |result|
     assert(result.exit_code == 0, 'stdlib content has been inappropriately purged')
   end
@@ -62,13 +62,15 @@ step 'Set up r10k and control repo' do
   create_remote_file(master, "#{control_repo_worktree}/Puppetfile", puppetfile1)
   git_add_commit_push(master, 'production', 'add Puppetfile for Basic Deployment test', control_repo_worktree)
 
+  # Ensure the production environment will be deployed anew
+  on(master, "rm -rf #{@env_path}/production")
 end
 
-test_path = "#{env_path}/production/modules/apache/metadata.json"
+test_path = "#{@env_path}/production/modules/apache/metadata.json"
 step 'Test initial environment deploy works' do
   on(master, "#{r10k_fqp} deploy environment production --verbose=info") do |result|
-    assert(result.stdout =~ /.*Deploying module to .*apache.*/, 'Did not log apache deployment')
-    assert(result.stdout =~ /.*Deploying module to .*stdlib.*/, 'Did not log stdlib deployment')
+    assert(result.output =~ /.*Deploying module to .*apache.*/, 'Did not log apache deployment')
+    assert(result.output =~ /.*Deploying module to .*stdlib.*/, 'Did not log stdlib deployment')
   end
   on(master, "test -f #{test_path}", accept_all_exit_codes: true) do |result|
     assert(result.exit_code == 0, 'Expected module in Puppetfile was not installed')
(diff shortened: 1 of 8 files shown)


Magisus (approved):
I think the first commit could use a short description of the bug.

Are you planning to add the acceptance test as part of this PR, or should we merge this first to unblock code-manager's unit tests?

justinstoller:
I was planning on doing the acceptance test as part of this work. We should see how long it takes me to run the damned things, if I don't have it by tomorrow we can merge, I guess.

Magisus:
I'm fine waiting for now, I don't think we're landing anything more in code-manager right away.

spec/unit/action/deploy/environment_spec.rb
```diff
@@ -147,11 +147,11 @@
           and_return(loader).at_least(:once)
       end
 
-      it "assume unchanged flag causes the module definitons to be preloaded by the loader" do
+      it "incremental unchanged flag causes the module definitons to be preloaded by the loader" do
```
mwaggett:
```suggestion
      it "incremental flag causes the module definitons to be preloaded by the loader" do
```

mwaggett (commented):
+1 to updating the commit message to describe the bug

spec/unit/action/deploy/environment_spec.rb
justinstoller: smh, I guess it really shows that my search term was just "assume"

justinstoller:
I'm having a lot of trouble running the acceptance tests locally. I think it's at a place where it can be reviewed at least while I continue to try and run the test.

I also believe I've address the comments so far and added a new commit cleaning up some logging.

Magisus:
Cool, I'll take a look, and also try to run your test.

Magisus (approved):
I think this looks good so long as the test passes. It checks all the boxes I would want to see for the basic workflow.

integration/tests/basic_functionality/basic_deployment.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  new_apache_info = JSON.parse(on(master, "cat #{test_path}").stdout)
+  apache_major_version = new_apache_info['version'].split('.').first.to_i
+  assert(apache_major_version > 5, 'Module not updated correctly using --modules')
+
+  and_stdlib_is_correct
+end
+
+step 'Test --modules --incremental updates changed modules' do
+  on(master, "#{r10k_fqp} deploy environment production --modules --incremental --verbose=debug1") do |result|
```
mwaggett, line 155: Without a new Puppetfile, this deploy isn't actually changing anything, is it? Is there a better way we can test this behavior other than checking the log messages? Just trying to think in terms of actual workflows, since I don't imagine someone would run a deploy with `--modules` and then again with `--modules --incremental` without at least making some sort of change in between?

integration/tests/basic_functionality/basic_deployment.rb
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  and_stdlib_is_correct
+end
+
+original_apache_info = JSON.parse(on(master, "cat #{test_path}").stdout)
+
+step 'Test second run of deploy updates control repo, but not modules' do
+  puppetfile2 =<<-EOS
+    mod 'puppetlabs/apache', :latest
```
mwaggett, line 87: Might be good to add a comment on what the latest version is at the time of writing this. Could help with future debugging and also explains why `apache_major_version > 5` is an appropriate check.

integration/tests/basic_functionality/basic_deployment.rb
justinstoller: Yeah, I struggled with this. The end result of running `--modules` vs `--modules --incremental` should be the same, modules that should be updated will be updated. However, the how it happens is different, with some modules being skipped completely and some still being checked. So the logging shows us at least that the static version checking is correct (for a forge module) and that unchanged static modules are skipped.

If you think I should I could include all the permutations of different types of versions as well as those that have changed definitions? Or are you mainly concerned about those modules whose definition changed on disk part? I could include one of those in the Puppetfile.

justinstoller: I'll add a module that is updated in this invocation to test the changeable vs static, changed vs static, unchanged. Let me know if you'd like more permutations of static version checking (fwiw, there is some unit tests around that: https://github.com/puppetlabs/r10k/commit/6f46384633531d82df10d490183259eaccd8ad93#diff-1bc365ddca4e6e4a7ba95ff4d655c036faa16ec1025e35bc559476265a5719aaR344-R389)

justinstoller:
I've tested this test, with current HEAD of pe-r10k ([80392485](http://builds.delivery.puppetlabs.net/pe-r10k/803924852a2619035e83b7ada4b3e1637feb6783/?C=M&O=D)):
```
              - Test Case Summary for suite 'tests' -
       Total Suite Time: 10.60 seconds
      Average Test Time: 10.60 seconds
              Attempted: 1
                 Passed: 0
                 Failed: 1
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1

      - Specific Test Case Status -
        
Failed Tests Cases:
  Test Case tests/basic_functionality/basic_deployment.rb reported: #<Minitest::Assertion: Expected module content in Puppetfile was inappropriately purged>
    Test line: tests/basic_functionality/basic_deployment.rb:102:in `block (3 levels) in run_test'
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:
```
As well as this content ([e5b714d29](http://builds.delivery.puppetlabs.net/pe-r10k/e5b714d29a4597187fe3b1938b84a73f10175b78/?C=M&O=D)):
```

              - Test Case Summary for suite 'tests' -
       Total Suite Time: 13.73 seconds
      Average Test Time: 13.73 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1

      - Specific Test Case Status -
        
Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:
```

justinstoller:
Running the whole suite now

integration/tests/basic_functionality/basic_deployment.rb
mwaggett: I think what you added looks good! I'm satisfied with the various permutations we test here.

mwaggett (approved):
spongebob giving many thumbs up

mwaggett:
Feel free to merge once you're satisfied with your suite run.

PR #1214, opened 2021-09-07
(RK-308) Allow respecting `forge` declaration in puppetfile

Magisus:
This commit adds a new setting `forge.allow_puppetfile_override`, that
when true will cause a `forge` declaration in a Puppetfile to override
the `forge.baseurl` setting for contacting the Forge. This allows users
to configure per-environment Forge URLs, for example, if an internal
Forge is used for a test environment and the live Forge for production.
Previously, a user would have had to choose one or the other. The
setting is false by default, since many users have erroneous `forge`
declarations in their Puppetfiles that were previously completely
ignored.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,7 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
-- (RK-308) Provide a `forge.allow_puppetfile_override` setting that, when true causes a `forge` declaration in the Puppetfile to override `forge.baseurl`. [#1214](https://github.com/puppetlabs/r10k/pull/1214)
+- (RK-308) Provide a `forge.allow_puppetfile_override` setting that, when true, causes a `forge` declaration in the Puppetfile to override `forge.baseurl`. [#1214](https://github.com/puppetlabs/r10k/pull/1214)
 - (CODEMGMT-1415) Provide a `--incremental` flag to only sync those modules in a Puppetfile whose definitions have changed since last sync, or those whose versions could change. [#1200](https://github.com/puppetlabs/r10k/pull/1200)
 - (CODEMGMT-1454) Ensure missing repo caches are re-synced [#1210](https://github.com/puppetlabs/r10k/pull/1210)
 - (PF-2437) Allow token authentication to be used with the Forge. [#1192](https://github.com/puppetlabs/r10k/pull/1192)
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
+- (RK-308) Provide a `forge.allow_puppetfile_override` setting that, when true causes a `forge` declaration in the Puppetfile to override `forge.baseurl`. [#1214](https://github.com/puppetlabs/r10k/pull/1214)
 - (CODEMGMT-1415) Provide a `--incremental` flag to only sync those modules in a Puppetfile whose definitions have changed since last sync, or those whose versions could change. [#1200](https://github.com/puppetlabs/r10k/pull/1200)
 - (CODEMGMT-1454) Ensure missing repo caches are re-synced [#1210](https://github.com/puppetlabs/r10k/pull/1210)
 - (PF-2437) Allow token authentication to be used with the Forge. [#1192](https://github.com/puppetlabs/r10k/pull/1192)
--- a/doc/puppetfile.mkd
+++ b/doc/puppetfile.mkd
@@ -51,10 +51,9 @@ handles modules.
 ### forge
 
 The `forge` setting specifies which server that Forge based modules are fetched
-from. This is currently a noop and is provided for compatibility with
-librarian-puppet.
-
-R10k supports setting the Forge to use _globally_ in `r10k.yaml`. see [Configuration](/doc/dynamic-environments/configuration.mkd#baseurl) for details.
+from. This declaration is only respected if [`forge.allow_puppetfile_override`](/dynamic-environments/configuration.mkd#allow_puppetfile_override)
+is set to true in the main `r10k.yaml`. Otherwise, use [`forge.baseurl`](/doc/dynamic-environments/configuration.mkd#baseurl)
+to globally configure where modules should be downloaded from.
 
 ### moduledir
 
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -9,7 +9,6 @@ module R10K
 
       DEFAULT_MODULEDIR = 'modules'
       DEFAULT_PUPPETFILE_NAME = 'Puppetfile'
-      DEFAULT_FORGE_API = 'forgeapi.puppetlabs.com'
 
       attr_accessor :default_branch_override, :environment
       attr_reader :modules, :moduledir, :puppetfile_path,
@@ -29,14 +28,12 @@ module R10K
       def initialize(basedir:,
                      moduledir: DEFAULT_MODULEDIR,
                      puppetfile: DEFAULT_PUPPETFILE_NAME,
-                     forge: DEFAULT_FORGE_API,
                      overrides: {},
                      environment: nil)
 
         @basedir     = cleanpath(basedir)
         @moduledir   = resolve_path(@basedir, moduledir)
         @puppetfile_path  = resolve_path(@basedir, puppetfile)
-        @forge       = forge
         @overrides   = overrides
         @environment = environment
         @default_branch_override = @overrides.dig(:environments, :default_branch_override)
@@ -114,13 +111,11 @@ module R10K
 
       # @param [String] forge
       def set_forge(forge)
-        @forge = forge
         if @allow_puppetfile_forge
           logger.debug _("Using Forge from Puppetfile: %{forge}") % { forge: forge }
-          # This method will append a trailing slash to the string, in place, so we
-          # dup it to avoid different behavior in the value of our instance var
-          # when `allow_puppetfile_forge` is true vs false
-          PuppetForge.host = forge.dup
+          PuppetForge.host = forge
+        else
+          logger.debug _("Ignoring Forge declaration in Puppetfile, using value from settings: %{forge}.") % { forge: PuppetForge.host }
         end
       end
 
--- a/lib/r10k/module_loader/puppetfile.rb
+++ b/lib/r10k/module_loader/puppetfile.rb
@@ -40,6 +40,7 @@ module R10K
         @overrides   = overrides
         @environment = environment
         @default_branch_override = @overrides.dig(:environments, :default_branch_override)
+        @allow_puppetfile_forge = @overrides.dig(:forge, :allow_puppetfile_override)
 
         @existing_module_metadata = []
         @existing_module_versions_by_name = {}
@@ -114,6 +115,13 @@ module R10K
       # @param [String] forge
       def set_forge(forge)
         @forge = forge
+        if @allow_puppetfile_forge
+          logger.debug _("Using Forge from Puppetfile: %{forge}") % { forge: forge }
+          # This method will append a trailing slash to the string, in place, so we
+          # dup it to avoid different behavior in the value of our instance var
+          # when `allow_puppetfile_forge` is true vs false
+          PuppetForge.host = forge.dup
+        end
       end
 
       # @param [String] moduledir
--- a/spec/unit/module_loader/puppetfile_spec.rb
+++ b/spec/unit/module_loader/puppetfile_spec.rb
@@ -276,6 +276,30 @@ describe R10K::ModuleLoader::Puppetfile do
       end
     end
 
+    describe 'forge declaration' do
+      before(:each) do
+        PuppetForge.host = ""
+      end
+
+      it 'is respected if `allow_puppetfile_override` is true' do
+        @path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'forge-override')
+        puppetfile = R10K::ModuleLoader::Puppetfile.new(basedir: @path, overrides: { forge: { allow_puppetfile_override: true } })
+        puppetfile.load!
+        expect(puppetfile.instance_variable_get(:@forge)).to eq("my.custom.forge.com")
+        expect(PuppetForge.host).to eq("my.custom.forge.com/")
+      end
+
+      it 'is ignored if `allow_puppetfile_override` is false' do
+        @path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'forge-override')
+        puppetfile = R10K::ModuleLoader::Puppetfile.new(basedir: @path, overrides: { forge: { allow_puppetfile_override: false } })
+        expect(PuppetForge).not_to receive(:host=).with("my.custom.forge.com")
+        puppetfile.load!
+        # instance variable is set, but PuppetForge.host is not updated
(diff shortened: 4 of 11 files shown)


Magisus:
So if I understand right, we shouldn't need to do anything to the `set_forge` [function](https://github.com/puppetlabs/r10k/blob/4a832b7e0090ca2ec04c60f48ce43b98a6b16d5f/lib/r10k/puppetfile.rb#L146-L148) in the old Puppetfile class, because it will just delegate to the one in `ModuleLoader::Puppetfile`.

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -114,6 +115,13 @@ def add_module_metadata(name, info)
       # @param [String] forge
       def set_forge(forge)
         @forge = forge
+        if @allow_puppetfile_forge
+          logger.debug _("Using Forge from Puppetfile: %{forge}") % { forge: forge }
+          # This method will append a trailing slash to the string, in place, so we
+          # dup it to avoid different behavior in the value of our instance var
+          # when `allow_puppetfile_forge` is true vs false
+          PuppetForge.host = forge.dup
```
Magisus: The dup is probably unnecessary, seeing as we don't really use this instance var for anything that I can see. Lemme know what you think. Also, if we're not using `@forge`... do we need the instance variable at all?

Magisus:
Manually tested this, seems to work.

Magisus:
Note this does not support custom auth for the overridden Forge. It will use the same auth as the normal `baseurl`, if any is configured. Fixing that is a separate effort.

justinstoller:
Looks good, but I think docs need updating for the setting.

cdenneen:
> Note this does not support custom auth for the overridden Forge. It will use the same auth as the normal `baseurl`, if any is configured. Fixing that is a separate effort.

So this won’t work with the Bearer token fix? Is that work being tracked anywhere? 

For some reason I thought here it would work if done as Basic Auth but that might have been just —module_repository parameter for .fixtures.yml

Magisus:
If you configure the Bearer token in the `forge` section of `r10k.yaml`, that token will be used with all forges declared in Puppetfiles. That might work for some use cases, but likely not for all of them. We should file a ticket about supporting per-environment credentials if you know that will be needed. I'm not sure what that should look like from a security/UX perspective...

I'm not super familiar with the Basic Auth usecase... I think I saw some ticket where someone thought it didn't work even now, with regular `baseurl`, but I'm not sure.

cdenneen:
> If you configure the Bearer token in the `forge` section of `r10k.yaml`, that token will be used with all forges declared in Puppetfiles. That might work for some use cases, but likely not for all of them. We should file a ticket about supporting per-environment credentials if you know that will be needed. I'm not sure what that should look like from a security/UX perspective...
> 
> I'm not super familiar with the Basic Auth usecase... I think I saw some ticket where someone thought it didn't work even now, with regular `baseurl`, but I'm not sure.

Basic Auth would be https://user:pass@forge/api

Magisus:
If you already know that basic auth does work (I have not been able to try it, but might be able to set something up), and are comfortable putting a password in your Puppetfile, then I think that could be a workaround. I sort of ruled it out because Puppetfiles are generally committed to source control, and putting a password in them seems like a bad idea. But I guess that would depend a lot on your individual context.

CHANGELOG.mkd
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
+- (RK-308) Provide a `forge.allow_puppetfile_override` setting, which when true causes a `forge` declaration in the Puppetfile to override `forge.baseurl` [#1214](https://github.com/puppetlabs/r10k/pull/1214)
```
mwaggett: (nit)
```suggestion
- (RK-308) Provide a `forge.allow_puppetfile_override` setting that, when true, causes a `forge` declaration in the Puppetfile to override `forge.baseurl`. [#1214](https://github.com/puppetlabs/r10k/pull/1214)
```

doc/puppetfile.mkd
```diff
@@ -51,10 +51,9 @@ handles modules.
 ### forge
 
 The `forge` setting specifies which server that Forge based modules are fetched
-from. This is currently a noop and is provided for compatibility with
-librarian-puppet.
-
-R10k supports setting the Forge to use _globally_ in `r10k.yaml`. see [Configuration](/doc/dynamic-environments/configuration.mkd#baseurl) for details.
+from. This declaration is only respected if [`forge.allow_puppetfile_override`](/dynamic-environments/configuration.mkd#allow_puppetfile_override)
+is set to true in the main `r10k.yaml`. Otherwise, use [`forge.baseurl`](/doc/dynamic-environments/configuration.mkd#baseurl)
+to configure where modules should be downloaded from.
```
mwaggett: (nit)
```suggestion
to globally configure where modules should be downloaded from.
```

lib/r10k/module_loader/puppetfile.rb
mwaggett: yeah, this is confusing to me.. I would be in favor of removing `@forge` and the various `set_forge` methods if they're not actually doing anything, since it sorta seems like they're not? but it's unclear. Or perhaps replace the `set_forge` methods with ones that actually do something.. I would have imagined something calling `set_forge(forge_from_puppetfile)` and then having `PuppetForge.host = @forge` somewhere where it's actually being used? But I don't really know, everything is a mess haha.

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -276,6 +276,30 @@ def expect_wrapped_error(error, pf_path, error_type)
...
+        expect(PuppetForge.host).to eq("my.custom.forge.com/")
+      end
+
+      it 'is ignored if `allow_puppetfile_override` is false' do
+        @path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'forge-override')
+        puppetfile = R10K::ModuleLoader::Puppetfile.new(basedir: @path, overrides: { forge: { allow_puppetfile_override: false } })
+        expect(PuppetForge).not_to receive(:host=).with("my.custom.forge.com")
+        puppetfile.load!
+        # instance variable is set, but PuppetForge.host is not updated
+        expect(puppetfile.instance_variable_get(:@forge)).to eq("my.custom.forge.com")
```
mwaggett: Regardless of what we decide to do with `@forge`, I think it's kinda weird to test this, since we don't actually want to be using "my.custom.forge.com" in this case.. It would be ideal imo if `@forge` was equal to the default. ¯\_(ツ)_/¯

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -276,6 +276,30 @@ def expect_wrapped_error(error, pf_path, error_type)
...
+        expect(PuppetForge.host).to eq("/")
```
mwaggett, line 287: Why doesn't this equal the default Forge value? It seems super weird to me to have various places where we set `PuppetForge.host` and that those places are different from where that host is actually used??

lib/r10k/module_loader/puppetfile.rb
Magisus: As the comment at the top of this block says, `set_forge` is part of the Puppetfile parsing DSL. It is [called](https://github.com/puppetlabs/r10k/blob/4a832b7e0090ca2ec04c60f48ce43b98a6b16d5f/lib/r10k/module_loader/puppetfile/dsl.rb#L29) when the Puppetfile is [evaluated](https://github.com/puppetlabs/r10k/blob/4a832b7e0090ca2ec04c60f48ce43b98a6b16d5f/lib/r10k/module_loader/puppetfile.rb#L63-L64) by that DSL. Until now, it has set this instance variable, but been a noop since, so far as I could tell, the var isn't used (we were deliberately ignoring `forge` declarations in the Puppetfile).

The actual way to make the `forge` declaration be used is to set the `host` field for the `PuppetForge` gem. That's what we do when [reading it from settings](https://github.com/puppetlabs/r10k/blob/4a832b7e0090ca2ec04c60f48ce43b98a6b16d5f/lib/r10k/initializers.rb#L67). I didn't delete the instance variable, in case something else is expecting it be set, but I think maybe that's not the case. This seems even more likely given there's no getter for it. I can try that and see what happens.

spec/unit/module_loader/puppetfile_spec.rb
Magisus: Yeah I see what you're saying, but I think I'm still kind of in favor of dropping the instance variable.

spec/unit/module_loader/puppetfile_spec.rb
Magisus: This is an artifact of the test: this is a unit test of the `ModuleLoader::Puppetfile`, so the thing that would normally set `PuppetForge.host` (settings init) doesn't happen here. If you notice above, I've provided a default value in a `before` block. If I didn't do that, the state bled between tests, because `host` is a class variable of `PuppetForge`. I could use a _different_ default value, but that doesn't change the nature of the test.

Magisus:
Updated to remove the `@forge` instance variable. Let me know what you think.

lib/r10k/module_loader/puppetfile.rb
mwaggett: yeah, I know the DSL has a method that calls this, but I didn't see anywhere calling the DSL's `forge` method, so wondering if that could be removed too.

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: that works too.

spec/unit/module_loader/puppetfile_spec.rb
mwaggett: oh right, I forgot you were resetting it by the time I got to this part haha. so if it was a single test where you didn't have the `before` block, `PuppetForge.host` would be the default value? Is that our default or the forge gem default?

CHANGELOG.mkd
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
+- (RK-308) Provide a `forge.allow_puppetfile_override` setting that, when true causes a `forge` declaration in the Puppetfile to override `forge.baseurl`. [#1214](https://github.com/puppetlabs/r10k/pull/1214)
```
```suggestion
- (RK-308) Provide a `forge.allow_puppetfile_override` setting that, when true, causes a `forge` declaration in the Puppetfile to override `forge.baseurl`. [#1214](https://github.com/puppetlabs/r10k/pull/1214)
```

spec/unit/module_loader/puppetfile_spec.rb
```diff
@@ -276,6 +267,27 @@ def expect_wrapped_error(error, pf_path, error_type)
...
+        @path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'forge-override')
+        puppetfile = R10K::ModuleLoader::Puppetfile.new(basedir: @path, overrides: { forge: { allow_puppetfile_override: true } })
+        puppetfile.load!
+        expect(PuppetForge.host).to eq("my.custom.forge.com/")
+      end
+
+      it 'is ignored if `allow_puppetfile_override` is false' do
+        @path = File.join(PROJECT_ROOT, 'spec', 'fixtures', 'unit', 'puppetfile', 'forge-override')
+        puppetfile = R10K::ModuleLoader::Puppetfile.new(basedir: @path, overrides: { forge: { allow_puppetfile_override: false } })
+        expect(PuppetForge).not_to receive(:host=).with("my.custom.forge.com")
```
mwaggett, line 285: Is this necessary if we're testing the value of `PuppetForge.host` below?

lib/r10k/module_loader/puppetfile.rb
```diff
@@ -113,7 +111,10 @@ def add_module_metadata(name, info)
 
       # @param [String] forge
       def set_forge(forge)
-        @forge = forge
+        if @allow_puppetfile_forge
+          logger.debug _("Using Forge from Puppetfile: %{forge}") % { forge: forge }
```
mwaggett, line 115: Should we also log that we're ignoring the `forge` setting from the Puppetfile if `@allow_puppetfile_forge` isn't true?

lib/r10k/module_loader/puppetfile.rb
Magisus: The DSL represents an implementation of how to evalute a Puppetfile as a Ruby script. So if a Puppetfile has
```
forge "my.awesome.forge

mod "puppetlabs/apache"
```
`forge` and `mod` there are Ruby function calls, where the definition is what's in the DSL class. So when `instance_eval` happens, we evaluate the Ruby script that is the Puppetfile in the context of the DSL class, which has those methods defined, effectively calling `forge`, then `mod`.

I was trying to look for a good example of using `instance_eval` this way, but it seems like most people use it to add methods to individual objects, instead of actually executing code directly like we're doing here. The docs might help though: https://www.rubydoc.info/stdlib/core/BasicObject:instance_eval

spec/unit/module_loader/puppetfile_spec.rb
Magisus: It would be the Forge gem's default, because even with one test, our code that sets it is not getting executed in this context. I don't know if that default is something valid, or just `nil`, I'd have to check.

CHANGELOG.mkd
Magisus: Honestly I don't really think either of these commas is necessary. The clause is short enough that it doesn't really need to be offset. But meh 🤷‍♀️

lib/r10k/module_loader/puppetfile.rb
Magisus: I could go either way, it's default behavior and has been that way a really long time. But I can add it.

spec/unit/module_loader/puppetfile_spec.rb
Magisus: What I like about this one is it ties us back to the specific value from the Puppetfile that we want to be ignored.

lib/r10k/module_loader/puppetfile.rb
mwaggett: oh right, ugh. didn't realize just `grep`ing wouldn't be enough 😅  but that makes sense.

CHANGELOG.mkd
mwaggett: well one was definitely wrong 😝  I usually just add them if it helps me parse the sentence ¯\_(ツ)_/¯

lib/r10k/module_loader/puppetfile.rb
mwaggett: Do we think we'd want to make this the default behavior in the next major release? If so, I think it could be good to log a message about it. ¯\_(ツ)_/¯ 
Otherwise, if we think this will be opt-in forever, I'm good with leaving it out.

Magisus: I seriously doubt we will ever make this the default behavior. We're leaning away from having people put more things in their puppetfiles.

CHANGELOG.mkd
Magisus: Yeah agreed, that was a typo. I put it in.

lib/r10k/module_loader/puppetfile.rb
mwaggett: In that case, I'm good with leaving out the logging. (sorry for the back and forth 😅 )

Magisus: Having tested this out, I think I'll leave it in, actually. There's isn't any other logging currently at `debug` or `debug1` that tells you which forge you're using.

mwaggett (approved):

Issue #1215, opened 2021-09-10
postrun script gets called with environments that weren't updated.


alexjfisher:
## Describe the Bug
3.11.0 contained https://github.com/puppetlabs/r10k/pull/1190
With this change, the postrun command is run after using `r10k deploy module`.

Unfortunately, the script is not called with only the environments that were updated, but with all environments that contain the module (even if they weren't updated).

## Expected Behavior
The postrun `$modifiedenvs` variable to only include environments that were actually modified.

## Steps to Reproduce

This can probably be reproduced in lots of way, but in my environment, I have a Puppetfile containing in house modules configured a bit like...
```ruby
mod 'somemod',  :git => 'ssh://git@git.example.com:7999/pupmod/somemod.git',
                :branch => :control_branch,
                :default_branch => 'main'
```

Most of the time, a module won't have many branches (possibly just the main branch), but there will be lots of control repo branches (with developers working on separate work, mainly editing profiles/hiera etc.).  Those environments will all be using `main` branch for these modules.

If you need to modify a module, you create a branch in the module repo with a branch name matching the control repo branch/puppet environment you're working on.

Pushing such a branch runs `r10k deploy module` for that module and *only* the matching control-repo branch/environment will actually get updated. But since 3.11.0 postrun (and possibly `generate types`) is then being run with *all* environments.  This is painfully slow when you have lots of environments (in my case ~150 as of today).

Magisus:
Thanks for filing this, I think [this logic](https://github.com/puppetlabs/r10k/blob/66c3bb825b6ef6ebbd90846e8a26f826467c6909/lib/r10k/action/deploy/module.rb#L109-L117) is probably at fault, I know I didn't test with `control_branch`. I'll take a look.

Magisus:
Just to clarify, you're running `r10k deploy module` without specifying any environments, right?

So what's happening is that r10k is _trying_ to update the module for all environments. At the moment, it doesn't care (and doesn't have a good way to tell) whether the module sync actually _did_ anything. And in fact, if the `main` branch of the module had changed, it _would_ do something, even for environments that don't match the new branch name, as a side effect of this workflow. It sounds like in your case that probably would never be the case, but that's out of discipline/automation, and not something general that can be guaranteed.

[...]

PR #1216, opened 2021-09-10
  [#1215] Ensure module postrun command only runs for modified envs

Magisus:
Previously, when a postrun command with interpolated environments was
specified, the `deploy module` command with no environments specified
would cause the command to be run for all environments that contained
the module, whether or not the env was actually updated by the
deploy.

Similarly, `puppet generate types` would be run for any environment that
contained the module, not just those for which it was updated. Together,
these two things could result in a huge amount of unnecessary work.

This commit makes modules return the result of their syncing, true for
"updated" and false for "not updated". The environment's `deploy`
function collects this information, recording the names of each module
that was updated as part of the deploy. The `module deploy` action then
uses this to determine which environments actually had modules modified,
and uses that to populate the `$modifiedenvs` in the postrun command. So
now, the command will only be run for environments that actually updated
a module.

This also updates the logic for running `puppet generate types` to
follow the same pattern: it is only run for environments that actually
had an updated module.

Note this does not affect environment deploys.

### DIFF
--- a/lib/r10k/action/deploy/module.rb
+++ b/lib/r10k/action/deploy/module.rb
@@ -78,16 +78,18 @@ module R10K
         ensure
           if (postcmd = @settings[:postrun])
             if @modified_envs.any?
+              envs_to_run = @modified_envs.join(' ')
+              logger.debug _("Running postrun command for environments: %{envs_to_run}.") % { envs_to_run: envs_to_run }
+
               if postcmd.grep('$modifiedenvs').any?
-                envs_to_run = @modified_envs.join(' ')
-                logger.debug "Running postrun command for environments #{envs_to_run}."
                 postcmd = postcmd.map { |e| e.gsub('$modifiedenvs', envs_to_run) }
               end
+
               subproc = R10K::Util::Subprocess.new(postcmd)
               subproc.logger = logger
               subproc.execute
             else
-              logger.debug "No environments were modified, not executing postrun command."
+              logger.debug _("No environments were modified, not executing postrun command.")
             end
           end
         end
@@ -103,11 +105,10 @@ module R10K
           else
             logger.debug1(_("Updating modules %{modules} in environment %{env_path}") % {modules: @settings.dig(:overrides, :modules, :requested_modules).inspect, env_path: environment.path})
 
-            environment.deploy
+            updated_modules = environment.deploy
 
-            requested_mods = @settings.dig(:overrides, :modules, :requested_modules) || []
             # We actually synced a module in this env
-            if !((environment.modules.map(&:name) & requested_mods).empty?)
+            if !updated_modules.nil? && !updated_modules.empty?
               # Record modified environment for postrun command
               @modified_envs << environment.dirname
 
--- a/lib/r10k/content_synchronizer.rb
+++ b/lib/r10k/content_synchronizer.rb
@@ -8,24 +8,31 @@ module R10K
     end
 
     def self.serial_sync(modules)
+      updated_modules = []
       modules.each do |mod|
-        mod.sync
+        updated = mod.sync
+        updated_modules << mod.name if updated
       end
+      updated_modules
     end
 
+    # Returns a Queue of the names of modules actually updated
     def self.concurrent_accept(modules, visitor, loader, pool_size, logger)
       mods_queue = modules_visit_queue(modules, visitor, loader)
       sync_queue(mods_queue, pool_size, logger)
     end
 
+    # Returns a Queue of the names of modules actually updated
     def self.concurrent_sync(modules, pool_size, logger)
       mods_queue = modules_sync_queue(modules)
       sync_queue(mods_queue, pool_size, logger)
     end
 
+    # Returns a Queue of the names of modules actually updated
     def self.sync_queue(mods_queue, pool_size, logger)
       logger.debug _("Updating modules with %{pool_size} threads") % {pool_size: pool_size}
-      thread_pool = pool_size.times.map { sync_thread(mods_queue, logger) }
+      updated_modules = Queue.new
+      thread_pool = pool_size.times.map { sync_thread(mods_queue, logger, updated_modules) }
       thread_exception = nil
 
       # If any threads raise an exception the deployment is considered a failure.
@@ -33,6 +40,8 @@ module R10K
       # current work, then re-raise the first exception caught.
       begin
         thread_pool.each(&:join)
+        # Return the list of all modules that were actually updated
+        updated_modules
       rescue => e
         logger.error _("Error during concurrent deploy of a module: %{message}") % {message: e.message}
         mods_queue.clear
@@ -65,11 +74,14 @@ module R10K
       modules_by_cachedir.values.each {|mods| queue << mods }
     end
 
-    def self.sync_thread(mods_queue, logger)
+    def self.sync_thread(mods_queue, logger, updated_modules)
       Thread.new do
         begin
           while mods = mods_queue.pop(true) do
-            mods.each { |mod| mod.sync }
+            mods.each do |mod|
+              updated = mod.sync
+              updated_modules << mod.name if updated
+            end
           end
         rescue ThreadError => e
           logger.debug _("Module thread %{id} exiting: %{message}") % {message: e.message, id: Thread.current.object_id}
--- a/lib/r10k/git/stateful_repository.rb
+++ b/lib/r10k/git/stateful_repository.rb
@@ -35,6 +35,7 @@ class R10K::Git::StatefulRepository
     @cache.resolve(ref)
   end
 
+  # Returns true if the sync actually updated the repo, false otherwise
   def sync(ref, force=true)
     @cache.sync if sync_cache?(ref)
 
@@ -46,6 +47,7 @@ class R10K::Git::StatefulRepository
 
     workdir_status = status(ref)
 
+    updated = true
     case workdir_status
     when :absent
       logger.debug(_("Cloning %{repo_path} and checking out %{ref}") % {repo_path: @repo.path, ref: ref })
@@ -64,10 +66,13 @@ class R10K::Git::StatefulRepository
         @repo.checkout(sha, {:force => force})
       else
         logger.warn(_("Skipping %{repo_path} due to local modifications") % {repo_path: @repo.path})
+        updated = false
       end
     else
       logger.debug(_("%{repo_path} is already at Git ref %{ref}") % {repo_path: @repo.path, ref: ref })
+      updated = false
     end
+    updated
   end
 
   def status(ref)
--- a/lib/r10k/module/forge.rb
+++ b/lib/r10k/module/forge.rb
@@ -60,18 +60,24 @@ class R10K::Module::Forge < R10K::Module::Base
   end
 
   # @param [Hash] opts Deprecated
+  # @return [Boolean] true if the module was updated, false otherwise
   def sync(opts={})
+    updated = false
     if should_sync?
       case status
       when :absent
         install
+        updated = true
       when :outdated
         upgrade
+        updated = true
       when :mismatched
         reinstall
+        updated = true
       end
       maybe_delete_spec_dir
     end
+    updated
   end
 
   def properties
--- a/lib/r10k/module/svn.rb
+++ b/lib/r10k/module/svn.rb
@@ -74,18 +74,24 @@ class R10K::Module::SVN < R10K::Module::Base
   end
 
   # @param [Hash] opts Deprecated
+  # @return [Boolean] true if the module was updated, false otherwise
   def sync(opts={})
+    updated = false
     if should_sync?
       case status
       when :absent
         install
+        updated = true
       when :mismatched
         reinstall
+        updated = true
       when :outdated
         update
+        updated = true
       end
       maybe_delete_spec_dir
     end
+    updated
   end
 
   def exist?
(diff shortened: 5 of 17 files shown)


spec/unit/action/deploy/module_spec.rb
```diff
@@ -103,21 +103,14 @@
...
 
       before do
-        @modules = []
         allow(subject).to receive(:visit_environment).and_wrap_original do |original, environment, &block|
-          mod = R10K::Module::Local.new(environment.name, '/fakedir', {}, environment)
-          if mod.name == 'first'
-            expect(environment).to receive(:generate_types!)
+          if environment.name == 'first'
+            expect(environment).to receive(:deploy).and_return(['first'])
+            expect(environment).to recieve(:generate_types!)
```
Magisus: Adding this expectation causes some kind of error that makes rspec bail in such a way that `environment.deploy` never gets called. I can't figure out why. Wrapping these exceptions in a `begin...rescue` makes it go away, but also doesn't catch any exception.

It's the same with the `generate_types` expectation in the `else` case too.

Magisus: Typo, derp. Thanks Molly!

Magisus:
Tested this out, deploying a module to all environments, but only expecting it to be actually updated in one of them. Previously, this caused the following:
```
Running postrun command for environments prod1 prod2 production.
```
Now, it only runs for the one env where an update happened.
```
Running postrun command for environments prod1.
```

lib/r10k/action/deploy/module.rb
```diff
@@ -103,11 +105,10 @@ def visit_environment(environment)
           else
             logger.debug1(_("Updating modules %{modules} in environment %{env_path}") % {modules: @settings.dig(:overrides, :modules, :requested_modules).inspect, env_path: environment.path})
 
-            environment.deploy
+            updated_modules = environment.deploy
 
-            requested_mods = @settings.dig(:overrides, :modules, :requested_modules) || []
             # We actually synced a module in this env
-            if !((environment.modules.map(&:name) & requested_mods).empty?)
+            if !(updated_modules.nil? || updated_modules.empty?)
```
mwaggett: (nit) might be clearer to apply De Morgan's law here:
```suggestion
            if !updated_modules.nil? && !updated_modules.empty?
```

lib/r10k/module/svn.rb
```diff
@@ -75,17 +75,22 @@ def status
 
   # @param [Hash] opts Deprecated
   def sync(opts={})
+    updated = false
     if should_sync?
       case status
       when :absent
+        updated = true
         install
```
mwaggett: Might not matter at all, but is there a case where `install` could fail and we wouldn't want to set `updated = true`? i.e. should we set `updated` after calling `install` (or whatever method is needed for the given module status)?

lib/r10k/module/svn.rb
```diff
@@ -75,17 +75,22 @@ def status
 
   # @param [Hash] opts Deprecated
```
mwaggett, line 76: Could you add a yardoc for the `return` here, like you did for the other methods you updated?

lib/r10k/git/stateful_repository.rb
```diff
@@ -67,7 +69,9 @@ def sync(ref, force=true)
       end
```
mwaggett, line 70: Do we need to set `updated = false` for this `else` case within the `:dirty` case?

lib/r10k/content_synchronizer.rb
```diff
@@ -65,11 +74,14 @@ def self.enqueue_modules(queue, modules)
...
 
-    def self.sync_thread(mods_queue, logger)
+    def self.sync_thread(mods_queue, logger, updated_modules)
       Thread.new do
         begin
           while mods = mods_queue.pop(true) do
-            mods.each { |mod| mod.sync }
+            mods.each do |mod|
+              udpated = mod.sync
+              updated_modules << mod.name if udpated
```
mwaggett:
```suggestion
              updated = mod.sync
              updated_modules << mod.name if updated
```

lib/r10k/content_synchronizer.rb
```diff
@@ -8,31 +8,40 @@ def self.serial_accept(modules, visitor, loader)
...
+      thread_pool = pool_size.times.map { sync_thread(mods_queue, logger, updated_modules) }
       thread_exception = nil
 
       # If any threads raise an exception the deployment is considered a failure.
       # In that event clear the queue, wait for other threads to finish their
       # current work, then re-raise the first exception caught.
       begin
         thread_pool.each(&:join)
+        # Return the list of all modules that were actually updated
+        updated_modules
```
mwaggett, line 44: I'm confused about how `updated_modules` actually ends up with modules in it.. I see that `sync_thread` adds modules to the empty Queue we pass in, but it doesn't look like the updated Queue gets returned anywhere? so I'm not sure how `updated_modules` isn't still an empty Queue down here.. Is there some thread magic happening or something?

spec/unit/action/deploy/module_spec.rb
```diff
@@ -126,9 +119,8 @@
         expect(subject.settings[:overrides][:environments][:generate_types]).to eq(true)
       end
 
-      it 'only calls puppet generate types on environments with specified module' do
+      it 'only calls puppet generate types on environments where the specified module was updated' do
         subject.call
-        expect(@modules.length).to be(2)
       end
```
mwaggett: I'm not sure what the purpose of this test is, now that there are no `expect`s.. Should this just get squashed into a single test for 'with generate-types enabled', rather than putting the meat of the test into a `before` block?

spec/unit/action/deploy/module_spec.rb
```diff
@@ -422,16 +414,13 @@
...
-          mock_mod = double('mock_mod', name: 'mod1')
-
           expect(R10K::Util::Subprocess).to receive(:new).
             with(["/generate/types/wrapper", "first"]).
             and_return(mock_subprocess)
 
           expect(subject).to receive(:visit_environment).and_wrap_original do |original, environment, &block|
             if environment.name == 'first'
-              expect(environment).to receive(:deploy).and_return(true)
-              expect(environment).to receive(:modules).and_return([mock_mod])
```
mwaggett, line 434: Why remove this?

spec/unit/module/git_spec.rb
```diff
@@ -128,6 +128,25 @@
...
+    it 'returns false if repo was not updated (in-sync)' do
+      allow(mock_repo).to receive(:resolve).with('master').and_return('abc123')
+      expect(mock_repo).to receive(:sync).and_return(false)
+      expect(subject.sync).to be false
+    end
+
+    it 'returns false if `should_sync?` is false' do
+      # modules do not sync if they are not requested
+      mod = described_class.new(title, dirname, { overrides: { modules: { requested_modules: ['other_mod'] } } })
+      allow(mock_repo).to receive(:resolve).with('master').and_return('abc123')
```
mwaggett: Should this get pulled into a `before` block, since we use it in all the tests within this `describe`?

lib/r10k/content_synchronizer.rb
tvpartytonight: In the block for `mods.each` there is the line `updated_modules << mod.name if udpated`; that's where `updated_modules` gets modified in a new thread. Does that answer your question?

mwaggett: No. I'm not understanding how that modification of `updated_modules` persists outside of the `sync_thread` method, since `sync_thread` doesn't return the modified `updated_modules`, does it? All we do to `updated_modules` in this method is pass it as an argument to `sync_thread`, so I wouldn't expect its value to have changed by this point. But I probably just don't understand how threads work lol.

tvpartytonight: Ok, I think maybe I see what you are asking about? The following code I think demonstrates what you are thinking of:

```
[143] pry(main)> def coolmethod(a, c); Thread.new {; while b = a.pop {; c <<  b ; }; end; } end
=> :coolmethod
[144] pry(main)> a = [1,2,3]
=> [1, 2, 3]
[145] pry(main)> c = []
=> []
[146] pry(main)> coolmethod(a,c)
=> #<Thread:0x00007fbbdb0ba830 (pry):86 dead>
[147] pry(main)> a
=> []
[148] pry(main)> c
=> []
```
We can see that the `pop` method removes the elements from the array passed in, but the `<<` does not successfully add elements do the second array. I think that magisus was expecting that second array to be populated...

alexjfisher:
@Magisus Thanks for all your work on this!  I've tested this PR out today in my environment and everything seems to work as advertised.

If I create a branch of a module with a random name not matching any environment name, the postrun isn't called.
If I create a branch that *does* match an existing environment, then the postrun is called with just that single environment.
If I merge a branch into main, most environments are affected and postrun is called with that long list of environments (this is still pretty slow for me, but something that I'd expect.  It's not the usual case, and I can optimize my postrun script to cater for this.  For example, if postrun is called with > 10 environments, call the puppetserver API once to flush *all* environments instead of looping through the environments and flushing them individually.)

lib/r10k/content_synchronizer.rb
alexjfisher: on the other hand, this PR does appear to work properly during my testing...

mwaggett: huh, yeah, mine seems to have worked as Maggie intended:
```
irb(main):017:0> def thread_method(queue, empty_thing); Thread.new do; while element = queue.pop; empty_thing << element; end; end; end
=> :thread_method
irb(main):018:0> q = [1,2,3,4]
=> [1, 2, 3, 4]
irb(main):019:0> q
=> [1, 2, 3, 4]
irb(main):020:0> empty = []
=> []
irb(main):021:0> empty
=> []
irb(main):022:0> thread_method(q, empty)
=> #<Thread:0x00007fd19904c8c0@(irb):17 dead>
irb(main):023:0> q
=> []
irb(main):024:0> empty
=> [4, 3, 2, 1]
```

mwaggett:
This will also need a changelog entry.

lib/r10k/git/stateful_repository.rb
tvpartytonight: yeah, I think this is a missed case where updated should be false; magisus can change it back if there are considerations that are not apparent to either of us.

lib/r10k/module/svn.rb
```diff
@@ -74,18 +74,24 @@ def status
...
+  # @return [Boolean] true if the module was updated, false otherwise
   def sync(opts={})
+    updated = false
     if should_sync?
       case status
       when :absent
         install
+        updated = true
```
justinstoller, line 84: IMO, this would be clearer if we did the `updated = true` as the first line after `if should_sync?` and then added an `else` clause where we set `updated = false`. It would also then match what we're doing in StatefulRepository.

mwaggett: good call, I'll update that.

lib/r10k/content_synchronizer.rb
```diff
@@ -8,31 +8,40 @@ def self.serial_accept(modules, visitor, loader)
...
+      updated_modules
     end
 
+    # Returns a Queue of the names of modules actually updated
     def self.concurrent_accept(modules, visitor, loader, pool_size, logger)
       mods_queue = modules_visit_queue(modules, visitor, loader)
       sync_queue(mods_queue, pool_size, logger)
     end
 
+    # Returns a Queue of the names of modules actually updated
```
justinstoller, line 25: fwiw, I think this will return either a Queue of module names, or nil, if there was an error during a deprecated `accept` call, or raise if there was an error during the newer `sync`s.

mwaggett: Do you want me to update this? I didn't dig into these methods, but I can if you want.

lib/r10k/module/svn.rb
mwaggett: updated, let me know what you think of what I ended up with.

mwaggett: welp, looks like that doesn't work 😅 will fix

lib/r10k/module/forge.rb
```diff
@@ -71,7 +73,9 @@ def sync(opts={})
         reinstall
       end
       maybe_delete_spec_dir
+      updated = true
```
justinstoller: No, what I was thinking was something like:
```ruby
def sync(opts = {})
  updated = false
  if should_sync?
    updated = true
    case status
    when ...
    ....
    else
      updated = false
    end
    maybe_delete_spec_dir
  end
  updated
end
```

Maybe that's more confusing? I think as it's written now if the module should be synced, but `status` returns `:insync` then it will still be marked as updated??

lib/r10k/content_synchronizer.rb
justinstoller: That's fine, I was more trying to tie together why we were checking for both `nil?` and `empty?` in the caller

lib/r10k/module/svn.rb
mwaggett: not sure if I love the flipping back and forth that would happen if I set `updated = false` in an `else`, I might prefer what was there originally.. but I'll ruminate on that.

lib/r10k/module/forge.rb
mwaggett: yeah, I know that's what you meant. I thought what I did would be clearer, but didn't realize that there was an `:insync` status to worry about. But yeah, not sure if your thing is more or less confusing than the original.

mwaggett: okay, I think I like the explicitness of how it was originally, even though it's a little more repetitious.

mwaggett: I put the original back.

justinstoller (approved):

PR #1218, opened 2021-09-11
Bump the version of puppet_forge

### LINKED ISSUE #1217: The puppet_forge version should be bumped to 3.1.0+
## Describe the Bug
The puppet_forge dependency is causing issues updating gems that depend on `r10k`.

## Expected Behavior
The latest functional dependencies would be used.

## Additional Context
At this time, I have not seen any issues using `puppet_forge` 3.1.0+

trevor-vaughan:
Fixes #1217

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'
 
-  s.add_dependency 'puppet_forge', '~> 2.3.0'
+  s.add_dependency 'puppet_forge', ['>= 2.3.0', '< 4.0.0']
 
   s.add_dependency 'gettext-setup', '~>0.24'
   # These two pins narrow what is allowed by gettext-setup,


r10k.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'log4r',     '1.1.10'
   s.add_dependency 'multi_json', '~> 1.10'
 
-  s.add_dependency 'puppet_forge', '~> 2.3.0'
+  s.add_dependency 'puppet_forge', '~> 3.1.0'
```
mwaggett: I wonder if using `>= 2.3.0` would work? That way anyone still using Ruby 2.3 could keep using older versions of `puppet_forge`.

trevor-vaughan: Oh sure, that would work just fine. I can update that.

mwaggett (approved):
Confirmed that bundler will install an appropriate `puppet_forge` version based on the ruby version.

mwaggett:
ahhh apologies @trevor-vaughan but this doesn't actually work when attempting to `gem install` on Ruby 2.3 (as opposed to `bundle install`). 
```
prosaic-legacy:~ root# ruby --version
ruby 2.3.7p456 (2018-03-28 revision 63024) [universal.x86_64-darwin18]
prosaic-legacy:~ root# gem install r10k-3.11.999.gem
Fetching: jwt-2.2.3.gem (100%)
Successfully installed jwt-2.2.3
Fetching: text-1.3.1.gem (100%)
Successfully installed text-1.3.1
Fetching: locale-2.1.3.gem (100%)
Successfully installed locale-2.1.3
Fetching: gettext-3.2.9.gem (100%)
Successfully installed gettext-3.2.9
Fetching: fast_gettext-1.1.2.gem (100%)
Successfully installed fast_gettext-1.1.2
Fetching: gettext-setup-0.34.gem (100%)
Successfully installed gettext-setup-0.34
Fetching: faraday-em_http-1.0.0.gem (100%)
ERROR:  Error installing r10k-3.11.999.gem:
	faraday-em_http requires Ruby version >= 2.4.0.
```

trevor-vaughan:
Does anything Puppet supports actually use Ruby 2.3 any longer?

mwaggett:
@trevor-vaughan no, but dropping Ruby 2.3 support is still a breaking change for r10k, which we're unfortunately not prepared to make right now.

justinstoller:
I agree we shouldn't do this now, but we should probably commit soon to when we will do a 4.0 release.

We talked about doing one a few months ago but I don't think we felt there was enough pending to warrant one. Maybe we should review that decision when Maggie comes back from vacation next week?

mwaggett:
Definitely down to discuss when we want to do a 4.0, though I was under the impression that there was _more_ we wanted to put into a 4.0 than we felt we had time for (e.g. removing deprecated stuff, standardizing on some things, etc.). I have a doc with things we were hoping to do.

Magisus:
I think we should create a 4.x feature branch that we can start landing stuff like this on. We don't have to tag it till we've gotten the important things on our list, but if any eager user really wants to start working with the 4.x stuff, they can build a gem from source from that branch (or similar) to try it out.

PR #1224, opened 2021-09-17
(maint) Define `statically_defined_version` for local modules

mwaggett:
This commit adds a `statically_defined_version` method to the
`R10K::Module::Local` class. This method is called when instantiating a new
module definition, so previously, Puppetfiles that defined local modules could
not be evaluated because the method was missing.

### DIFF
--- a/lib/r10k/module/local.rb
+++ b/lib/r10k/module/local.rb
@@ -11,8 +11,12 @@ class R10K::Module::Local < R10K::Module::Base
     args.is_a?(Hash) && (args[:local] || args[:type].to_s == 'local')
   end
 
-  def version
+  def self.statically_defined_version(name, args)
     "0.0.0"
+  end
+
+  def version
+    self.class.statically_defined_version(self.name, nil)
   end
 
   def properties
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,8 @@ CHANGELOG
 Unreleased
 ----------
 
+- Fix evaluation of Puppetfiles that include local modules. [#1224](https://github.com/puppetlabs/r10k/pull/1224)
+
 3.12.0
 ------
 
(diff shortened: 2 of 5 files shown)


lib/r10k/module/local.rb
```diff
@@ -11,10 +11,14 @@ def self.implement?(name, args)
     args.is_a?(Hash) && (args[:local] || args[:type].to_s == 'local')
   end
 
-  def version
+  def self.statically_defined_version(name, args)
     "0.0.0"
   end
 
+  def version
+    self.statically_defined_version(self.name, nil)
```
justinstoller: shouldn't this be `self.class.statically_defined_version(self.name, nil)`

mwaggett: maybe/probably.. I've never been solid when to use `self` vs. not, etc. 😅

lib/r10k/module/local.rb
```diff
@@ -11,10 +11,14 @@ def self.implement?(name, args)
     args.is_a?(Hash) && (args[:local] || args[:type].to_s == 'local')
   end
 
-  def version
+  def self.statically_defined_version(name, args)
```
tvpartytonight, line 14: Since these args aren't used, should we drop them?

mwaggett: No, we call this with args, since the other module types use them in their implementations: https://github.com/puppetlabs/r10k/blob/main/lib/r10k/module/definition.rb#L11

justinstoller (approved):

lib/r10k/module/local.rb
tvpartytonight: In that case, it should be just `*`.

mwaggett: Do you want to make that edit?

PR #1227, opened 2021-09-24
Add R10K::ConfigError class

reidmv:
To use in raising general configuration errors. Discovered this not to
be a thing based on an attempt to raise an error in R10K::Source::Yaml.
Modified R10K::Source::Yaml to raise R10K::ConfigError now that it
exists.

```
```


Magisus (approved):
There are probably other places we could use this error type, but that can be left for later.

PR #1228, opened 2021-09-24
Align terminology with Puppet; bare => plain

reidmv:
Puppet already uses a term in the indirector system to indicate that an implementation doesn't really do anything, and that term is "plain". For the R10K environment type that doesn't do anything, we should use this same terminology. The initial choice of the term "bare" isn't ecosystem congruent, and has the language problem of being used meaningfully when describing Git repos, but that meaning doesn't translate at all to the environment type.

Using the term "plain" both aligns terminology with Puppet, and eliminates any potential language collisions with the Git meaning of the term.

I doubt this feature is much-used, and it's still documented as "experimental" so we could probably just change the name and be done. The PR includes a formal deprecation just in case anyone thinks it's needed pre-4.0.

```
```

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -760,16 +760,16 @@ modules:
     version: 62d07f2
 ```
 
-### Bare Environment Type
+### Plain Environment Type
 
 A "control repository" typically contains a hiera.yaml, an environment.conf, a manifests/site.pp file, and a few other things. However, none of these are strictly necessary for an environment to be functional if modules can be deployed to it.
 
-The bare environment type allows sources that support environment modules to operate without a control repo being required. Modules can be deployed directly.
+The plain environment type allows sources that support environment modules to operate without a control repo being required. Modules can be deployed directly.
 
 ```yaml
 ---
 production:
-  type: bare
+  type: plain
   modules:
     puppetlabs-stdlib:
       type: forge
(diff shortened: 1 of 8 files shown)


doc/dynamic-environments/configuration.mkd
```diff
@@ -764,12 +764,12 @@ modules:
```
mwaggett, line 764: ^Probably also want to change the title of this section to `Plain Environment Type`.

spec/unit/environment/plain_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
+require 'r10k/environment'
+
+describe R10K::Environment::Plain do
+  let(:basepath) { '/some/imaginary/path' }
+  let(:envname) { 'env_name' }
```
mwaggett: Doesn't look like these `let`s are needed, if you're just using string params below.

mwaggett (approved):
This generally looks good! Just a couple small nits, and this needs a rebase.

We're planning to do an r10k release today, so if you could make those tweaks ASAP, that would be great! Otherwise, if you don't mind (and perms allow), I can force-push to your branch to get this fixed up.

mwaggett:
also @reidmv if you wouldn't mind opening an RK ticket for removing the `bare` type altogether in 4.0.0, that would be much appreciated!

reidmv:
@mwaggett rebased, addressed nits, filed https://github.com/puppetlabs/r10k/issues/1235

mwaggett (approved):

PR #1231, opened 2021-10-08
[fixes #1230] Add support for specifying additional log outputs

### LINKED ISSUE #1230: Support logging to syslog/file
## Use Case

When triggering R10k deployments from remote scripts or pipelines, there's no supported method to locally see the output if something fails.

## Describe the Solution You Would Like

Adding support for specifying that R10k should output logging onto syslog or to a file either instead of _or_ in addition to stdout would allow for a local sysadmin to read through the logs from a triggered run.

## Describe Alternatives You've Considered

Creating local wrappers that pipe r10k output into syslog or file could be used, though it would require either owning every location where a call could occur from, or by replacing all r10k binaries with wrappers.

ananace:
This will add another configuration block where users can specify a list of additional log4r outputters to enable.

The example YAML has been expanded with some examples for common use-cases.

Fixes #1230

### DIFF
--- a/spec/fixtures/unit/action/r10k_logging.yaml
+++ b/spec/fixtures/unit/action/r10k_logging.yaml
@@ -1,0 +1,12 @@
+---
+  logging:
+    level: FATAL
+
+    outputs:
+      - type: file
+        parameters:
+          filename: r10k.log
+
+      - type: syslog
+
+    disable_default_stderr: true
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -8,6 +8,7 @@ Unreleased
 - (RK-381) Do not recurse into symlinked dirs when finding files to purge. [#1233](https://github.com/puppetlabs/r10k/pull/1233)
 - Purge should remove unmanaged directories, in addition to unmanaged files. [#1222](https://github.com/puppetlabs/r10k/pull/1222)
 - Rename experimental environment type "bare" to "plain". [#1228](https://github.com/puppetlabs/r10k/pull/1228)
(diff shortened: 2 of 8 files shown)


justinstoller (approved):
I've validated this on the command line, setting the default terminal log level, a different file log level, overriding the log level in the file on the command line, as well as disabling terminal output via the file. Looks good to me. Thanks so much!

spec/fixtures/unit/action/r10k_logging.yaml
```diff
@@ -0,0 +1,12 @@
+---
+  logging:
+    level: FATAL
+
+    outputs:
+      - type: file
+        parameters:
+          filename: /tmp/r10k.log
```
justinstoller: We should generate this file content and set this to a tempdir created by Ruby. I'll push a change to this PR to do so.

ananace: Honestly, for a test fixture where it's only used to check that configuration is instantiated correctly I'd rather just store it in a relative path, so the tests work for both Windows and Linux
I'm not really looking to test that log4r does what it's supposed to after all :)

justinstoller:
Can you rebase on to main and resolve the conflicts?

justinstoller:
Weird, I thought the tests passed before the rebse. Now, it doesn't look like the syslog output is available on windows. You may want to just gate those tests on a `R10K::Util::Platform.windows?` check.

ananace:
There we go, now the tests that would try to load the syslog outputter should be skipped on Windows.

justinstoller:
Thanks so much for your contribution Alexander! This should be out in a release shortly.

PR #1232, opened 2021-10-11
[WIP] Adds ability to resolve dependencies into a lockfile

binford2k:
This modifies the path resolution so that /path/to/Puppetfile.lock is
preferred. If the lockfile exists, then it will be used instead of the
Puppetfile. It also adds a `puppetfile resolve` action that will resolve
dependencies and write out that lockfile.

Note: all dependencies will be satisfied from the Forge, no matter what
the original source was.

Expected user workflow:

* Write Puppetfile that describes only the modules you intend to use.
* Run `r10k puppetfile resolve` to resolve all dependencies of those
  modules.
* Optional: review the dependent modules for quality and security
  purposes.
* Deploy.

@glennsarti what do you think about this implementation? It feels a bit
weird to have multiple Puppetfile parsers in the same tool, but don't
really see a better way. And am I using the dependency graph properly?

### DIFF
--- a/lib/r10k/action/puppetfile/resolve.rb
+++ b/lib/r10k/action/puppetfile/resolve.rb
@@ -1,0 +1,68 @@
+require 'r10k/action/base'
+require 'r10k/content_synchronizer'
+require 'r10k/errors/formatting'
+require 'r10k/module_loader/puppetfile'
+require 'r10k/util/cleaner'
+require 'puppetfile-resolver'
+require 'puppetfile-resolver/puppetfile/parser/r10k_eval'
+
+module R10K
+  module Action
+    module Puppetfile
+      class Resolve < R10K::Action::Base
+
+        def call
+          begin
+            @puppetfile ||= 'Puppetfile'
+            @lockfile     = "#{@puppetfile}.lock"
+
+            unless @force
+              logger.error "Pass --force to overwrite existing lockfile" if File.exist? @lockfile
+              return false
+            end
+
+            content    = File.read(@puppetfile)
+            puppetfile = PuppetfileResolver::Puppetfile::Parser::R10KEval.parse(content)
+
+            # Make sure the Puppetfile is valid
+            unless puppetfile.valid?
+              logger.error 'Puppetfile is not valid'
+              puppetfile.validation_errors.each { |err| logger.error err }
+              return false
+            end
+
+            resolver = PuppetfileResolver::Resolver.new(puppetfile, nil)
+            result   = resolver.resolve(strict_mode: true)
+
+            # Output resolution validation errors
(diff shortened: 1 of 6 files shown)


r10k.gemspec
```diff
@@ -38,6 +38,8 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'jwt', '~> 2.2.3'
 
+  s.add_dependendy 'puppetfile-resolver', '~0.5.0'
```
glennsarti: Typo!  :smile:

lib/r10k/action/puppetfile/resolve.rb
```diff
@@ -0,0 +1,68 @@
...
+            content    = File.read(@puppetfile)
+            puppetfile = PuppetfileResolver::Puppetfile::Parser::R10KEval.parse(content)
+
+            # Make sure the Puppetfile is valid
+            unless puppetfile.valid?
+              logger.error 'Puppetfile is not valid'
+              puppetfile.validation_errors.each { |err| logger.error err }
+              return false
+            end
+
```
glennsarti, line 33: Hi Ben!  @binford2k This is using it correctly.  If you want to avoid using my parser (which is fine btw.  I too feel _bad_ about have two parsers here), you can construct the Puppetfile directly.

You create a Document Model as per (https://sarti.dev/puppetfile-resolver/architecture). See `Creating a document model`

ekohl:
For reference, in https://github.com/theforeman/foreman-installer we have a repository with a Puppetfile which is resolved using librarian-puppet now. We have nightly packages from our `develop` branch. After we branch `develop` to `x.y-stable` we run `rake pin_modules` to rewrite the git URLs in Puppetfile (to pin them to y.x forge versions) and also add `Puppetfile.lock` to the repository. I would be interested to see if we could use r10k for this since librarian-puppet is on life-support.

It would probably be replacing https://github.com/theforeman/foreman-installer/blob/bef5bd620a5b1494d3b5779030bc6b29e861645c/Rakefile#L232 with some r10k equivalent. I opened https://github.com/theforeman/foreman-installer/pull/725 to see if that works. The first impression is that r10k is way too strict in its dependencies which makes it hell to integrate. I couldn't run it on Ruby 3 but CI also fails on Ruby 2.5.

bastelfreak (commented):
I'm not the Ruby expert you need to review and judge the code quality, but I think this is a good feature and should be within r10k (so librarian puppet can be deprecated)

rnelson0:
I can't review the ruby code effectively either but logically, it looks like the necessary components are there. I would like to see this enabled as a feature, either CLI or in r10k.yaml, rather than always-on, both because it's a new feature some may not want turned on right away or because they are using some other tool like librarian-puppet and will have to adjust their workflow.

PR #1233, opened 2021-10-12
(RK-381) Do not recurse into symlinked dirs when finding files to purge

Magisus:
Previously, if an environent contained a symlink pointing to another
location in the environment, r10k would purge the files pointed to, even
if they were being tracked by the environment. This occurred because we
would check whether such files were purgable twice: once at their real
location, and again via the symlink. The former would be flagged for
keeping, but the latter would be flagged for purging, because git does
not follow symlinks when listing tracked files. This means that the
non-symlinked path is considered desired (because Git "tracks" it), but
the symlink path is not, because Git does not report it as tracked.

This commit fixes this bug by not recursing into directories that are
also symlinks, meaning that the dir's contents will not show up in the
"purgeable" list. Note that symlinks pointing to files (not dirs) are
not impacted, because in this case the two paths are both explicitly
tracked by Git, since it tracks the symlink itself.

```
```


Magisus:
I think there's maybe one more test I need to add. And let me know if anyone else sees any other corner cases!

Magisus:
Added tests:
* symlink we want to keep, pointing to a file
* symlink we want to keep, pointing to a dir
* symlink we want to purge, pointing to a file
* symlink we want to purge, pointing to a dir

In all cases, the things pointed to are not purged. In the latter two, the symlinks themselves are purged, leaving the source files alone.

spec/unit/util/purgeable_spec.rb
```diff
@@ -30,7 +32,8 @@
 
     describe '#current_contents' do
       it 'collects direct contents of all managed directories' do
-        expect(subject.current_contents(recurse)).to contain_exactly(/\/expected_1/, /\/expected_2/, /\/unmanaged_1/, /\/unmanaged_2/, /\/managed_subdir_1/)
+        expect(subject.current_contents(recurse)).to contain_exactly(/\/expected_1/, /\/expected_2/, /\/unmanaged_1/, /\/unmanaged_2/,
+                                                                     /\/managed_subdir_1/, /\/managed_symlink_dir/, /\/unmanaged_symlink_file/)
```
mwaggett, line 36: I'm a little confused about your test fixtures.. I assume some are symlinks to other ones, but it's not totally clear to me what's what. I'm also not sure why you've called some of them `unmanaged`, but then expect those files to be in this collection of managed directories.

CHANGELOG.mkd
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
+- (RK-381) Do not purge files that are pointed to by a symlink. [#1233](https://github.com/puppetlabs/r10k/pull/1233)
```
mwaggett: Would it be clearer to say something like "Do not follow symlinks when purging files."? Or maybe "Do not purge symlink target files."? Just thinking "files that are pointed to by a symlink" sounds a little clunky. ¯\_(ツ)_/¯

spec/unit/util/purgeable_spec.rb
Magisus: This test is for _current_ contents, not managed contents. As in, what is everything that is in the dir before purging happens? The class then filters this list against `desired_contents`, and removes anything that is leftover.

The "managed" ones are listed in "desired contents", which is what the `Purgeable` class considers managed: the entity (e.g. the environment) is tracking those files, so they shouldn't be purged. Unmanaged ones are not listed there, so they show up as "stale" when collecting files to purge.

Without my change, the tests for `stale_contents` return all kinds of fun, nasty stuff:
```
["spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/unmanaged_symlink_dir/.hidden/unmanaged_3",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/unmanaged_symlink_dir/.hidden",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/unmanaged_symlink_dir/expected_2",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/unmanaged_symlink_dir/unmanaged_2",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/unmanaged_symlink_dir",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/subdir_allowlisted_2/ignored_1",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/subdir_allowlisted_2",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/managed_symlink_file",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/subdir_unmanaged_1",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir/subdir_expected_1",
 "spec/fixtures/unit/util/purgeable/managed_one/unmanaged_1",
 "spec/fixtures/unit/util/purgeable/managed_one/unmanaged_symlink_file",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/unmanaged_symlink_dir/.hidden/unmanaged_3",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/unmanaged_symlink_dir/.hidden",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/unmanaged_symlink_dir/expected_2",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/unmanaged_symlink_dir/unmanaged_2",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/unmanaged_symlink_dir",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_allowlisted_2/ignored_1",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_allowlisted_2",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_unmanaged_1",
 "spec/fixtures/unit/util/purgeable/managed_two/unmanaged_2"]
```
You can see, for example, that there are two separate erroneous ways that `.hidden/unmanaged_3` would get purged with this bug, when in fact it shouldn't be in the list at all. Whether the symlink itself is supposed to be purged does not affect whether the source files get deleted: both the managed and unmanaged symlinks cause a path that leads to those files to be added to the "stale" contents, which [`FileUtils.rm_r`](https://github.com/puppetlabs/r10k/blob/3f2b97a5d6955809385a34ccce7f065504e04e93/lib/r10k/util/purgeable.rb#L141) will happily delete.

With my change, the list is correct:
```
["spec/fixtures/unit/util/purgeable/managed_one/unmanaged_1",
 "spec/fixtures/unit/util/purgeable/managed_one/unmanaged_symlink_file",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/unmanaged_symlink_dir",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_allowlisted_2/ignored_1",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_allowlisted_2",
 "spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_unmanaged_1",
 "spec/fixtures/unit/util/purgeable/managed_two/unmanaged_2"]
```
The unmanaged symlinks (the ones not in `desired_contents`) are added for purging, but the files they point to are not. I found in my manual tests that in this case, `FileUtils.rm_r` will not follow the link, but just delete the link itself. I think I would like to add an integration level test proving this.

Magisus: Note that this is a bug specifically for git-based environments, because Git does not follow symlinks when listing files it tracks: it tracks the symlink itself, but not what it points to, so the `desired_contents` that an environment passes to this class will contain the link but not the path to files that includes the link. So by calculating our contents differently than Git does, we introduced some overzealous purging by accident.

I don't know if other types of environments, which have a different way of collecting their desired contents, would behave the same way, but this is safe in those cases too. It's better to not purge something, and maybe leave around some cruft, than purge too much and risk nuking something a user needs (e.g. what if someone had a bad relative symlink that wound up pointing _outside_ the environment? Bad times).

spec/unit/util/purgeable_spec.rb
```diff
@@ -15,8 +15,10 @@
       'spec/fixtures/unit/util/purgeable/managed_one/expected_1',
       'spec/fixtures/unit/util/purgeable/managed_one/new_1',
       'spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1',
+      'spec/fixtures/unit/util/purgeable/managed_one/managed_symlink_dir',
       'spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_expected_1',
       'spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/subdir_new_1',
+      'spec/fixtures/unit/util/purgeable/managed_one/managed_subdir_1/managed_symlink_file',
```
Magisus, line 21: Only "managed" symlinks appear in this list of desired contents, which is what makes them "managed".

CHANGELOG.mkd
Magisus: Yeah I can rephrase. 

I did it like this because I find the terminology around symlinks to be very confusing: people avoid the word "target" and call it the "source" file instead, which has always seemed a little backwards to me, because I think we have a tendency to think of an arrow with a source (the location of the link) and a target (the file pointed to), but that's the exact opposite of how Unix talks about them.

And I didn't say "follow symlinks when purging" because that's not really what was happening (and I spent way too long thinking that it was). We previously we were recursing into symlinked directories _explicitly_ when figuring out which files to purge. But that subtlety probably isn't important here.

spec/unit/util/purgeable_spec.rb
mwaggett: mmm okay, so the fixtures you've added are _all_ symlinks and what's showing up as the file content is actually what they're pointing to? 

I think that makes sense.

CHANGELOG.mkd
mwaggett: heh yeah, I do know what you mean about symlink terminology being pretty confusing. I guess in my mind "source" and "target" both mean the real file, as opposed to the symlink pointing to it haha. I can definitely see how "target" could mean either, but do you think "source" is clear?

mwaggett: You could also just copy the title of this PR ¯\_(ツ)_/¯

spec/unit/util/purgeable_spec.rb
Magisus: Yeah I added four symlinks, two pointing to dirs and two pointing to files, one managed and one unmanaged each. They're all relative, and they all point to fixture files that previously existed and are considered "desired". In general, these tests verify that "duplicate" paths to these desired files do not show up in the list of things to purge, and on the flip side, that "undesired" symlinks _do_ show up in the "stale" list, even after my fix.

I'm working on another test right now that actually runs `purge!` on a copy of this fixture set, to prove that `rm_r` doesn't clean up anything it shouldn't when purging the unmanaged symlinks themselves.

CHANGELOG.mkd
Magisus: Haha good point.

spec/unit/util/purgeable_spec.rb
mwaggett: okay cool, that seems like a good test to add. otherwise I think this looks good!

Magisus: Added! See above.

spec/integration/util/purageable_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      cleaner = R10K::Util::Cleaner.new([managed_directory], desired_contents)
+
+      cleaner.purge!({ recurse: true, whitelist: ["**/subdir_allowlisted_2"] })
+
+      # Files present after purge
+      expect(File.exist?("#{managed_directory}/expected_1")).to be true
+      expect(File.exist?("#{managed_directory}/managed_subdir_1")).to be true
+      expect(File.exist?("#{managed_directory}/managed_symlink_dir")).to be true
+      expect(File.exist?("#{managed_directory}/managed_subdir_1/subdir_expected_1")).to be true
+      expect(File.exist?("#{managed_directory}/managed_subdir_1/managed_symlink_file")).to be true
```
Magisus, line 31: When run without this change, this line and the one above it fail, because `managed_subdir_1` got cleaned out when `managed_synlink_dir` was recursed into, and the mismatched path names added to `stale_contents`. This reproduces what the user saw in the original escalation.

mwaggett, line 31: might be nice to just do
```suggestion
      desired_contents.each do |path|
        expect(File.exist?(path)).to be true
      end
```

mwaggett (approved):

spec/integration/util/purageable_spec.rb
Magisus: I had that first and the problem is that if it fails, all you get is:
```
Failure/Error: expect(File.exist?(path)).to be true

expected true
got false
```
Which is utterly useless.

mwaggett: ugh that's so annoying

Issue #1221, opened 2021-09-16
docker images are no longer built


skoef:
## Describe the Bug
Since about [6 months](/puppetlabs/r10k/actions/workflows/docker.yml) new docker images are no longer built automatically. The latest tag on docker hub is [3.8.0](https://hub.docker.com/r/puppet/r10k/tags).

## Expected Behavior
I would expect that on each release, a new docker container would be created

## Additional Context
Obviously I could build my own docker images for later releases but I'd rather just use the official images.

Magisus:
We are currently discussing the future of these containers internally, hope to have an update for you soon.

Out of curiosity, and to help inform our path forward, how are you using this container? Are you using it by itself, or in combination with other open-source containers we publish (e.g. puppetserver)?

justinstoller:
We've finished that discussion and we will get to fixing this issue in the near future, though we don't currently have an ETA.

skoef:
@Magisus I completely forgot to follow up, my apologies! I'm using the r10k container in the [puppetserver-helm-chart](https://github.com/puppetlabs/puppetserver-helm-chart)!

Magisus:
Oh, the master->main switch broke this. PR: https://github.com/puppetlabs/r10k/pull/1237 Thanks for pointing this out!

Magisus:
A new container has been published! Let us know if you have any more issues.

PR #1237, opened 2021-11-01
[#1221] Build docker images from main branch instead of master

Magisus:
We renamed the master branch to main, and that broke the GH Action that
build docker images. This commit updates that action to track the main
branch instead.


tvpartytonight (approved):
LGTM

PR #1240, opened 2021-11-03
use correct reference when using strip_component

b4ldr:
The current implmnetation sets the environment name to the "striped"
version of the name, this is then later used as the git reference when
syncing environments.

With the following configuration
```yaml
sources:
  b4ldr:
    strip_component: "wmcs/"
    ignore_branch_prefixes:
    - production
    - ops
    - sandbox
    - pontoon
    remote: https://github.com/b4ldr/puppet-1
```

where upstream has a branch with wmcs/production then we see the
following error

  Unable to sync repo to unresolvable ref 'production' at
    /etc/puppet/code/environments/production/.git

If upstream also has a production branch then the above configuration
ends up syncing  /etc/puppet/code/environments/production/ with the
remote production branch.  even thought it should be striped via
`ignore_branch_prefixes`

This PR updates the name object to keep hold of the original name and
store it as the ref


Magisus:
This fix makes sense. Could you add a test, maybe in [here](https://github.com/puppetlabs/r10k/blob/3e059e8b74c4e63e49e831f2cdda1eccb50a6632/spec/unit/environment/name_spec.rb#L5), showing that your new `ref` field contains the unmodified name (while `name` still gets the derived version)?

b4ldr:
Sorry for the delay, have updated with additional rspec test now

Magisus:
So I was testing this out, and I think you're missing a line in the sample config you give. It should also have `strip_component: "wmcs/"`, right? If I add that and deploy your control repo, I see the issue you're seeing. Is that correct?

Magisus (approved):
Other than that small but important update to the commit message, I think this looks good! It definitely fixes the issue.

lib/r10k/environment/name.rb
```diff
@@ -7,7 +7,7 @@ class Name
 
       # @!attribute [r] name
       #   @return [String] The unmodified name of the environment
-      attr_reader :name
+      attr_reader :name, :ref
```
reidmv: Possibly a nit: this new attribute, `:ref`, is not documented. I think it needs to be. Without understanding exactly what it is for, and what the difference between it and name is, it would probably be a little bit confusing to look at later and try to figure out what was going on and why.

b4ldr:
> So I was testing this out, and I think you're missing a line in the sample config you give. It should also have `strip_component: "wmcs/"`, right? If I add that and deploy your control repo, I see the issue you're seeing. Is that correct?

yes, updated

lib/r10k/environment/name.rb
b4ldr: have added something now

lib/r10k/environment/name.rb
```diff
@@ -5,9 +5,12 @@ module Environment
...
     class Name
 
+      # @!attribute [r] ref
+      #   @return [String] The unmodified reference of the upstream repository
       # @!attribute [r] name
-      #   @return [String] The unmodified name of the environment
-      attr_reader :name
+      #   @return [String] The name of the environment (with component striped)
+      attr_reader :name, :ref
+      attr_reader :name, :ref
```
mwaggett: This doesn't need to be here twice.

lib/r10k/environment/name.rb
```diff
@@ -5,9 +5,12 @@ module Environment
     # @api private
     class Name
 
+      # @!attribute [r] ref
+      #   @return [String] The unmodified reference of the upstream repository
       # @!attribute [r] name
-      #   @return [String] The unmodified name of the environment
-      attr_reader :name
+      #   @return [String] The name of the environment (with component striped)
```
mwaggett: What does 'component' refer to?

mwaggett (commented):
Would you mind adding an entry to the 'Unreleased' section of the CHANGELOG describing this fix?

spec/unit/environment/name_spec.rb
```diff
@@ -149,6 +165,8 @@
         it "replaces invalid characters in #{branch} with underscores" do
           bn = described_class.new(branch.dup, {:correct => true})
           expect(bn.dirname).to eq branch.gsub(/\W/, '_')
+          expect(bn.name).to eq branch
```
mwaggett, line 168: huh, I would have thought this would also get corrected, but I guess I'm not totally sure where all the `name` is used.

reidmv:
As I'm thinking about this I have some thoughts on how to clarify what's going on here, without mixing in an assumption in `R10K::Environment::Name` that a VCS is involved (since not all environments come from version control systems, or have "references"). I will push some suggestions in just a bit.

reidmv:
@b4ldr I've suggested some generalizations in 6c95162. The biggest changes:

* Renaming `R10K::Environment::Name#ref` to `R10K::Environment::Name#original_name`
* Updating the documentation for the old and new `R10K::Environment::Name` attributes
 
The rest of the changes in the commit are just renaming some variables in `R10K::Source::Git` to better differentiate between branch names vs. environment names, since they can be different.

Unless you have any objections or feedback we'll keep the process moving forward. 🙂

spec/unit/environment/name_spec.rb
reidmv: Yeah. After taking a look through it, I realized that it's currently inconsistent. The original implementation of `strip_component` introduced the inconsistency. I'll take responsibility for that. There would ideally be a clear separation between the input value given to `R10K::Environment::Name` and the derived code-environment name on disk (currently called `#dirname`). Bit of work to clean it all up at this point.

reidmv (approved):
I have a commit in here myself now so somebody should sign off on that, but I'm 👍  overall

Magisus (approved):
Thanks for weighing in here, Reid, I agree, this is better.

Magisus:
Dang, was really hoping git would be able to resolve that without a conflict :P

spec/unit/environment/name_spec.rb
mwaggett: ahh. Do you think it's a worthwhile bit of work to do eventually? If so, would you mind opening a ticket for it?

mwaggett (approved):

mwaggett:
Let us know what you think of these changes, @b4ldr ! And if you wouldn't mind rebasing to resolve the merge conflicts in the CHANGELOG, that would be appreciated (we didn't want to force-push to your branch)!

b4ldr:
> Let us know what you think of these changes, @b4ldr ! And if you wouldn't mind rebasing to resolve the merge conflicts in the CHANGELOG, that would be appreciated (we didn't want to force-push to your branch)!

changes all good with me, think i have rebased and unblocked but let me know if not

Magisus:
Looks good, we'll get this in!

mwaggett (approved):

PR #1241, opened 2021-11-05
(RK-378) Restore access to the environment name from the Puppetfile

jarretlavallee:
This PR restores the `basedir` getter in the Puppetfile DSL with a deprecation warning and adds a new 'environment_name' getter that can be used to get the name of the environment.


Magisus (approved):
Thanks for doing this, Jarret! I'll open a ticket for removing the old version and add it to our r10k 4 epic.

Magisus:
Filed as https://tickets.puppetlabs.com/browse/RK-383

PR #1242, opened 2021-11-08
Implement R10K::Util::Cacheable

reidmv:

The purpose of `R10K::Util::Cacheable` is to centrally define values and functions used by objects that implement caches. We already have two such objects in the codebase (`R10K::Git::Cache`, and `R10K::Forge::ModuleRelease`). Adding Tarball-based types would introduce a third.

Centralizing in adherence with the DRY principle.

### DIFF
--- a/lib/r10k/util/cacheable.rb
+++ b/lib/r10k/util/cacheable.rb
@@ -1,0 +1,31 @@
+module R10K
+  module Util
+
+    # Utility mixin for classes that need to implement caches
+    #
+    # @abstract Classes using this mixin need to implement {#managed_directory} and
+    #   {#desired_contents}
+    module Cacheable
+
+      # Provide a default cachedir location. This is consumed by R10K::Settings
+      # for appropriate global default values.
+      #
+      # @return [String] Path to the default cache directory
+      def self.default_cachedir(basename = 'cache')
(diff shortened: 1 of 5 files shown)


lib/r10k/util/cacheable.rb
```diff
@@ -0,0 +1,31 @@
...
+    #
+    # @abstract Classes using this mixin need to implement {#managed_directory} and
+    #   {#desired_contents}
+    module Cacheable
+
+      # Provide a default cachedir location. This is consumed by R10K::Settings
+      # for appropriate global default values.
+      #
+      # @return [String] Path to the default cache directory
+      def self.default_cachedir(basename = 'cache')
```
Magisus, line 14: Is `cache` what you're planning to use for the third usage of this util, too?

Magisus (approved):

lib/r10k/util/cacheable.rb
reidmv: Yeah. The only reason this is parameterized here is to allow Git's behavior to remain unchanged, and not invalidate caches on upgrade.

I would say there's an argument for removing the argument in a major release, and making all cacheable classes query for the same default cachedir value without parameterization. They can create their own subdirs from there if they want.

PR #1243, opened 2021-11-08
Implmement R10K::Util::Downloader

reidmv:

This PR implements a mixin for R10K objects which need to download, checksum, and save/copy files. The mixin isn't used by anything yet in this PR. A subsequent PR will introduce a consumer of the mixin, the Tarball object.

### DIFF
--- a/lib/r10k/util/downloader.rb
+++ b/lib/r10k/util/downloader.rb
@@ -1,0 +1,134 @@
+require 'digest'
+require 'net/http'
+
+module R10K
+  module Util
+
+    # Utility mixin for classes that need to download files
+    module Downloader
+
+      # Downloader objects need to checksum downloaded or saved content. The
+      # algorithm used to perform this checksumming (and therefore the kinds of
+      # checksums returned by various methods) is reported by this method.
+      #
+      # @return [Symbol] The checksum algorithm the downloader uses
+      def checksum_algorithm
+        @checksum_algorithm ||= :SHA256
+      end
+
+      private
+
+      # Set the checksum algorithm the downloader should use. It should be a
+      # symbol, and a valid Ruby 'digest' library algorithm. The default is
+      # :SHA256.
+      #
+      # @param algorithm [Symbol] The checksum algorithm the downloader should use
+      def checksum_algorithm=(algorithm)
+        @checksum_algorithm = algorithm
+      end
+
+      CHUNK_SIZE = 64 * 1024 # 64 kb
+
+      # @param src_uri [URI] The URI to download from
+      # @param dst_file [String] The file or path to save to
+      # @return [String] The downloaded file's hex digest
+      def download(src_uri, dst_file)
+        digest = Digest(checksum_algorithm).new
+        http_get(src_uri) do |resp|
+          File.open(dst_file, 'wb') do |output_stream|
+            resp.read_body do |chunk|
+              output_stream.write(chunk)
+              digest.update(chunk)
+            end
+          end
+        end
+
+        digest.hexdigest
+      end
+
+      # @param src_file The file or path to copy from
+      # @param dst_file The file or path to copy to
+      # @return [String] The copied file's sha256 hex digest
+      def copy(src_file, dst_file)
+        digest = Digest(checksum_algorithm).new
+        File.open(src_file, 'rb') do |input_stream|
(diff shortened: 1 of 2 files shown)


lib/r10k/util/downloader.rb
```diff
@@ -0,0 +1,135 @@
...
+      # @param algorithm [Symbol] The checksum algorithm the downloader should use
+      def checksum_algorithm=(algorithm)
+        @checksum_algorithm = algorithm
+      end
+
+      CHUNK_SIZE = 64 * 1024 # 64 kb
+
+      # @param src_uri [URI] The URI to download from
+      # @param dst_file [String] The file or path to save to
+      # @param checksum_algorithm [Symbol] The Digest algorithm to use
```
Magisus: Looks like this is no longer a param

lib/r10k/util/downloader.rb
```diff
@@ -0,0 +1,135 @@
...
+      end
+
+      # Start a Net::HTTP::Get connection, then yield the Net::HTTPSuccess object
+      # to the caller's block. Follow redirects if Net::HTTPRedirection responses
+      # are encountered, and use a proxy if directed.
+      #
+      # @param uri [URI] The URI to download the file from
+      # @param redirect_limit [Integer] How many redirects to permit before failing
+      # @param proxy [URI, String] The URI to use as a proxy
+      def http_get(uri, redirect_limit: 10, proxy: nil, &block)
```
Magisus, line 74: I'm actually surprised we don't already have anything like this, but I guess it makes sense. All of our current downloading goes through other libs (forge-ruby/faraday, git/rugged).

lib/r10k/util/downloader.rb
```diff
@@ -0,0 +1,135 @@
...
+            end
+          end
+        end
+
+        digest.hexdigest
+      end
+
+      # @param input The file or path to copy from
+      # @param output The file or path to copy to
+      # @return [String] The copied file's sha256 hex digest
```
Magisus, line 51: These have different names now.

Magisus (commented):
This seems alright for most use cases. If anything else comes up that we need to add (e.g. timeouts), we can do that separately I think.

Magisus:
Took a look at the way this is used in the subsequent PR, and the API seems fine to me (fairly idiomatic anyway).

lib/r10k/util/downloader.rb
reidmv: Removed yard tag

lib/r10k/util/downloader.rb
reidmv: fixed yard tag argument names

lib/r10k/util/downloader.rb
reidmv: Yeah. As a point of interest I have a Faraday version of this in another branch but at the time it was written R10K couldn't use a new enough version of Faraday to support streaming downloads, which I would prefer to do for checksumming, and testing streaming downloads with Faraday seemed hard.

Digging up the faraday code, it looked something like this in comparison:

```ruby
# @param uri [URI] The URI to download from
# @param output The file or path to save to
# @return [String] The downloaded file's sha256 hex digest
def download_to_file(uri, output)
  digest = DIGEST_CLASS.new

  File.open(output, 'wb') do |output_stream|
    faraday_connection.get(uri) do |req|
      req.options.on_data = Proc.new do |chunk, _|
        output_stream.write(chunk)
        digest.update(chunk)
      end
    end
  end

  digest.hexdigest
end

# Defined separately to enable test stubbing
def faraday_connection
  Faraday.new do |faraday|
    faraday.response(:follow_redirects)
    faraday.adapter(:net_http)
    faraday.proxy(settings[:proxy]) unless settings[:proxy].nil?
  end
end
```

I honestly have no idea if it worked as expected, as I said testing was hard.

The alternative to streaming downloading would be to download the file first, then checksum it separately.

Since I already had the non-Faraday code and it seemed to work well, I didn't pursue it further.

Magisus: Yeah honestly I'm totally fine not depending more directly on faraday. There are way too many layers over in that family of libraries. We're using Net::HTTP directly in Puppet, in a similar (but more elaborate) way to what you're doing here.

Magisus (approved):

PR #1244, opened 2021-11-08
Tarball

### LINKED ISSUE #567: Support for plain tarballs

reidmv:

This pull request **DEPENDS ON** #1242, #1243. Do not merge until after those PRs are merged.

This PR implements:

* An R10K::Tarball base class
* An R10K::Module::Tarball class, providing a "tarball" module type
* An R10K::Environment::Tarball class, providing a "tarball" environment type

Example usage:

```
---
production:
  type: tarball
  source: https://example.com/puppet-code-env-v1.0.0.tar.gz
  version: 99a906c99c2f144de43f2ae500509a7474ed11c583fb623efa8e5b377a3157f0 # sha256digest
  modules:
    puppetlabs-peadm:
      type: forge
      version: 2.5.0
    reidmv-xampl:
      type: git
      version: 62d07f2
    reidmv-mymodule:
      type: tarball
      source: https://example.com/reidmv-mymodule-1.0.0.tar.gz
      version: 6128ada158622cd90f8e1360fb7c2c3830a812d1ec26ddf0db7eb16d61b7293f # sha256digest
```

Later enhancements built on top of this might include Artifactory types, or S3 types, with options for real versions, depending on how the remote repository works.

Fixes #567
Closes #1159


lib/r10k/module/tarball.rb
```diff
@@ -0,0 +1,95 @@
...
+      :absent
+    elsif not (tarball.cache_valid? && tarball.insync?(path.to_s))
+      :mismatched
+    else
+      :insync
+    end
+  end
+
+  # Synchronize this module with the indicated state.
+  def sync(opts={})
```
Magisus, line 59: Can you make this return whether it was updated or not? Like this does: https://github.com/puppetlabs/r10k/blob/c5628d2063a8a936f083f2e4c724205e4002de4a/lib/r10k/module/git.rb#L100-L111 We use that info to determine whether to run post-deploy hooks on a `deploy module` command.

lib/r10k/module/tarball.rb
```diff
@@ -0,0 +1,95 @@
...
+  def properties
+    {
+      :expected => @checksum,
+      :actual   => status,
+      :type     => :tarball,
+    }
+  end
+
+  # Return the module's cachedir. Subclasses that implement a cache
+  # will override this to return a real directory location.
```
Magisus: I think this comment got copied from `Module::Base` and isn't accurate here.

spec/unit/module/tarball_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  subject do
+    described_class.new(
+      'fixture-tarball',
+      moduledir,
+      base_params,
+    )
+  end
+
+  describe "setting the owner and name" do
+    describe "with a title of 'mock-tarball'" do
```
Magisus: I don't see this title coming from anywhere, does this `describe` need updating?

spec/unit/module/tarball_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    it 'defaults to keeping the spec dir' do
+      subject.sync
+      expect(Dir.exist?(File.join(moduledir, 'tarball', 'spec'))).to be(true)
+    end
+  end
+
+  describe "determining the status" do
+    it "delegates to R10K::Tarball" do
+      mock_tarball = instance_double('R10K::Tarball', cache_valid?: true, insync?: true)
+      expect(subject).to receive(:tarball).twice.and_return instance_double('R10K::Tarball', cache_valid?: true, insync?: true)
```
Magisus, line 53: Did you mean to use `mock_tarball` here as the thing to return?

spec/unit/module/tarball_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      expect(subject.status).to eq(:insync)
+    end
+  end
+
+  describe "option parsing" do
+    describe "version" do
+      context "when no version is given" do
+        subject { described_class.new('fixture-tarball', moduledir, base_params.reject { |k| k.eql?(:version) }) }
+        it "does not require a version" do
```
Magisus, line 64: So, from a user's perspective, what does this do? I assume since these tarballs are downloaded from somewhere specific, they still would get a fixed version. Does this just allow them to not specify the checksum explicitly?

lib/r10k/environment/tarball.rb
```diff
@@ -0,0 +1,78 @@
...
+    @path ||= Pathname.new(File.join(@basedir, @dirname))
+  end
+
+  def sync
+    tarball.get unless tarball.cache_valid?
+    case status
+    when :absent, :mismatched
+      tarball.unpack(path.to_s)
+      # TODO: Remove untracked files. It's possible r10k's purge facility will
+      #       kinda sorta deal with this, but we shouldn't rely on it.
```
Magisus: Given what you've done with `desired_contents` below, I think r10k's purging will take care of this, assuming environment-level purging is enabled (which it is in PE). Are you concerned about the case where it's disabled? Or was there some other edge case you had in mind?

Magisus (commented):
I think I want to play with this a bit.

Magisus:
Would you mind adding something to the docs with examples on how to use this?

lib/r10k/module/tarball.rb
reidmv: Adjusted return value in c95c9c9.

lib/r10k/module/tarball.rb
reidmv: Totally. Fixed in c95c9c9.

spec/unit/module/tarball_spec.rb
reidmv: Indeed it does. Fixed in c95c9c9.

spec/unit/module/tarball_spec.rb
reidmv: Either that or don't declare a variable for this. I went that route and deleted the line in c95c9c9.

spec/unit/module/tarball_spec.rb
reidmv: I added a mention of that to the docs in bc24f6b. Basically, if the user doesn't specify a checksum, the cache name will be based on the source URL and won't be invalidated until/unless the source URL changes, or a version string is supplied. With no version, whatever gets downloaded and cached is always considered valid and correct.

lib/r10k/environment/tarball.rb
reidmv: I didn't have any specific edge cases in mind, it just felt unclean to leave it. But you're right, this is basically what r10k's purging is for. Updated comment to indicate recognition that old files could be left behind, and that we are relying on purging to come through and clean up after.

lib/r10k/environment/tarball.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  include R10K::Util::Purgeable
+
+  # Returns an array of the full paths to all the content being managed.
+  # @note This implements a required method for the Purgeable mixin
+  # @return [Array<String>]
+  def desired_contents
+    desired = []
+    desired += @tarball.paths.map { |entry| File.join(@full_path, entry) }
```
Magisus, line 75: Discovered an interesting edge case here: if you build the tarball a bit wrong, even though it unpacks correctly, it might purge incorrectly. In my case, I created the tarball with `.`. So the paths inside it look like this:
```
"/Users/aileen/code/r10k/yaml-envs/tarball/.",
 "/Users/aileen/code/r10k/yaml-envs/tarball/./site-modules",
 "/Users/aileen/code/r10k/yaml-envs/tarball/./hiera.yaml",
...
```
But r10k's `current_contents` list looks like this:
```
"/Users/aileen/code/r10k/yaml-envs/tarball/site-modules",
 "/Users/aileen/code/r10k/yaml-envs/tarball/hiera.yaml",
...
```
So it purges everything except the `modules` directory (which is ignored by the `environment` purge level).

I realize this was my mistake, and I should get better at tarballs 😛 But it seems like something that could trip people up and cause a very confusing bug. Maybe we can add some normalization in `Tarball.paths`? Adding `Pathname.new(path).cleanpath` seemed to do the trick for me.

reidmv: Ha! Excellent catch. Although I do have unbridled confidence that users of software would NEVER imperfectly construct a tarball 😉, this does seem like something we could fix.

reidmv: Huh. I just snap-pushed what I thought would be the fix in 5a1711e, and then realized I was looking at a completely different part of the code. But, it might actually be a valid fix anyway? Don't have a test for it yet.

Magisus: That's the spot I edited when fixing this locally 🤷‍♀️ Seemed to work for my bad tarball.

Magisus: Were you working on a test for this? I think besides that, I'm happy to merge this.

Magisus:
So, are the tarball modules usable from the Puppetfile too? Or just with yaml envs? If they are usable from a Puppetfile, it might be nice to note that [here](https://github.com/puppetlabs/r10k/blob/main/doc/puppetfile.mkd).

lib/r10k/environment/tarball.rb
reidmv: Yeah, lemme add some thing for that. I might just re-construct the fixture tarball to include the `./` name components, I think that'd make it easy to validate. Will push something shortly.

reidmv:
> So, are the tarball modules usable from the Puppetfile too? Or just with yaml envs? If they are usable from a Puppetfile, it might be nice to note that here.

Totally, and yeah, they are usable. I believe bc24f6bb29a84819e39f81b5756c77a43d0db797 contains some reference to that. Do you think it should say more or say it differently?

Magisus:
Oh duh, no, I missed which file that got added to, my bad.

lib/r10k/environment/tarball.rb
Magisus: Yeah that's what I was thinking 🙂

reidmv:
Two logical changes:

* Changed the fixture'd tarball to include `./` path prefixes, such as a user might incidentally include in a constructed tarball. Did this in a rebase so as not to leave an extra binary blob in the Git history.
* Made a new commit that adjusts R10K::Tarball#paths to return clean paths only, with a test

I didn't go so far as adding a purge test specifically. Seems like the most thorough fix would actually be to R10K::Util::Purgable itself, if we really wanted to do that. It seems reasonable enough for Tarball to just make sure paths returned are clean paths.

Magisus (approved):
Retested this, seems to work well.

lib/r10k/module/tarball.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  # Return the properties of the module
+  #
+  # @return [Hash]
+  # @abstract
+  def properties
+    {
+      :expected => @checksum,
+      :actual   => status,
```
justinstoller: This would be something like `:insync`, `:absent`, `:mismatch`, right? I think consumers of this (for existing modules) determine the status of insync/mismatch/absent from whether the actual equals the expected, or if it differs, and if it differs if the actual was nil, respectively. I think we want to continue that behavior.

reidmv: Better? e783dcc6366d8ec6badbdc5a8c0d6d2b6ddc71ec

I'm not actually sure how/where/why #properties is used, or what expectations exist for this method. The YARDoc definition in R10K::Module::Base isn't super helpful on that front.

justinstoller: They are collected to write the r10k-deploy.json, which is used by Impact Analysis.

justinstoller (approved):

Magisus (approved):

PR #1245, opened 2021-11-09
(CODEMGMT-1300) Update rugged cache remote if it has changed

jarretlavallee:
Prior to this commit, if the rugged cache repository had a different
remote url than the current configuration, it would not use the current
configured url. This would cause r10k deployment failures when there was
a malformed or incorrect cached remote URL. This commit changes that
behavior by updating the remote URL if the current configuration and the
cache differ. The current configuration will be used instead of the
cached URL.

To repro this issue, one can simply edit the `config` file in the `r10k` cache and change the `origin` remote URL to be malformed. E.g

```
sed -i 's#github.com/puppetlabs#github.com:puppetlabs#' /var/cache/r10k/https---github.com-puppetlabs-puppetlabs-apache.git/config
 /opt/puppetlabs/puppet/bin/r10k deploy environment -v debug2 -p --trace
```

Without this change, we get the following until the cache is cleared or fixed.

```
[2021-11-09 14:15:00 - ERROR] malformed URL 'https://github.com:puppetlabs/puppetlabs-apache.git' at /var/cache/r10k/https---github.com-puppetlabs-puppetlabs-apache.git at /var/cache/r10k/https---github.com-puppetlabs-puppetlabs-apache.git
```

With this change, we have a successful deployment and the following is logged. 

```
[2021-11-09 14:14:31 - INFO] Deploying module to /etc/puppetlabs/code/environments/production/modules/apache
[2021-11-09 14:14:31 - DEBUG2] Remote URL is different from cache, updating https://github.com:puppetlabs/puppetlabs-apache.git to https://github.com/puppetlabs/puppetlabs-apache.git
[2021-11-09 14:14:31 - DEBUG1] Fetching remote 'origin' at /var/cache/r10k/https---github.com-puppetlabs-puppetlabs-apache.git
[2021-11-09 14:14:32 - DEBUG2] Transferred 0 objects (0 bytes) from 'origin' into /var/cache/r10k/https---github.com-puppetlabs-puppetlabs-apache.git'
```

### DIFF
--- a/lib/r10k/git/rugged/base_repository.rb
+++ b/lib/r10k/git/rugged/base_repository.rb
@@ -61,6 +61,17 @@ class R10K::Git::Rugged::BaseRepository
     remotes_hash
   end
 
+  # Update a remote URL
+  # @param [String] The remote URL of the git repository
+  # @param [String] An optional remote name for the git repository
+  def update_remote(remote, remote_name='origin')
+
+    if @_rugged_repo && remotes[remote_name] != remote
+      logger.debug2(_("Remote URL is different from cache, updating %{orig} to %{update}") % {orig: remotes[remote_name], update: remote})
+      @_rugged_repo.remotes.set_url(remote_name, remote)
+    end
+  end
+
   private
 
   def with_repo(opts={})
--- a/lib/r10k/git/rugged/base_repository.rb
+++ b/lib/r10k/git/rugged/base_repository.rb
@@ -65,8 +65,7 @@ class R10K::Git::Rugged::BaseRepository
   # @param [String] The remote URL of the git repository
   # @param [String] An optional remote name for the git repository
   def update_remote(remote, remote_name='origin')
-
-    if @_rugged_repo && remotes[remote_name] != remote
+    if @_rugged_repo
       logger.debug2(_("Remote URL is different from cache, updating %{orig} to %{update}") % {orig: remotes[remote_name], update: remote})
       @_rugged_repo.remotes.set_url(remote_name, remote)
     end
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -6,6 +6,7 @@ Unreleased
 
 - (CODEMGMT-1294) Resync repos with unresolvable refs [#1239](https://github.com/puppetlabs/r10k/pull/1239)
 - (RK-378) Restore access to the environment name from the Puppetfile [#1241](https://github.com/puppetlabs/r10k/pull/1241)
+- (CODEMGMT-1300) Ensure the remote url in rugged cache directories is current [#1245](https://github.com/puppetlabs/r10k/pull/1245)
 
 3.13.0
 ------
--- a/lib/r10k/git/rugged/cache.rb
+++ b/lib/r10k/git/rugged/cache.rb
@@ -8,4 +8,12 @@ class R10K::Git::Rugged::Cache < R10K::Git::Cache
   def self.bare_repository
     R10K::Git::Rugged::BareRepository
   end
+
+  # Update the remote URL if the cache differs from the current configuration
+  def sync!
+    if cached?
+      @repo.update_remote(@remote)
+    end
+    super
+  end
 end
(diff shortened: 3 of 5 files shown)


lib/r10k/git/rugged/base_repository.rb
```diff
@@ -61,6 +61,17 @@ def remotes
     remotes_hash
   end
 
+  # Update a remote URL
+  # @param [String] The remote URL of the git repository
+  # @param [String] An optional remote name for the git repository
+  def update_remote(remote, remote_name='origin')
+
```
Magisus: Extra newline here

Magisus (commented):
I'm debating whether it's worth trying to add some tests here, perhaps in https://github.com/puppetlabs/r10k/blob/main/spec/unit/git/rugged/cache_spec.rb. But I'm not sure how complicated that would be.

Almost feels like it would be better to do as a beaker test, mirroring the user workflow that can trigger this bug, as described in the ticket (change the remote in the Puppetfile). But I think I'd be fine merging this even without that.

jarretlavallee:
I initially gave some unit tests a try. Unfortunately, I got to the point where I was mocking everything up except for some if statements. The integration tests seem like a better place for this since we can use a real repo. I'll give that a go tomorrow.

jarretlavallee:
I put in some spec tests for this change. The integration tests out of the box are failing on the thin_repository. Are these tests used or should I write tests in a different place?

```
  1) R10K::Git::Rugged::ThinRepository behaves like a git thin repository cloning creates a working copy of the repo
     Failure/Error: object = with_repo { |repo| repo.rev_parse(pattern) }
     
     TypeError:
       wrong argument type nil (expected String)
     Shared Example Group: "a git thin repository" called from ./spec/integration/git/rugged/thin_repository_spec.rb:13
     # ./vendor/gems/ruby/2.7.0/gems/rugged-1.2.0/lib/rugged/repository.rb:154:in `rev_parse'
     # ./vendor/gems/ruby/2.7.0/gems/rugged-1.2.0/lib/rugged/repository.rb:154:in `rev_parse'
     # ./lib/r10k/git/rugged/base_repository.rb:14:in `block in resolve'
     # ./lib/r10k/git/rugged/base_repository.rb:78:in `with_repo'
     # ./lib/r10k/git/rugged/working_repository.rb:139:in `with_repo'
     # ./lib/r10k/git/rugged/base_repository.rb:14:in `resolve'
     # ./lib/r10k/git/rugged/working_repository.rb:64:in `checkout'
     # ./lib/r10k/git/rugged/thin_repository.rb:48:in `checkout'
     # ./lib/r10k/git/rugged/thin_repository.rb:44:in `clone'
     # ./spec/shared-examples/git/thin_repository.rb:4:in `block (3 levels) in <top (required)>'
```

Magisus:
Looks like you sorted out your integration test issue, since all the tests are passing now.

jarretlavallee:
👍  for passing tests. The integration tests are still failing on the HEAD of `main`.  Travis is configured to only runs `unit/spec` and not `unit/integration`. https://github.com/puppetlabs/r10k/blob/main/.travis.yml#L4 With that in mind, is it worth implementing some tests there since they are not called anyway?

Magisus:
Oh I see, let me take a look. Good catch, that seems odd that Travis doesn't run everything.

Magisus:
All tests (unit and integration) pass for me locally, using Ruby 2.7.1 and rugged 1.2.0 (which looks the same as you). I'm going to update travis to run integration tests too and see what happens. See https://github.com/puppetlabs/r10k/pull/1246

Magisus:
I updated both GHA and Travis to run all specs, and they appear to be passing... must be something with your env, but I don't really know what it could be...

jarretlavallee:
Thanks for doing that! I'll figure out my local environment issues and add some integration tests for this one.

Magisus:
Sounds good, thanks! Let us know if you need any more help.

jarretlavallee:
It turns out that I had the `defaultBranch` set to main in my `.gitconfig`, which caused the cache repo to have the `HEAD` of `main` instead of `master` which the remote repo was using. Since HEAD didn't resolve, it was failing to check out the repo.

```
[init]
	defaultBranch = main
```

Anyway, I added some integration tests for this feature which look to be working now.

Magisus (approved):
Thanks for adding those!

I also went ahead and manually tested this following the exact repro from the ticket, and can confirm this solves this issue 👍

PR #1252, opened 2022-01-02
Use correct number of components for minitar version

tambry:
minitar is versioned only with two components. When packaging for Debian this causes a dependency on version 0.9.0, which is not satisfied by 0.9 that Debian does have.


Magisus (approved):
I checked, looks like we're specifying this correctly in our packaging project already. Thanks for catching this!

PR #1256, opened 2022-01-04
(maint) Make travis and GH Actions do the right thing on the 4.x branch

Magisus:
This drops Ruby 2.3 testing from Travis, in preparation for us dropping
support for that version, leaving just container testing in Travis. It
also updates the spec and release Github Actions to trigger on relevant
actions on the 4.x branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,11 @@
 ---
 language: ruby
 bundler_args: "--without system"
-script: "bundle exec rspec --color --format documentation spec"
 notifications:
   email: false
 sudo: false
-jdk:
-  - openjdk11
-before_install: gem install bundler -v '< 2' --no-document
-matrix:
+jobs:
   include:
-    - stage: r10k tests
-      rvm: 2.3.0
     - stage: r10k container tests
       dist: focal
       language: ruby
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

.github/workflows/release.yml
```diff
@@ -3,7 +3,7 @@ name: Tag and release
 on:
   push:
     branches:
-      - main
+      - 4.x
```
bastelfreak: will this branch be merged into main? in that case I suggest keeping main here

Magisus: Yeah, I hadn't decided whether we would release 4.0 off of this and keep main as a 3.x stream, since we'll likely be stuck on that in PE for a while. But I think that doesn't make sense, and we should branch `3.x` when the time comes, then merge this into main. I'll take this out.

bastelfreak (approved):

.travis.yml
```diff
@@ -1,17 +1,11 @@
 ---
 language: ruby
 bundler_args: "--without system"
-script: "bundle exec rspec --color --format documentation spec"
 notifications:
   email: false
 sudo: false
-jdk:
-  - openjdk11
```
mwaggett, line 9: The docker tests don't need this? If not, looks good to me!

Magisus: Everything seems to run fine. I don't think docker depends on Java... I honestly don't know why we had that in here, might have been copy-pasta.

mwaggett (approved):

PR #1258, opened 2022-01-13
(CODEMGMT-1295) Update default module git branch to `main`

Magisus:
The git ecosystem in general and the Puppet module ecosystem in
particular are moving from `master` as the default branch to `main`.
This commit updates r10k's default to match this shift.

Now, if a git module is defined in a Puppetfile with no ref specified,
r10k will attempt to deploy its `main` branch. If this is undesirable
for a user's module, they should specify a ref explicitly to use
instead.


Magisus:
Tested this out, seems to work as expected, cloning `main` if no ref is specified for a Git module in the Puppetfile.

When we go to integrate this into pe-r10k, there might be some test changes needed, but that isn't possible for this 4.x branch right now, so I haven't dealt with that in here.

justinstoller (approved):

PR #1268, opened 2022-02-03
(RK-380) Document `oauth_token` setting

mwaggett:


doc/dynamic-environments/configuration.mkd
```diff
@@ -128,6 +128,18 @@ git:
   private_key: "/etc/puppetlabs/r10k/ssh/id_rsa"
 ```
 
+#### oauth_token
+
+The oauth_token setting is only used by the Rugged git provider.
+
+The oauth_token option specifies the path to the default access token for Git HTTPS remotes.
+Public git repositories can be accessed via HTTPS without authentication, but the oauth_token setting must be set if any non-public HTTPS remotes are used.
```
Magisus: Note that this only works for repos that are actually configured to use OAuth (as opposed to just user/pass). But I suppose that's self-explanatory?

mwaggett: yeah, I wasn't sure whether to mention that or not.. I sort of figured no mention of user/pass would imply that that was not supported. I don't _think_ there are any specific repo settings for allowing tokens vs. user/pass, besides just having the token configured with repo permissions. but I could be wrong. What do you think? Should I add a note?

Magisus: It's probably okay. I think you're right that the rest can be inferred.

Magisus: Though, you can use user/pass, just not from these settings, it can be parsed from the URL: https://github.com/puppetlabs/r10k/blob/e032473cee6948d7ba1e072ed8e098d1c8e5be32/lib/r10k/git/rugged/credentials.rb#L93-L94. So maybe rephrase to make it sound less like "you must use oauth"?

mwaggett: ahh okay!

Magisus: I haven't tested that behavior myself though. It doesn't look like we document it anywhere I can find...

Magisus (approved):

PR #1282, opened 2022-02-21
Release docker image for arm64

odidev:
The following file has been modified:
Modified **Makefile** and **.github/workflows/docker.yml** file to release the docker image for both the platforms using buildx.

Signed-off-by: odidev [odidev@puresoftware.com](mailto:odidev@puresoftware.com)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -27,6 +27,11 @@ jobs:
         with:
           ruby-version: 2.6.x
       - run: gem install bundler
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
       - name: Build container
         working-directory: docker
         run: |
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -60,8 +60,19 @@ build: prep
 		--build-arg pupperware_analytics_stream=$(PUPPERWARE_ANALYTICS_STREAM) \
 		--file r10k/$(dockerfile) \
 		--tag $(NAMESPACE)/r10k:$(VERSION) $(dockerfile_context)
+	docker buildx build \
+		--platform linux/arm64 \
+		--build-arg alpine_version=$(alpine_version) \
+		--build-arg vcs_ref=$(vcs_ref) \
+		--build-arg build_date=$(build_date) \
+		--build-arg version=$(VERSION) \
+		--build-arg pupperware_analytics_stream=$(PUPPERWARE_ANALYTICS_STREAM) \
+		--file r10k/$(dockerfile) \
+		--tag $(NAMESPACE)/r10k:$(VERSION)-arm64 \
+		--load $(dockerfile_context)
 ifeq ($(IS_LATEST),true)
 	@docker tag $(NAMESPACE)/r10k:$(VERSION) puppet/r10k:$(LATEST_VERSION)
+	@docker tag $(NAMESPACE)/r10k:$(VERSION) puppet/r10k:$(LATEST_VERSION)-arm64
 endif
 
 test: prep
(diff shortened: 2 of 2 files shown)


docker/Makefile
```diff
@@ -72,9 +72,9 @@ test: prep
 		rspec spec
 
 push-image: prep
-	@docker push $(NAMESPACE)/r10k:$(VERSION)
+	docker buildx build --platform linux/arm64,linux/amd64 -t $(NAMESPACE)/r10k:$(VERSION) -f r10k/$(dockerfile) --push $(dockerfile_context)
```
justinstoller: Thanks so much for this PR!

This is probably just showing my own ignorance about docker and this set up, but why wouldn't this go into the `build` target?

odidev: [@justinstoller](https://github.com/justinstoller)
Thanks for your response,
Here, we are using **buildx build** for building multiarchitecture images, where we need to use `--push` flag with buildx in makefile. 
If we want to build for single platform at a time, we can use `--load` flag to build and then push later.

MikaelSmith: I think Justin's point was that there's a separate `docker buildx build` at line 53, we should probably update that step.

If we also need to use buildx to push, this change also makes sense. Does buildx not really let you separate out build and push when working with multiple platforms?

.github/workflows/docker.yml
```diff
@@ -27,6 +27,11 @@ jobs:
         with:
           ruby-version: 2.6.x
       - run: gem install bundler
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
```
Iristyle, line 33: We're already setting up Docker with buildx and shouldn't need this

docker/Makefile
Iristyle: Build and push are different steps in our Makefile. This will rebuild a container image we've already built... and without all the build flags / tags we expect.

justinstoller:
Hey @odidev , I'm happy to merge your work once we make sure it's integrated with our existing patterns for building and releasing in the Makefile. (I asked Ethan and Michael to review since they spearheaded writing our docker workflows).

odidev:
@justinstoller, 

 

As per the suggestion that build and push are different steps in Makefile. I have made some code changes in the Makefile in which arm64 image is build and push as it is done for amd64. 

  

Commit link:  https://github.com/odidev/r10k/commit/376e77500723df351026672763beeccf8fda9869 

  

If the above changes looks good, I will amend the PR.

MikaelSmith:
https://github.com/odidev/r10k/commit/376e77500723df351026672763beeccf8fda9869 looks reasonable to me.

justinstoller:
@odidev :+1: If Michael's happy, I'm happy. I'll merge once amended.

odidev:
I have amended the PR, please let me know if any more changes are required.

justinstoller (approved):

justinstoller:
Thanks so much for working with us, @odidev !

docker/Makefile
```diff
@@ -60,8 +60,19 @@ build: prep
 		--build-arg pupperware_analytics_stream=$(PUPPERWARE_ANALYTICS_STREAM) \
 		--file r10k/$(dockerfile) \
 		--tag $(NAMESPACE)/r10k:$(VERSION) $(dockerfile_context)
+	docker buildx build \
+		--platform linux/arm64 \
```
Iristyle, line 64: This should have `${DOCKER_BUILD_FLAGS`}` like above to properly output to Travis

Iristyle: Without it we get a TON of noise like https://app.travis-ci.com/github/puppetlabs/r10k/jobs/564599445#L498-L515

docker/Makefile
```diff
@@ -60,8 +60,19 @@ build: prep
...
+		--build-arg vcs_ref=$(vcs_ref) \
+		--build-arg build_date=$(build_date) \
+		--build-arg version=$(VERSION) \
+		--build-arg pupperware_analytics_stream=$(PUPPERWARE_ANALYTICS_STREAM) \
+		--file r10k/$(dockerfile) \
+		--tag $(NAMESPACE)/r10k:$(VERSION)-arm64 \
+		--load $(dockerfile_context)
 ifeq ($(IS_LATEST),true)
 	@docker tag $(NAMESPACE)/r10k:$(VERSION) puppet/r10k:$(LATEST_VERSION)
+	@docker tag $(NAMESPACE)/r10k:$(VERSION) puppet/r10k:$(LATEST_VERSION)-arm64
```
Iristyle, line 75: We shouldn't be using different tags for different architectures

Iristyle:
Put up #1286 to fix the issues with this PR

PR #1285, opened 2022-03-11
(RK-216) Make tarball module and environment sources require wrapper directories; be compatible with `pdk build`

reidmv:
This commit modifies the tarball module and environment types to expect wrapper directories.

Previously, a tarball was expected to contain content with no wrapper. E.g.

```
shell$ tar -tzf archive-1.2.3.tar.gz
./Puppetfile
./metadata.json
…
```

This is not typically how people create tarballs, and it is not how `pdk build` creates tarballs. Rather, it is typical for tarballs to have a "wrapper" directory, which helps avoid clutter when unpacking them. E.g.

```
shell$ tar -tzf archive-1.2.3.tar.gz
archive-1.2.3/Puppetfile
archive-1.2.3/metadata.json
archive-1.2.3/…
```

This commit updates the R10K tarball facilities to always expect a wrapper directory, and handle it appropriately.

Most notably, this change means that content downloaded directly from the forge, or created with `pdk build`, is automatically compatible with R10K's tarball module type.

This commit is not backwards compatible with previous versions of the tarball type(s). Open for discussion as to whether or not this is acceptable, given feature maturity (experimental?) and likely/possible adoption personas (innovators/experimenters vs. mainstream).

This behavior would constitute a finished implementation of RK-216.


lib/r10k/tarball.rb
```diff
@@ -70,28 +87,54 @@ def cache_basename
     end
 
     # Extract the cached tarball to the target directory.
-    #
+    # 
     # @param target_dir [String] Where to unpack the tarball
+    # @param remove_wrapper_dir [Boolean] Whether or not to remove the wrapper
```
tvpartytonight: does this need to be added to the parameter list for `unpack`?

reidmv: Ah; nope, that's an rdoc `@param` leftover from a previous implementation. `remove_wrapper_dir` is now an attr reader method. I will remove the `@param` note.

reidmv:
@tvpartytonight are you still the person to get approval for this from, or has that changed?

jonathannewman:
Marked as DNM pending security review.

reidmv:
@jonathannewman is the security review relevant to the changes this PR makes, or is that more appropriately captured as a separate issue? I believe the area of interest is already shipped code.

reidmv:
Upon self-review, caught one possible issue involving an assumption that a method input was an absolute path. Fixed concern and added some comments in a new commit [c24206d](https://github.com/puppetlabs/r10k/pull/1285/commits/c24206d360df2c1f47bc238272a598f9720a955c).

mwaggett:
Hi @reidmv, thanks for opening this! r10k is now considered to be feature complete and we are unfortunately discontinuing development. As a result, we are closing all open issues and pull requests. I sincerely apologize that we’re unable to move forward with this PR.

PR #1287, opened 2022-03-30
(maint) Replace `git://` protocol with `https://`

mwaggett:
This commit updates tests and documentation to use the `https` protocol for
cloning repos, rather than `git`. On March 15, 2022, GitHub permanently disabled
usage of the unencrypted `git` protocol, so tests began to fail with the
following error:
```
The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

```
```

### DIFF
--- a/doc/dynamic-environments/configuration.mkd
+++ b/doc/dynamic-environments/configuration.mkd
@@ -404,7 +404,7 @@ passwords cannot be prompted for in order to fetch the remote.
 ---
 sources:
   mysource:
-    remote: 'git://git-server.site/my-org/main-modules'
+    remote: 'https://git-server.site/my-org/main-modules'
 ```
 
 ### basedir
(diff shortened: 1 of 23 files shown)


doc/dynamic-environments/configuration.mkd
```diff
@@ -404,7 +404,7 @@ passwords cannot be prompted for in order to fetch the remote.
 ---
 sources:
   mysource:
-    remote: 'git://git-server.site/my-org/main-modules'
+    remote: 'https://git-server.site/my-org/main-modules'
```
Magisus, line 410: Maybe in the blurb above this would be a good place to call out that we support `git`, `ssh`, and `https`?

Magisus (commented):
I think this is probably fine, but it might be good to note somewhere that _r10k_ does still support the Git protocol. Just because Github disabled it doesn't mean it wouldn't work with people's private git servers. I have no idea how common that usecase is, but I know for sure not everyone is using Github.

doc/dynamic-environments/configuration.mkd
mwaggett: sure, I can add something

mwaggett: Added.

mwaggett:
Confirmed locally that this fixes the integration test failures I was seeing.

Magisus (approved):

PR #1291, opened 2022-04-07
(RK-399) Do not warn on spec changes with exclude_spec

jarretlavallee:
Prior to this change, r10k would warn about local changes on the files
in the spec dir when exclude_spec was set. Since r10k was removing the
files, the warning would always be present for each git module when
exclude_spec was set. This change removes the spec files from the insync
checks so that the repo will not be marked as dirty if there are only changes
to the spec files. After this change, the warning about local
modifications on the spec files should not be present.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,8 @@ CHANGELOG
 Unreleased
 ----------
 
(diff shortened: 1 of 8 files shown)


tvpartytonight:
Thanks for doing this @jarretlavallee ! I'll take it for a spin later today.

tvpartytonight (approved):

PR #1295, opened 2022-04-24
(maint) Convert gettext-setup to development dep

bastelfreak:
based on a [slack](https://puppetcommunity.slack.com/archives/C0W1X7ZAL/p1650650138731649) thread this is only a development dependency, not a
runtime dependency.

```
```

### DIFF
--- a/lib/r10k.rb
+++ b/lib/r10k.rb
@@ -1,8 +1,16 @@
-require 'gettext-setup'
-
 module R10K
-  GettextSetup.initialize(File.absolute_path('../locales', File.dirname(__FILE__)))
+  begin
+    # requires the `documentation` gem group
+    require 'gettext-setup'
+    GettextSetup.initialize(File.absolute_path('../locales', File.dirname(__FILE__)))
+  rescue LoadError
+  end
 
+  require 'fast_gettext'
+  require 'locale'
+  class ::Object
+    include FastGettext::Translation
+  end
   # Attempt to set the R10k error and log message locale
   FastGettext.locale = ENV["LANG"]
 end
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,11 @@ group :development do
   gem 'ruby-prof', :platforms => :ruby
 end
 
+group :documentation do
+  # only required to update the pot files, not a runtime dep
+  gem 'gettext-setup', '~> 0.24'
+end
+
 if File.exists? "#{__FILE__}.local"
   eval(File.read("#{__FILE__}.local"), binding)
 end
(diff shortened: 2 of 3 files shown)


bastelfreak:
As per conversation on slack, I'm moving the gettext-setup Gem to another bundler group. So it isn't a runtime dep. It's still installed during CI because puppet_forge pulls it in. I will provide a PR there as well.

Edit: Oh wow I kinda already did that in the past: https://github.com/puppetlabs/forge-ruby/pull/91

bastelfreak:
I do not know a lot about the whole gettext magic. I it's really optional lib/r10k.rb needs to be updated. I copied some logic from the puppet gem but it's behaving as I like to be. If someone has some guidance please let me know :)

lib/r10k.rb
```diff
@@ -1,8 +1,16 @@
-require 'gettext-setup'
-
 module R10K
-  GettextSetup.initialize(File.absolute_path('../locales', File.dirname(__FILE__)))
+  begin
+    # requires the `documentation` gem group
+    require 'gettext-setup'
+    GettextSetup.initialize(File.absolute_path('../locales', File.dirname(__FILE__)))
+  rescue LoadError
+  end
```
Magisus, line 7: Did you try without this bit? I think we don't have this in Puppet anymore, and I don't _think_ it's necessary, but it's been a while and I might be forgetting something.

bastelfreak: I'm not sure how the whole gettext setup is supposed to look like, so I didn't really test it :(. I copied that code from puppet, where Josh found it. I noticed that my markdown in the PR description was wrong and the slack link, with some background infos, wasn't visible. I fixed that.

Magisus: Okay I double checked, and without `GettextSetup.initialize` or the stuff Puppet does in [this class](https://github.com/puppetlabs/puppet/blob/d894861565aa95bd6f4f1cb85fe44ab62cc37854/lib/puppet/gettext/config.rb), we will never use translations at runtime. Since we don't expect `GettextSetup` to be installed ever, translations will not work.

However, we don't _have_ any translations, and seem unlikely to add any, so I'm not sure we care. Given that, I'd be in favor of removing this block, and just leaving the bit below where we add the translation methods (or even just adding [stubs](https://github.com/puppetlabs/puppet/blob/d894861565aa95bd6f4f1cb85fe44ab62cc37854/lib/puppet/gettext/stubs.rb) instead), along with a comment that translation is deprecated and the translation wrapper methods will be removed later.

Magisus: Yeah this definitely requires `GettextSetup.initialize`. It blows up hard if you don't install the `documentation` group. I think we'll have to move to stubs if we want this. However, without this PR, I am able to successfully install on both Ruby 3 with bundler, and using the agent's Ruby 2.6, with the same gemspec (your other PRs besides this one). Is that sufficient? Or is there another reason you wanted to convert gettext-setup to a dev dep?

PR #1297, opened 2022-04-24
delete broken .travis.yml

bastelfreak:
travis is dead :(

```
```


Magisus:
That's unfortunate, that was the only place we were getting Ruby 2.3 coverage. Wonder if it's worth trying to fix it...

Magisus:
We decided it's not worth trying to fix. We won't officially drop Ruby 2.3 support, but we're okay not testing it on every PR.

Magisus (approved):
Agreed, we replaced it mostly with GH Actions.

PR #1299, opened 2022-04-25
Add TOC to configuration docs

### LINKED ISSUE #1298: Add a TOC to the configuration doc
The doc/dynamic-environments/configuration.mkd doc is structured, but it's too large to easily discern that structure.  It can be difficult for users to find their way around without losing the context of the documentation they eventually find, especially for beginners.

It would helpful to add a hyperlinked table of contents to the beginning of the document (like [Puppet modules's README](https://puppet.com/docs/puppet/7/modules_documentation.html#readme-table-of-contents))  to give readers a roadmap to the document and an index for each option under the context of its section.

op-ct:
The configuration doc is structured, but it's too large to easily
understand what that structure is.  It can be difficult for users to find 
their way around without losing the context of the documentation 
they eventually find.

This patch adds a hyperlinked table of contents to the configuration
doc, to clarify the document's structure and provide an index of each
option with context.

Fixes #1298


mwaggett (approved):
This looks good to me, thanks so much for adding this @op-ct !

PR #1307, opened 2022-04-30
Merge main -> 4.x

mwaggett:


lib/r10k/git/rugged/working_repository.rb
```diff
@@ -117,11 +117,15 @@ def origin
     with_repo { |repo| repo.config['remote.origin.url'] }
   end
 
-  def dirty?
+  def dirty?(exclude_spec=false)
```
justinstoller: I forget, is this default supposed to be `true` in 4.x?

mwaggett: ooh hmm, not sure, I'll look

mwaggett: looks like it is https://github.com/puppetlabs/r10k/blob/4.x/doc/dynamic-environments/configuration.mkd#exclude_spec 
good catch!

mwaggett: updated!

justinstoller (approved):

PR #1308, opened 2022-05-13
Clarify "modifiedenvs" in postrun documentaton

theothertom:
Document how multiple environments will be specified when there is more than one


mwaggett (approved):

mwaggett:
Thanks @theothertom !

PR #1292, opened 2022-04-13
Implement exclude regex for puppetfile install

baurmatt:
This implements an exclude regex option for `puppetfile install` as described in #1248. I've tested this on CLI and it seems to work as intended. Sadly I wasn't able to understand how I'm supposed to add a unit test for this. Could you please help me with it?

---
Fixes #1248.

PR #1313, opened 2022-06-15
Add maintenance notice to the README

binford2k:
r10k is now a mature product. This is intended to clarify our maintenance
committment and roadmap going forward. Approved by @david-piekny

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -5,13 +5,12 @@ Puppet environment and module deployment
 
 [![Build Status](https://travis-ci.org/puppetlabs/r10k.png?branch=master)](https://travis-ci.org/puppetlabs/r10k)
 
-```
-R10k is supported and maintained by Puppet, but we consider it to be feature complete
-and  currently have no plans for future feature development. We will keep it working
-with current code deployment standards and the Ruby ecosystem that ships with the
-puppet-agent package, but any new feature development will come from peer-reviewed
-community contributions, with the onus on the contributor and community for validation.
-```
+> R10k is supported and maintained by Puppet, but we consider it to be feature complete
+> and  currently have no plans for future feature development. We will keep it working
+> with current code deployment standards and the Ruby ecosystem that ships with the
+> puppet-agent package, but any new feature development will come from peer-reviewed
+> community contributions, with the onus on the contributor and community for validation.
+
 
 Description
 -----------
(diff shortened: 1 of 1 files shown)


README.mkd
```diff
@@ -5,6 +5,14 @@ Puppet environment and module deployment
 
 [![Build Status](https://travis-ci.org/puppetlabs/r10k.png?branch=master)](https://travis-ci.org/puppetlabs/r10k)
 
+```
```
mwaggett: I don't think this needs to be in a code block?

README.mkd
```diff
@@ -5,6 +5,14 @@ Puppet environment and module deployment
...
+R10k is supported and maintained by Puppet, but we consider it to be feature complete
+and  currently have no plans for future feature development. We will keep it working
+with current code deployment standards and the Ruby ecosystem that ships with the
+puppet-agent package, but any new feature development will come from peer-reviewed
+community contributions, with the onus on the contributor and community for validation.
```
mwaggett: I don't even know if we should mention this part about new feature development - it would still require a Puppet engineer to review, merge, release the change, and make sure it doesn't negatively impact PE, and it sounds like that won't happen unless Product prioritizes it.

mwaggett: @binford2k ^this point still stands. I don't want to set the expectation that we will be merging community contributions on a regular basis.

themicahwilson: @mwaggett I think the message is fine as is. Let's get it up, can you approve it? Any open issues can be closed out with a "won't do", but it's not a priority to do this, it's really if someone has time and wants to. Looks like https://github.com/puppetlabs/r10k/pull/1315 and https://github.com/puppetlabs/r10k/pull/1292 might be worth pushing through, but it's not a priority, so if @binford2k has time, awesome, otherwise we will leave them.

mwaggett: The hesitation I have with this message is that "peer-reviewed community contributions, with the onus on the contributor and community for validation" still requires a Puppet engineer to review and merge the PR and confirm that it doesn't negatively impact PE or CD4PE. So any such PRs will continue to sit open unless you continue to monitor this repo and let us know when we should review those PRs. If that's part of the plan, I am happy to merge this, but if not, I don't want to set the set the expectation that we will merge PRs purely on the faith of the community, since the community can't validate against PE.

mwaggett: okay @themicahwilson wants this left as-is, so merging.

mwaggett (approved):

PR #1315, opened 2022-06-28
Remove spec from gem spec

davidsandilands:
This change updates gemspec to ignore the Spec directory when creating a gem file to allow the r10k gem to be installed on windows servers without admin or symlink creation access, this in turn allows PDK to be installed as a rubygem as per https://github.com/puppetlabs/r10k/issues/1316

### DIFF
--- a/r10k.gemspec
+++ b/r10k.gemspec
@@ -43,7 +43,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'yard', '~> 0.9.11'
 
-  s.files        = %x[git ls-files].split($/)
+  s.files        = %x[git ls-files].split($/).reject { |f| f.match(%r{^spec}) }
   s.require_path = 'lib'
   s.bindir       = 'bin'
   s.executables  = 'r10k'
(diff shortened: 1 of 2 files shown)


mwaggett (approved):

r10k.gemspec
```diff
@@ -43,7 +43,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'yard', '~> 0.9.11'
 
-  s.files        = %x[git ls-files].split($/)
+  s.files        = %x[git ls-files].split($/).reject { |f| f.match(%r{^spec}) }
```
donoghuc, line 46: Shelling out to git seems like a very strange way to determine what files to ship as the gem artifact. Would it not better to just explicitly ship the code we expect to be needed in the gem? maybe something like:
```ruby
s.files = Dir['bin/**'] + Dir['lib/**']
```
Using ^^ pattern would also avoid shipping the files under `docker` `docs` `integration` etc (which the filter in this pr does not exclude).

mwaggett: @donoghuc Feel free to open a PR if you wanna change this. 😜

PR #1325, opened 2022-12-05
(maint) Allow gettext-setup < 2

joshcooper:
Accept gettext-setup >= 1.1.0 for compatibility with fast-gettext 2.1 and Ruby 3.2.

See https://github.com/puppetlabs/gettext-setup-gem/commit/f4b292c51d2c72e76515f3b81d5ebd379a2bbda8

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -5,6 +5,7 @@ Unreleased
 ----------
 
 - Fix dirty working copy debug logging [#1321](https://github.com/puppetlabs/r10k/pull/1321)
+- Allow gettext-setup < 2 for compatibility with Ruby 3.2 and Puppet 8 [#1325](https://github.com/puppetlabs/r10k/pull/1325)
 
 3.15.2
 ------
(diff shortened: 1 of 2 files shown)


CHANGELOG.mkd
```diff
@@ -5,6 +5,7 @@ Unreleased
 ----------
 
 - Fix dirty working copy debug logging [#1321](https://github.com/puppetlabs/r10k/pull/1321)
+- Allow gettext-setup < 2 [#1325](https://github.com/puppetlabs/r10k/pull/1325)
```
mwaggett: I think we should add a little more detail here, explaining that this will be necessary for Puppet 8 compatibility.

joshcooper: Updated!

joshcooper:
I can't add 3.2.0-rc1 to the github action because rugged doesn't support 3.2 yet (it still calls the removed method File.exists?) https://github.com/libgit2/rugged/issues/943

mwaggett:
ahh okay. That is something we should keep track of - once we ship Ruby 3.2 with PE, sounds like we'll need to update the version of rugged that we vendor.

mwaggett (approved):

mwaggett:
I opened https://tickets.puppetlabs.com/browse/PE-35010 to track the rugged update.

PR #1328, opened 2022-12-19
(maint) Updates build and publish workflow

mhashizume:
This commit updates the build and publish GitHub workflow to use the ruby/setup-ruby Action (instead of the deprecated action/setup-ruby) and use Ruby 2.7 (instead of the end-of-life Ruby 2.6).


mhashizume:
The latest gem release failed because of the deprecated action that this PR updates: https://github.com/puppetlabs/r10k/actions/runs/3735539538

mhashizume:
I opened this PR because of this failing workflow: https://github.com/puppetlabs/r10k/actions/runs/3735539538

But I realized that it may be moot since we're removing Docker support anyway https://github.com/puppetlabs/r10k/pull/1326

PR #1330, opened 2023-01-19
Merge main -> 4.x

mwaggett:
```
```


jonathannewman:
We aren't planning a 4.x release of r10k now that it is in maintenance only mode, so this isn't necessary.

mwaggett:
There may be community members who may be using the 4.x branch, and there may be a time when we need to drop support for old Ruby versions and a 4.0 release is necessary, even though we don't have a plan for one right now. I don't see the harm in just merging this.

PR #1334, opened 2023-03-15
(maint) Emit more debug output when mod sync fails

justinstoller:
```
```

### DIFF
--- a/lib/r10k/content_synchronizer.rb
+++ b/lib/r10k/content_synchronizer.rb
@@ -79,8 +79,13 @@ module R10K
         begin
           while mods = mods_queue.pop(true) do
             mods.each do |mod|
-              updated = mod.sync
-              updated_modules << mod.name if updated
+              begin
+                updated = mod.sync
+                updated_modules << mod.name if updated
+              rescue Exception => e
+                logger.debug _("Module %{mod_name} failed to synchronize due to %{message}") % {mod_name: mod.name, message: e.message}
+                raise e
+              end
             end
           end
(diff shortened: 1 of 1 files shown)


lib/r10k/content_synchronizer.rb
```diff
@@ -79,8 +79,13 @@ def self.sync_thread(mods_queue, logger, updated_modules)
         begin
           while mods = mods_queue.pop(true) do
             mods.each do |mod|
-              updated = mod.sync
-              updated_modules << mod.name if updated
+              begin
+                updated = mod.sync
+                updated_modules << mod.name if updated
+              rescue Exception => e
+                logger.debug _("Module %{mod_name} failed to synchronize due to %{message}") % {mod_name: mod.name, message: e.message}
```
Sharpie, line 86: Looks good, but this should be at `error` or `warn` level so that the information is always visible without having to re-run with debug enabled (difficult to do in Code Manager, and doubly so with ephemeral environments like Impact Analysis branches).

Sharpie (changes requested):

lib/r10k/content_synchronizer.rb
Sharpie: `error` level seems the most appropriate as the sort of occurrence causes `r10k` to exit with a non-0 status, which is fatal to the code deployment process.

justinstoller:
Couldn't reopen this PR so I created https://github.com/puppetlabs/r10k/pull/1347

PR #1351, opened 2023-07-25
(PE-36512) Move testing to production forge

justinstoller:
As part of the work to update r10k for Puppet 8 we noticed many outdated test module dependencies. We have been relying on an internal forge for testing artifacts for years but it has fallen out of maintenance due to internal restructuring. The production forge's backend has been re-written in the mean time and the forge team would rather we point our testing at the production forge than maintain an internal testing-only forge.

This moves us away from stubbing forge connections, deletes several tests that are of limited value but expensive to transition to the production forge, and makes the tests more robust in the face of the forge being under the puppet.com or puppetlabs.com domains.


donoghuc (approved):
RIP

PR #1359, opened 2023-08-04
Feature: Add `extra_delete` option: rm files after sync

nyetsche:
This PR implements the option to each module of a list of files to delete after the copy into the local directory.

The use case is repositories/modules with "unwanted" information. In our environment all modules downloaded with `r10k` are checked into our main Puppet git repository. We want to skip files in our repo that we don't need, and do so in a repeatable way.

Example `Puppetfile` with this feature:

```
mod 'puppetlabs-cron_core',
  type:    'forge',
  version: '1.2.0',
  extra_delete: ['.github', 'locales', '*/README_*.md', ]
```

This downloads `puppetlabs-cron_core`, then removes the directory `.github`, the directory `locales`, and the glob `*/README_*.md`.

The module on filesytem without this option:

```
original
└── cron_core
    ├── CHANGELOG.md
    ├── CODEOWNERS
    ├── LICENSE
    ├── README.md
    ├── REFERENCE.md
    ├── data
    │   └── common.yaml
    ├── hiera.yaml
    ├── lib
    │   └── puppet
    │       ├── provider
    │       │   └── cron
    │       │       ├── crontab.rb
    │       │       └── filetype.rb
    │       └── type
    │           └── cron.rb
    ├── locales
    │   ├── config.yaml
    │   ├── ja
    │   │   └── puppetlabs-cron_core.po
    │   └── puppetlabs-cron_core.pot
    ├── metadata.json
    ├── pdk.yaml
    └── readmes
        └── README_ja_JP.md

11 directories, 16 files
```

The module on the filesytem with this option as show above:

```
excluded
└── cron_core
    ├── CHANGELOG.md
    ├── CODEOWNERS
    ├── LICENSE
    ├── README.md
    ├── REFERENCE.md
    ├── data
    │   └── common.yaml
    ├── hiera.yaml
    ├── lib
    │   └── puppet
    │       ├── provider
    │       │   └── cron
    │       │       ├── crontab.rb
    │       │       └── filetype.rb
    │       └── type
    │           └── cron.rb
    ├── metadata.json
    ├── pdk.yaml
    └── readmes

9 directories, 12 files
```

This option uses `Dir.glob` for matching: https://ruby-doc.org/core-2.6.3/Dir.html#method-c-glob

If a directory is listed, the entire directory is removed: (`rm -rf`).

Not setting `extra_delete: ...` is equivalent to `extra_delete: []`, which will cause no files to be deleted.

### DIFF
--- a/lib/r10k/module/base.rb
+++ b/lib/r10k/module/base.rb
@@ -42,6 +42,10 @@ class R10K::Module::Base
   #   @return [Boolean] set this to true if the spec dir can be safely removed, ie in the moduledir
   attr_accessor :spec_deletable
 
+  # @!attribute [rw] extra_delete
+  #   @return [List[String]] set this to non-empty array of strings of files to remove from module after sync
+  attr_accessor :extra_delete
+
   # There's been some churn over `author` vs `owner` and `full_name` over
   # `title`, so in the short run it's easier to support both and deprecate one
   # later.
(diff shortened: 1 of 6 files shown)


dbarrous:
This would be such a great feature to have!

PR #1371, opened 2024-01-24
fetch tags to local, if different to remote

sprd-tmu:
We have the use case, that r10k is pulling modules from our own Git and some modules are applied via `puppet apply` based on a tag in the Git repository. Therefore it's needed, that the git tags are also updated in the deployed modules by r10k.

This PR adds updating the tags, when they differ local and remote.
This is running currently in our production and is working fine.

### DIFF
--- a/lib/r10k/git/rugged/working_repository.rb
+++ b/lib/r10k/git/rugged/working_repository.rb
@@ -90,7 +90,7 @@ class R10K::Git::Rugged::WorkingRepository < R10K::Git::Rugged::BaseRepository
     proxy = R10K::Git.get_proxy_for_remote(remote)
 
     options = {:credentials => credentials, :proxy_url => proxy, :prune => true}
-    refspecs = ["+refs/heads/*:refs/remotes/#{remote_name}/*"]
+    refspecs = ["+refs/heads/*:refs/remotes/#{remote_name}/*", '+refs/tags/*:refs/tags/*']
 
     results = nil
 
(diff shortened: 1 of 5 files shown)


sprd-tmu:
Is there anything, I can do, to bring this forward?

justinstoller:
Sorry, I missed this.

The change looks good, but let me double check the `prune => true` behavior won't break any use cases.

sprd-tmu:
I'm still interested to get this merged. Just commenting, to remove the stale label.

sprd-tmu:
Still interested.

sprd-tmu:
And again: Still interested

bastelfreak:
@tmu-sprd I'm not sure if I understand your usecase. Does that mean that you delete and recreate tags, that then point to a different commit?

nabertrand:
While our use case doesn't involve tags as @tmu-sprd describes, I'm hoping the `--prune` flags added in this PR will fix the frequent `fatal: bad object refs/remotes/cache/XXX` errors we see when deploying environments as described in #1399

bastelfreak:
(I am not employed by Perforce, but I've some thoughts)

I'm totally fine with adding the prune option. And that should probably be it's own PR, because it's a bugfix.

---

I'm not sure about updating the tags / comparing them. For years, we taught people to rely on git tags or commits in their Puppetfile when they want to speed up deployments. The assumption was that those are immutable. Because of that, r10k has the `incremental` option to check if the deployed modules match the tags/commits in the Puppetfile. This assumption breaks when tags are manipulated on the gitserver. I'm not sure if that's something we should support.

sprd-tmu:
Thanks for replying and I'll try to describe the use case.

Our Delivery Engineering team use git tags to deploy applications via Puppet. In the source repository there is Puppet code to deploy the applications to Kubernetes, QA, VMs etc. When a new version of an application is ready for deployment a SemVer version tag is added to the feature branch and then this is merged to the main branch. The main branch is referenced in the Puppetfile.
When this tag is added, a CI pipeline starts, which builds the application with the tag version and pushes it to the binary repository. Additionally r10k is triggered via webhook to deploy the source repo as Puppet module. There are custom Puppet functions in place, which evaluate the tags on the Puppet server during catalog compilation and deploys the latest tag version from the binary repo.

By this, it's not needed to change any Puppet code, YAML or Puppetfile to deploy a new version of an app. Just add the new version tag and the app is deployed.

For this to work, it's needed, that the tags are synchronized at the Puppet server. Hence this PR.

I hope, this makes sense.

sprd-tmu:
And again²: Still interested

BoomchainLabs (approved):

PR #1376, opened 2024-03-19
(maint) Update beaker-pe to 3

nmburgan:


donoghuc (approved):
in other projects we've seen some other places there may be issues with this. But as long as we are using the default we should be clear. Thanks.

steveax (approved):

PR #1379, opened 2024-03-29
r10k::module::git: expose remote as getter

bastelfreak:
This allows us to access the remote property to figure out the git URI for a module:

```
$ bundle exec irb
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
irb(main):001:0> require 'json'
=> true
irb(main):002:0> require 'r10k/module_loader/puppetfile'
=> true
irb(main):003:0> puppetfile = R10K::ModuleLoader::Puppetfile.new(basedir: '/home/bastelfreak/code/controlrepo/').load
=>
{:modules=>
...
irb(main):004:0> mod = puppetfile[:modules].map {|mod| mod if mod.class == R10K::Module::Git}.compact.first
=>
...
irb(main):005:0> mod.remote
=> "https://github.com/voxpupuli/puppet-borg"
irb(main):006:0>
```

```
```

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,8 @@ CHANGELOG
 Unreleased
 ----------
 
+- r10k::module::git: expose remote as getter  [#1379](https://github.com/puppetlabs/r10k/pull/1379)
+
 4.0.2
 -----
 
--- a/lib/r10k/module/git.rb
+++ b/lib/r10k/module/git.rb
@@ -46,6 +46,11 @@ class R10K::Module::Git < R10K::Module::Base
   #   @return [String]
   attr_reader :default_override_ref
 
+  # @!attribute[r] remote
+  #   @api private
+  #   @return [String]
+  attr_reader :remote
+
   include R10K::Util::Setopts
 
   def initialize(title, dirname, opts, environment=nil)


binford2k:
what will this be used for?

bastelfreak:
I have a PE customer that needs to parse a Puppetfile in a Plan because he wants to checkout and rebase certain modules. I would like to use this code to get the remote URL so I can clone it.


Edit: some background: the customer has a complex pipeline with around 12 environments. When something is fine in the earliest dev stage it will be tagged and the following dev environment will be rebased to this tag. This ensures quite good that no broken code will land in production. Code has to be staged. Now there might be a situation where external modules with defined resources are updated. And a server in one environment collects the exported defined resources from multiple environments (for example the monitoring server).

When there is a breaking change in the defined resources parameters we are required to update this module in all environments at once which really breaks the whole deployment module. To do this we have a few modules where we say `branch: :control_branch`.

So each module that might need to updated in multiple envs will use this. They will have a branch for each environment. And in anither repo we track a yaml file that maps the desired state of module git tag <> env branch name. We have a PE plan that can/will checkout all repo from one environment with `branch: :control_branch` and rebase their env branch to the desired tag.

When all of this is fully working and tested we want to document it here and provide examples.

binford2k:
I think the big question for me, is whether r10k should be the canonical standard for Puppetfile parsing. It seems slightly odd, but at the same time I can't really think of a better place. I don't think we're quite to the point where it makes sense to break out `libpuppetfile` yet.

binford2k (approved):

donoghuc (approved):

donoghuc:
You only care about r10k 4.y series right?

bastelfreak:
@donoghuc I need this in PE 2021 and want to backport it to 3.x after it got merged.

justinstoller:
Your comment

> When there is a breaking change in the defined resources parameters we are required to update this module in all environments at once which really breaks the whole deployment module.

Makes me think that this workflow needs something like: https://github.com/puppetlabs/r10k/blob/main/doc/dynamic-environments/usage.mkd#deploying-modules , is that an alternative?

bastelfreak:
@justinstoller I would like to see this as well in code-manager. But I still need an option to parse the Puppetfile.

justinstoller:
@bastelfreak fwiw, the module deploy workflow isn't well documented but it can be done by passing the `modules` key in the deploy payload (see: https://www.puppet.com/docs/pe/2021.1/code_manager_api#code_manager_api).

I'm not sure I fully understand the workflow you're doing, I'm just hoping that with some of the already existing, lesser used functionality we could get your deployments working without new patches.

bastelfreak:
Triggering a deploy for a module in all environments is one thing, but the customer wants to rebase branches of a module. And to do this in a PE plan I need to get the remote URL from the Puppetfile. This patch allows me to do this.

justinstoller:
Gotcha. Thanks, for the contribution, @bastelfreak !

justinstoller (approved):

PR #1402, opened 2024-09-20
Require Ruby 3.1

bastelfreak:
This allows us to use the latest minitar and forge_ruby dependencies.

Replaces:
* https://github.com/puppetlabs/r10k/pull/1401
* https://github.com/puppetlabs/r10k/pull/1397

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,10 +22,10 @@ jobs:
         with: # This doesn't seem to work unless we point directly to the secrets
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
-      - name: Set up Ruby 2.6
+      - name: Set up Ruby 3.1
         uses: actions/setup-ruby@v1
         with:
-          ruby-version: 2.6.x
+          ruby-version: 3.1
       - run: gem install bundler
       - uses: actions/checkout@v3
       - name: Set up QEMU
--- a/.github/workflows/rspec_tests.yml
+++ b/.github/workflows/rspec_tests.yml
@@ -12,14 +12,10 @@ jobs:
     strategy:
       matrix:
         cfg:
-          - {os: ubuntu-latest, ruby: 2.6}
-          - {os: ubuntu-latest, ruby: 2.7}
           - {os: ubuntu-latest, ruby: 3.1}
           - {os: ubuntu-latest, ruby: 3.2}
           - {os: ubuntu-latest, ruby: 3.3}
-          - {os: ubuntu-latest, ruby: jruby-9.3}
           - {os: ubuntu-latest, ruby: jruby-9.4}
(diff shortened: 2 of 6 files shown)


smortex (approved):

nmburgan (approved):

justinstoller:
I've been looking into releasing R10K with this PR and there's one question I have. Our puppet-agent and PDK packages support Ruby 3.1 but have not been updated to vendor or even support in code `minitar` 1.x. That means if we release this change the latest r10k gem will not work out of the box with the shipped puppet-agent and pdk packages. Is that a problem?

Internally, we've been trying to consolidate duplicated gems between our vendored r10k package and the puppet-runtime. However, this would only work if we definitely ship r10k with dependency versions that may conflict with newer agent and pdk builds.

justinstoller:
I would be much more comfortable releasing this if we keep 3.1 as a requirement and open up the dependencies to 0.9+ or 1.x compatibility with minitar (and deps that depend on minitar) and change the code to use whichever constants are available.

That way consumers can still load Puppet & R10K as libraries in the same process and users can install the r10k gem without having to install multiple copies of minitar and puppet_forge, but when the pdk and/or the puppet-agent upgrades to using newer minitar we will automatically start using it.

bastelfreak:
Hi!
> Our puppet-agent and PDK packages support Ruby 3.1 but have not been updated to vendor or even support in code minitar 1.x.

This is only relevant for pdk I think? the I just checked puppet-agent-8.10.0-1.el8.x86_64.rpm and it doesn't contain minitar. it's maybe only vendored on windows, but I doubt that r10k works on Windows?

And that leaves pdk. PDK pins to `minitar ~> 0.8`. So we don't have a situation with a too new minitar, it just won't be installed. I don't think that's a problem. If people want to use the Puppet Agent Ruby or PDK Ruby for r10k, it should keep working out of the box and pulls in r10k 4.1.0. And other people already get the newer version. 

I would say this is actually a benefit for Perforce. The new r10k isn't rolled out to all users at the same time, it's basically split into pdk vs non-pdk people. And the group that you officially support, the pdk people, will get it later after it's more tested. 

I don't think it's worth the time to implement support for minitar 0.X and 1.X and I certainly do not have the time. If someone wants to do this, I don't mind. But I think it makes more sense to work on moving forward . Let's drop minitar 0.x in r10k, then in pdk, beaker and puppet-modulebuilder.

justinstoller:
Thanks for your input, Tim! I'm finding an answer to the question of how and when we plan to integrate these updates into PE and will move forward with the release process once I hear back.

justinstoller:
I should have an answer for you end of day Wednesday

justinstoller:
Okay, I _think_ this is what we should do: https://github.com/puppetlabs/r10k/pull/1408 let me know what you think, Tim.

davidsandilands:
@bastelfreak just so we don't lose track of this

PR #1412, opened 2024-12-02
Limit Rugged Control Repo refspec

seanmil:
Some Git platforms (e.g. GitLab) store additional references beyond the ones typically found in refs/heads for various tracking/historical reference purposes. For Control Repos which have been around a long time these additional refs can number in the tens of thousands. Refs that numerous seem to cause issues with some versions of Rugged being unable to properly clone and reference the Control Repo.

This patch limits Control Repo reference fetching to refs/heads, which is all R10K should need anyways.

See https://docs.gitlab.com/ee/development/gitaly.html#gitlab-specific-references for details on some of these additional references.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -8,6 +8,7 @@ Unreleased
 - Require Ruby 3.1 [#1402](https://github.com/puppetlabs/r10k/pull/1402)
 - Replace deprecated action: actions/setup-ruby->ruby/setup-ruby [#1406](https://github.com/puppetlabs/r10k/pull/1406)
 - Ensure git repositories are pruned on fetch [#1410](https://github.com/puppetlabs/r10k/pull/1410)
+- Limit Rugged Control Repo refspec to only clone refs/heads [#1412](https://github.com/puppetlabs/r10k/pull/1412)
 
 4.1.0
 -----
--- a/lib/r10k/git/rugged/bare_repository.rb
+++ b/lib/r10k/git/rugged/bare_repository.rb
@@ -36,7 +36,7 @@ class R10K::Git::Rugged::BareRepository < R10K::Git::Rugged::BaseRepository
     @_rugged_repo = ::Rugged::Repository.init_at(@path.to_s, true).tap do |repo|
       config = repo.config
       config['remote.origin.url']    = remote
-      config['remote.origin.fetch']  = '+refs/*:refs/*'
+      config['remote.origin.fetch']  = '+refs/heads/*:refs/heads/*'
       config['remote.origin.mirror'] = 'true'
     end
 
@@ -60,7 +60,7 @@ class R10K::Git::Rugged::BareRepository < R10K::Git::Rugged::BaseRepository
     proxy = R10K::Git.get_proxy_for_remote(remote)
 
     options = {:credentials => credentials, :prune => true, :proxy_url => proxy}
-    refspecs = ['+refs/*:refs/*']
+    refspecs = ['+refs/heads/*:refs/heads/*']
 
     results = nil
 


justinstoller:
Thanks for the PR, Sean. I'll try to figure this out this week. I'm surprised that the `:prune => true` isn't cleaning up the old refs. I also think we'd need `refs/tags/*` as well. But I'll try to verify those questions.

seanmil:
> Thanks for the PR, Sean. I'll try to figure this out this week. I'm surprised that the `:prune => true` isn't cleaning up the old refs. I also think we'd need `refs/tags/*` as well. But I'll try to verify those questions.

Hi Justin, thanks for the quick update.

The `:prune => true` wouldn't clean up all of these references because a number of them are permanent references GitLab uses to point to key points in time in the repository such as commits used to show MR state, commit refs for pipeline jobs, the commit at which a (GitLab) "project environment" was deployed, etc..  Those just accumulate as the project has more and more history associated with it, so they aren't eligible for pruning.  You can remove them from GitLab (as I did to work around my critical issue at the time) but it isn't a normal procedure and as I recall requires admin-level console access to GitLab.

As far as the refspec, adding `refs/tags/*` may be a good idea regardless, but I'm not sure I've heard of any Control Repo workflows that deploy from tags instead of branch names.  That particular bit of code is, I believe, only used to clone/fetch the bare repository containing the control repo itself.  The clone/fetch for any of the repositories managed by the control repo is in the `lib/r10k/git/working_repository.rb` file instead.  It has been too long since I did this work to remember with clarity, but I believe I based the refspec based on what I saw when running the `git --mirror` command in `lib/r10k/git/shellgit/bare_repository.rb`.  Making the refspec configurable would be okay too, but that seems like a lot of additional work for questionable value.

Hope that helps, thanks for taking a look at it!

justinstoller:
Thanks for clarifying the gitlab usage!

The repository in the codedir (both the environment and any git modules within it) are `ThinRepository`s which inherit from `WorkingRepository`s. The `BareRepository` is used by the git cache. eg, every git module of "puppetlabs-apache" across all your environments are thin repos that point to a cached bare apache repository in your cachedir.

A brief repro with shellgit shows this, though it also seems to be using `refs/heads`, so maybe my concern is unnecessary. I'll keep looking:
```
apache on  HEAD (c4310ce) [✘] Ruby v3.2.0 
 ➜  cat .git/config 
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true
[remote "origin"]
        url = https://github.com/puppetlabs/puppetlabs-apache
        fetch = +refs/heads/*:refs/remotes/origin/*
[branch "main"]
        remote = origin
        merge = refs/heads/main
[remote "cache"]
        url = /home/justin/.r10k/git/https---github.com-puppetlabs-puppetlabs-apache
        fetch = +refs/heads/*:refs/remotes/cache/*

apache on  HEAD (c4310ce) [✘] Ruby v3.2.0 
 ➜  cat /home/justin/.r10k/git/https---github.com-puppetlabs-puppetlabs-apache/config
[core]
        repositoryformatversion = 0
        filemode = true
        bare = true
[remote "origin"]
        url = https://github.com/puppetlabs/puppetlabs-apache
        fetch = +refs/*:refs/*
        mirror = true
````

justinstoller:
From what I can find in the docs the default behavior of git is to pull down any tags that are reachable from a head that is being fetched. So pulling down all of `refs/heads/*` should involve pulling down all of the reachable tags as well.

I tested this by installing a dev build of PE, updating pe-r10k as done in this PR and pulling down a control-repo that used git modules with tags. I then, in a module that the control repo was pulling down, created a non-default branch with a new commit and tag on it, pushed that to Github, and updated my control repo to only reference that tag. It was able to deploy just fine.

justinstoller (approved):

justinstoller:
It looks like this optimization has caused a series of regressions, see https://github.com/puppetlabs/r10k/pull/1417 and https://github.com/puppetlabs/r10k/pull/1422 and I believe we're going to revert and do a new release.

justinstoller:
Maybe we can work together to figure out a solution that works for everyone, like doing some kind of deny list for refs we know will never be needed.

justinstoller:
Okay, @seanmil . We're going to revert this PR and do a release to unblock some folks and then we will will work on making the refspec configurable before the next full release of PE.

justinstoller:
Don't upgrade to 5.0.2! It contains the reversion of this work. We will work on getting a 5.1.0 that makes the refspec configurable out in the next month.

PR #1422, opened 2025-07-01
Revert "Limit Rugged Control Repo refspec"

kenyon:
This reverts commit c582010a1f4f7f40a3888b26e8c8360589219350 and the subsequent tweak commit 9f9432b779e3ab769a2b65fb4d33f222c351669c, which prevented fetching refs such as `refs/pull/*`, which may be used in a `Puppetfile` for testing pull requests of modules. The commit message of c582010a1f4f7f40a3888b26e8c8360589219350 makes it sound like the change only applies to control repos, but it in fact applies to any git repository used in a `Puppetfile`.

### DIFF
--- a/CHANGELOG.mkd
+++ b/CHANGELOG.mkd
@@ -4,6 +4,7 @@ CHANGELOG
 Unreleased
 ----------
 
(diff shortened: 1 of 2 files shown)


justinstoller (approved):
Thanks for your contribution, @kenyon !

PR #1430, opened 2025-12-15
(maint) Upgrade beaker version

mcdonaldseanp:
Update versions of beaker and friends to latest, so that we can take up changes requried to be able to test on el10 (including things like no longer using `ntpdate`)

```
```

### DIFF
--- a/integration/Gemfile
+++ b/integration/Gemfile
@@ -10,9 +10,9 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 4.40')
-gem 'beaker-pe', '~> 3.0'
+gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 6')
+gem 'beaker-pe', '~> 3.6'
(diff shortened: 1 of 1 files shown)


jonathannewman (approved):

integration/Gemfile
```diff
@@ -10,9 +10,9 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 4.40')
-gem 'beaker-pe', '~> 3.0'
+gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 6')
```
bastelfreak, line 13: is there a reason for going to the old beaker 6, instead of latest 7?

justinstoller (approved):

integration/Gemfile
mcdonaldseanp: Unfortunately for us our CI system doesn't support ruby 3.2 yet 😢 with the ruby requirement in beaker 7 for ruby 3.2 we wouldn't be able to run our test suites on r10k for PE with beaker 7

PR #1441, opened 2026-08-18
(PE-45753) Guard every r10k agent run against catalog-run lock race

steveax:
## Problem

The r10k acceptance gates (`enterprise_pe-r10k-vanagon_pkg-int-sys-testing_{main,2025.12.x,2023.8.x}`) are chronically flaky: a different random cell/test file fails each build with

```
puppet agent --test  ->  exited 1
Notice: Run of Puppet configuration client already in progress; skipping
        (/opt/puppetlabs/puppet/cache/state/agent_catalog_run.lock exists)
```

The daemonized puppet agent's periodic run holds `agent_catalog_run.lock` exactly when a test fires its own `puppet agent --test`.

## Why the previous fix was insufficient

The prior fix disabled `Service[puppet]` **once** at the end of the pre-suite (`20_pe_r10k.rb`). But many of the suite's agent runs execute against the PE `production` environment, which re-applies the PE agent profile and **re-enables/restarts `Service[puppet]`** — re-arming the daemon for every subsequent run. Post-fix builds still failed with the identical signature (verified on all three lines), because that one-shot disable protected only 1 of the suite's agent runs.

## Fix

Replace the one-shot disable with a **per-run guard**, following the pattern beaker-pe uses for exactly this race:

- New `run_puppet_agent(host, environment, acceptable_exit_codes, &block)` helper in `integration/lib/r10k_utils.rb`:
  1. `stop_agent_on(host)` — stops the agent service **and waits for any in-flight run to release `agent_catalog_run.lock`**.
  2. runs `puppet agent --test --environment <env> --waitforlock 1` — so the explicit run **waits** for the lock instead of erroring if the daemon re-armed in the interim.
- **All 45** agent invocations across the pre-suite and tests now route through the helper (both the literal `--environment production` and the interpolated `--environment #{env}` forms).
- The stale one-shot disable step and its (incorrect) "nothing re-enables the service" comment are removed.

## Verification

- All 45 call sites preserve their original host, environment, and exact `acceptable_exit_codes` (`2` / `1` / `[0,2]` / bare) — verified by tuple comparison against `main`.
- `ruby -c` clean on all 30 changed files.
- Independent review confirmed block-forwarding, exit-code fidelity, `stop_agent_on` resolvability (beaker-pe is loaded in `00_pe_install.rb`), and that no call site used the `on(...)` return value.

Real validation is a run of the three `pkg-int-sys-testing` gates on this branch.

### DIFF
--- a/integration/lib/r10k_utils.rb
+++ b/integration/lib/r10k_utils.rb
@@ -1,5 +1,46 @@
 require 'git_utils'
 
+# Run "puppet agent --test" on a host without racing the daemonized agent's
+# periodic catalog run (PE-45753).
+#
+# Several r10k tests run against the PE "production" environment, which
+# re-applies the PE agent profile and re-enables/restarts Service[puppet].
+# The daemon's background run can then hold agent_catalog_run.lock exactly
+# when a test fires its own "puppet agent --test", failing it with "Run of
+# Puppet configuration client already in progress; skipping".
+#
+# The pre-suite (20_pe_r10k.rb) stops and disables Service[puppet] once, which
+# covers the windows between explicit runs (r10k deploys, config writes,
+# teardown). This helper handles the remaining case: a run against "production"
+# re-enables the service, so "--waitforlock 1" makes each explicit run wait for
+# the lock (polling once a second) rather than erroring if the daemon re-armed.
+# "--maxwaitforlock 5m" raises the wait ceiling above Puppet's 60s default so a
+# slow daemonized production catalog run cannot clip the wait on slow CI hosts.
+# --waitforlock also clears a stale lock (dead PID) via Puppet's pidlock, so it
+# never hangs the run the way an unconditional lockfile poll would.
+#
+# ==== Attributes
+#
+# * +host+ - The host (or hosts array) to run the agent on.
+# * +environment+ - The Puppet environment to run against. Defaults to
+#   "production".
+# * +acceptable_exit_codes+ - Exit codes to accept, passed through to +on+.
+#   When nil (default), +on+'s default is used.
+#
+# ==== Returns
+#
+# The result of the +on+ call (yields to a block if given, like +on+).
+#
+# ==== Examples
+#
+# run_puppet_agent(agent, 'production', 2) do |result|
+#   assert_match(/expected/, result.stdout)
+# end
+def run_puppet_agent(host, environment = 'production', acceptable_exit_codes = nil, &block)
+  opts = acceptable_exit_codes.nil? ? {} : { :acceptable_exit_codes => acceptable_exit_codes }
+  on(host, puppet('agent', '--test', "--environment #{environment}", '--waitforlock 1', '--maxwaitforlock 5m'), opts, &block)
+end
+
 # Retrieve the file path for the "r10k.yaml" configuration file.
 #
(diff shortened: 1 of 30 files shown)


coreymbe:
Two asks before this merges. First, `stop_agent_on` ends in an unbounded `while` loop polling `agent_catalog_run.lock` (`beaker-puppet/lib/beaker-puppet/helpers/puppet_helpers.rb`, still present on `main` as of 4.4.3), so a stale lock now hangs the whole run until the Jenkins timeout instead of failing in seconds with a greppable message, and that's on all 45 sites including one inside a `teardown`.

Second, the deleted pre-suite step did `ensure=stopped enable=false`, whereas `stop_agent_on` only stops (line 727) and never disables, so the daemon stays re-armable and we lose the protection we had during `r10k deploy`, `r10k.yaml` writes, and teardown. Both are fixed by the same move: put the one-shot stop-and-disable back in the pre-suite, keep `--waitforlock 1` in the helper for the re-arm window, and drop `stop_agent_on` from it. That also drops at least 90-odd redundant remote puppet invocations, since every call re-reads `vardir` via `--configprint` (beaker builds a fresh `PuppetConfigReader` per access) and re-applies `service puppet ensure=stopped`, neither of which changes across the suite.

steveax:
Thanks @coreymbe — both addressed in the force-pushed update (`7aede61`):

1. **Unbounded lock poll / teardown hang** — dropped `stop_agent_on` entirely, so there's no unconditional `while file_exist?` loop anywhere (including the teardown site). The helper now relies solely on `--waitforlock 1`, which uses Puppet's pidlock: a live lock is waited on (up to `maxwaitforlock`), and a *stale* (dead-PID) lock is cleared and the run proceeds — so a stale lock no longer hangs the run to the Jenkins timeout.

2. **Re-armable daemon** — restored the one-shot `ensure=stopped enable=false` in `20_pe_r10k.rb` to cover the between-run windows (r10k deploys, config writes, teardown). `--waitforlock 1` in the helper covers the remaining case where a run against `production` re-applies the PE profile and re-enables the service.

This also removes the ~90 redundant remote calls you flagged — no more per-invocation `--configprint` (`vardir`) + `ensure=stopped`. All 45 sites still route through `run_puppet_agent`; exit codes unchanged.

coreymbe (approved):
LGTM! Going to approve this, but want to mention that `--waitforlock 1` is now the only per-run guard, which makes `maxwaitforlock`'s 60 second default more load-bearing than it was when `stop_agent_on` was also in the path. If a daemonized production catalog run ever exceeds a minute, the wait expires and `puppet agent --test` exits 1 (`puppet/agent.rb:97-99` into `application/agent.rb:444-445`), so the test fails with "Exiting now because the maxwaitforlock timeout has been exceeded" instead of the old message. An explicit `--maxwaitforlock 5m` would close that. Non-blocking, since the restored pre-suite disable should mean the daemon is rarely running at all, but it is cheap insurance on the one path that's left.

steveax:
Took the insurance, @coreymbe — added `--maxwaitforlock 5m` alongside `--waitforlock 1` in the helper so a slow daemonized production catalog run on a 2-core pooler host can't clip the 60s default wait. Amended into `b6cdbbe`.

